PR #20, opened 2019-01-29
documentation fixes

Mic92:


cpriebe:
Hi @Mic92,

first of all thanks for those fixes. Let us know when you think the pull request is complete and I'm happy to merge it.

Regarding building the JDK: It is almost exactly the same as the Alpine one with two exceptions:

1) One of the Alpine patches results in a fork call which SGX-LKL currently does not support. So we undo/remove that patch.
2) We disable mmap file support for libzip as file-backed mappings don't work particularly well in an enclave without proper virtual memory support.

I'll see if I can commit a Dockerfile later this week that will allow you to build the JVM (or rather a more up-to-date one) for SGX-LKL yourself.

Mic92:
I think this is all for this pull request.

cpriebe:
Thanks a lot for your contribution!

I am leaving out the helloworld maven commit, since we would like to keep the examples simple. This is not an example about how to build a Java project (with Maven) but about how to run the JVM with SGX-LKL :) If people want to run existing Java applications there most likely is a build system in place already.

Also I have adjusted your commit message regarding the test operator. The problem is not that `[` doesn't support `==` but that the variable in this case can be empty/unset.

I have merged your changes in c055b43, b40705b, and a386521. Thanks again!

PR #25, opened 2019-03-22
Fix debug build in sgx-lkl-docker.sh

letmaik:
The build checks against the string `true`, not `1`. Leaving out `-r` in sgx-lkl-docker.sh did not produce a debug build. This PR fixes that.


cpriebe:
Thanks a lot! Looks good to me. I'll merge it!

PR #43, opened 2019-06-28
add support for block devices as disks

Mic92:


cpriebe:
Thanks a lot, that's definitely useful! Looks good to me!

PR #44, opened 2019-06-28
add support for ipv6

Mic92:
Ipv4 is on life support


src/main/sgxlkl_config.h
```diff
@@ -10,70 +10,76 @@
...
  * either we are in simulation mode, or we are in HW mode and a key is provided
  * via SGXLKL_KEY.
  */
+#define DEFAULT_SGXLKL_GW6 "fe80::1"
 #define DEFAULT_SGXLKL_HEAP_SIZE 200 * 1024 * 1024
 #define DEFAULT_SGXLKL_HOSTNAME "lkl"
 #define DEFAULT_SGXLKL_IAS_QUOTE_TYPE "Unlinkable"
 #define DEFAULT_SGXLKL_IAS_SERVER "test-as.sgx.trustedservices.intel.com:443"
 #define DEFAULT_SGXLKL_IP4 "10.0.1.1"
+#define DEFAULT_SGXLKL_IP6 "fd16:012a:4639:184d::1"
```
Mic92: ULA-Address chosen by fair dice role!

src/main/sgxlkl_config.h
```diff
@@ -10,70 +10,76 @@
...
+#define SGXLKL_WG_KEY                   61
+#define SGXLKL_WG_PEERS                 62
 
 
 #define DEFAULT_SGXLKL_GW4 "10.0.1.254"
 /* The default heap size will only be used if no heap size is specified and
  * either we are in simulation mode, or we are in HW mode and a key is provided
  * via SGXLKL_KEY.
  */
+#define DEFAULT_SGXLKL_GW6 "fe80::1"
```
Mic92: Also router advertisement could be used instead to get a gateway using the link-local address `fe80::1` is a popular choice.

src/lkl/setup.c
```diff
@@ -625,6 +625,25 @@ void lkl_poststart_net(enclave_config_t* encl, int net_dev_id) {
             }
         }
 
+        res = lkl_if_set_ipv6(ifidx, encl->net_ip6.s6_addr, encl->net_mask6);
```
Mic92: We have to set the ipv6 address after bringing the interface up because lkl performs duplicate address detection when applying an address.

src/lkl/setup.c
```diff
@@ -625,6 +625,25 @@ void lkl_poststart_net(enclave_config_t* encl, int net_dev_id) {
...
+        if (res < 0) {
+            fprintf(stderr, "Error: lkl_if_set_ipv6(): %s\n",
+                    lkl_strerror(res));
+            exit(res);
+        }
+
+        if (encl->net_gw6.s6_addr > 0) {
+          res = lkl_if_set_ipv6_gateway(ifidx,
```
Mic92: Since we use a link-local ipv6 address by default, we need to specify the interface for the default gateway.

cpriebe:
That's great! Thanks a lot!

PR #45, opened 2019-07-01
Only align percpu variables lkl linker scripts

Mic92:
Adding a plt.bnd section breaks the enclave signing somehow,
now idea why. Having code in a plt section also looks fishy to me.
Now only percpu variables are page aligned and put before the
normal data section.


cpriebe:
Thanks! Looks good to me.

PR #46, opened 2019-07-01
Add option to set working directory

Mic92:


cpriebe:
Looks mostly good to me! The working directory configuration should most likely be part of the app configuration though. The app configuration can be provided remotely and it's probably a bad idea to allow the host to have full control over the working directory?

cpriebe:
Merged manually.

PR #61, opened 2019-08-30
Use the correct target name for liblkl to make tools build depend on lkl

jhand2:
I recently learned about this project and wanted to check it out (cool stuff!)

I did hit a small hiccup when building though. The build for the tools target does not properly depend on liblkl as it uses a nonexistent variable name in the Makefile. So if you run `make -j$(nproc) tools`, the lkl headers are missing. If you rune make with one processor, this problem does not manifest.


cpriebe:
Hi @jhand2,

Thanks a lot for the pull request! Looks good to me! Merged.

PR #65, opened 2019-09-24
Use Intel attestation services API v3

santoshx:
Removes use of authentication certificate & key.
Makes use of subscription key.


cpriebe:
Thanks a lot! That looks very useful. I assume there is a corresponding patch for sgx-lkl-musl (see Travis errors)? Also, why is the remote attestation port configuration being removed? It's used for attesting SGX-LKL via `sgx-lkl-ctl`. Wouldn't that still be required?

santoshx:
> Thanks a lot! That looks very useful. I assume there is a corresponding patch for sgx-lkl-musl (see Travis errors)? Also, why is the remote attestation port configuration being removed? It's used for attesting SGX-LKL via `sgx-lkl-ctl`. Wouldn't that still be required?

@cpriebe I'll look into travis errors and also bring back port config.

cpriebe:
Apologies for the delay! I didn't see the recent changes until now.

I have tested the changes and they look good to me. I have also updated the wiki page accordingly:

https://github.com/lsds/sgx-lkl/wiki/Remote-Attestation-and-Remote-Control

Thanks a lot!

PR #66, opened 2019-10-03
Remove lthread_madvise() so that apps that modify esp will not violat…

thomasten:
…e the assert.

An example is an app that uses lthread itself.
last_stack_size is unused so I removed that, too.


cpriebe:
Makes sense, especially considering that this function doesn't fulfil its original purpose anyway (calling madvise).

Thanks!

PR #67, opened 2019-10-03
Enable encryption of non-root disks.

thomasten:
Key and root hash of additional disks in the disk_config array of the
application configuration are now applied.


cpriebe:
Thanks a lot! I had this on a todo list somewhere for a while now, so this is very welcome :)

PR #68, opened 2019-10-04
Allow mounting disk images on existing folders.

thomasten:
This enables the root image to be read-only.


cpriebe:
Looks good to me. Thanks a lot!

PR #70, opened 2019-10-08
Improve docker runtime

bengreenier:
This adds support for enhanced docker image generation using the tooling built into the core project. It's based off the discoveries made in https://github.com/bengreenier/sgx-lkl-turtles by myself, and my team.

> Note: for completeness, [here is the conversation and initial review](https://github.com/bengreenier/sgx-lkl/pull/4) with my team.

We've made these changes to support a development scenario that we think is quite promising going forward...Where the LKL app is defined in a self-contained (and `alpine` based) Dockerfile, that some tooling parses and processes to produce an image containing the lkl runtime (`sgx-lkl-run`) and the `img`, along with the desired runtime arguments and environment variables. Hopefully this pattern is found to be beneficial to other folks as well!

This doesn't change the shell API, but does enhance it's capabilities in the following ways:
+ Accepts container tags that aren't running/haven't run
+ Accepts container tags with `:` on oses that don't support `:` in fs paths (replaces with `_`)
+ Reads `WORKDIR` from docker image, and maps it at runtime
+ Reads `ENTRYPOINT` from docker image, and maps it at runtime
+ Reads `ENV` from docker image, and maps it at runtime
+ Overrides `ENTRYPOINT` with command line args (shell API args take precedence)
+ Exits (`exit 1`) when neither a docker `ENTRYPOINT` nor shell API arg are given (we must have an entrypoint)
+ Clarifies output tag via echo message (`{inputTag}-secure`)
+ Clarifies runtime requirement for `--privileged` via echo message
+ Reduces stdout noise in final container (no more `net.ipv4.ip_forward=1` message)


cpriebe:
This looks very useful! Thanks a lot for the pull request! I'm happy to merge it.

PR #72, opened 2019-10-10
Remove an unused declaration of do_ereport

santoshx:


cpriebe:
Hey @santoshx,

Thanks for the pull request! Could you say something about why this is necessary? Does the declaration produce warnings? I believe the header file is an unmodified copy from the linux-sgx repository needed by some of the remote attestation code if I remember correctly. While `do_ereport` might be unused that's probably also true for some (or all?) of the other function declaration in that file?

Of course we could think about stripping those header files down to only include exactly what's needed. For the same reason we previously removed a dependency on the full `linux-sgx` repository as a git submodule and rather include the few header files we depend on directly.

santoshx:
@cpriebe 
I was doing something and this unused declaration created bit of confusion. I did not realize file is an unmodified copy from linux-sgx.  Thank you for clarification.  I close this PR.

PR #73, opened 2019-10-14
Improve generate_syscall_remap.py to get more syscall numbers from un…

thomasten:
I changed the generate_syscall_remap script because the mapping was missing some syscalls I needed. However, in syscall.c some syscalls are mapped to numbers >1000 that are not generated by this script. I don't know where those come from and they seem to always end up in lkl_syscall() returning ENOSYS.
What is the origin of these numbers? I think it would be good to keep them instead of replacing them with -1. That would be helpful when using SGXLKL_TRACE_SYSCALL. Also, one can implement them in syscall.h. (E.g., there is no dup2 in LKL, but the call can easily converted to dup3.)
If these numbers have no meaning (anymore), the script may generate ascending numbers >1000 instead of -1. What do you think?

### DIFF
--- a/tools/generate_syscall_remap.py
+++ b/tools/generate_syscall_remap.py
@@ -8,8 +8,6 @@
 LKL_UNISTD_PATH="../lkl/tools/lkl/include/lkl/asm-generic/unistd.h"
 NATIVE_TBL_PATH="../lkl/arch/x86/entry/syscalls/syscall_64.tbl"
 
-LKL_PREFIX='#define __lkl__NR_'
-
 class Syscall:
     def __init__(self, name):
         self.name = name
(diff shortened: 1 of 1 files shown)


cpriebe:
Hey!

Thanks a lot for the pull request. To be honest, I don't really know what the reasoning was behind the high LKL syscall numbers as this wasn't done by me. Presumably they were simply chosen to be high enough to ensure that they won't map to new syscalls in the future and to guarantee that LKL would return `ENOSYS`. In any case, I'd say the choice of the exact numbers is rather arbitrary so it would be fine to auto-generate them as part of the script as you have proposed.

The advantage of that is that they will appear in the `SGXLKL_TRACE_SYSCALL` output as you have said. On the other hand, `syscall` will already output a warning to stderr if the x86 syscall number maps to `-1`. I think I actually prefer this as it's far more visible than if people have to guess why their applications don't work, build SGX-LKL in debug mode, enable `SGXLKL_TRACE_SYSCALL` and look through the logs to find out about the missing mapping.

Regarding the `dup2` example you gave: Most applications wouldn't use the `syscall` libc function directly but rather use the corresponding libc wrapper. In that case, what you propose is already done transparently. sgx-lkl-musl is built against the LKL headers. If `SYS_dup2` is not defined, it will use the `dup3` syscall to implement `dup2` (see `https://github.com/lsds/sgx-lkl-musl/blob/master/src/unistd/dup2.c`). 

Of course since we would like to be as portable as possible, it would be good if we could handle the case of applications using `syscall` directly too but I guess that would have to be implemented on a case by case basis. The easiest way to solve this would be to call the existing libc wrapper function. We could probably do this as part of `syscall.c#syscall` itself?

I would propose to do the following:

1) Merge this pull request and use the mapping table as generated by the `generate_syscall_remap.py`, i.e. replace the high numbers for non-implemented LKL syscalls with `-1`.
2) Add verbose logging to `syscall.c#syscall` so that `SGXLKL_TRACE_SYSCALL` logs still include the original request and whether it is either not supported or has been redirected.
3) I would say it makes sense to print an additional warning to `stderr` for missing mappings as I think more often than not applications would not expect the request to return `ENOSYS` and handle it. This way, users that try to run applications that crash due to an unsupported system call would immediately know the cuase. But that's a judgement call I guess. I am unsure how often this happens in general, and how often applications actually do handle that case correctly in which case you might not want to see that warning?
4) We can catch special cases like `dup2` for which libc wrappers exist in `syscall.c#syscall` and redirect them.


What do you think?

thomasten:
Thanks for the detailed explanation. I haven't been aware of the warning on -1 because dup2 currently maps to a high number so it didn't appear. I'd also prefer that solution.

1. Agreed.
2. Agreed.
3. I'd also tend to print the warning, i.e., to keep the code as is.
4. This is probably a good solution. However, we must be careful as the wrappers do not always behave exactly like syscall(). E.g., dup2 handles EBUSY. I cannot think of a case where this difference matters so in this particular case I would prefer to call dup2() instead of reimplementing the logic that considers the edge case.

For my part, this PR can be merged. I will then proceed with the changes in sgx-lkl-musl.

cpriebe:
Great! 4 is a good point. This definitely requires some extra care.

I'm going to merge the pull request. Thanks again!

PR #101, opened 2020-04-29
Add CI

letmaik:
Various changes to the CI scripts to make them work on a stock Ubuntu 18.04 image. Note that the `$USER` work-around is only temporary and is being fixed in Azure Pipelines.

Also, note that I changed the submodule URLs to use http consistently so that it can be cloned without private ssh key being available. This allows to run CI on external PR builds without allowing access to any secrets. If you want to use ssh locally for the submodules then you can still do that without changing the URLs, see https://stackoverflow.com/a/31185110.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,11 +1,11 @@
 [submodule "lkl"]
 	path = lkl
-	url = ssh://git@github.com/lsds/lkl.git
+	url = https://github.com/lsds/lkl.git
 	branch = upstream-refactor
 [submodule "host-musl"]
 	path = host-musl
 	url = https://github.com/lsds/musl.git
 [submodule "sgx-lkl-musl"]
 	path = sgx-lkl-musl
-	url = ssh://git@github.com/lsds/sgx-lkl-musl.git
+	url = https://github.com/lsds/sgx-lkl-musl.git
 	branch = oe_port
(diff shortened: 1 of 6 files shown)


.gitmodules
```diff
@@ -1,11 +1,11 @@
 [submodule "lkl"]
 	path = lkl
-	url = ssh://git@github.com/lsds/lkl.git
+	url = https://github.com/lsds/lkl.git
```
davidchisnall, line 3: Can you do the relative path thing from the previous version?  This version now means that it's annoying for people who want to commit changes to any of the submodules because they will be cloned from the read-only (or write-horribly-insecurely) path.

davidchisnall (commented):
Looks good, apart from one minor nit.

.gitmodules
letmaik: I don't know to be honest. I think it's quite standard to use full https URLs for submodules. And if you prefer to use ssh then you can tell git to do that: https://stackoverflow.com/a/31185110. Would that not work? I'm not a big fan of using relative paths in the end because then anyone who forks wouldn't be able to clone --recursive if they haven't forked all submodules as well.

davidchisnall: So, I feel stupid now.  I forgot that my `.gitconfig` already contained this section:

```ini
[url "ssh://git@github.com"]
        insteadOf = https://github.com
```

davidchisnall (approved):
Looks good, thank you!

PR #108, opened 2020-04-30
Dump attested TLS credentials upon starting up

jxyang:


letmaik:
For each failure mode, could you add a SGXLKL_VERBOSE line?

letmaik:
I wonder if this should be an optional feature enabled in the app config since it adds a strict dependency to DCAP and hence network.

davidchisnall:
[ Duplicating our private discussion here ]

We need to address the layering here before it can be merged.  We should not be introducing new code into `src/enclave` that has nontrivial dependencies on libc.  Probably the best structure for this is to:

 - Provide functionality in src/enclave for getting the raw attestation information, without mbedTLS dependencies.  This can talk to OE APIs directly.
 - Modify LKL to allow exposing that somewhere in sysfs (or possibly as a device node) by calling the APIs added in src/enclave.
 - Add the userspace component that reads the raw data from the kernel-owned file, uses mbedTLS to generate the certificates, and puts them in a well-known location.

src/enclave/gencreds.c
```diff
@@ -0,0 +1,415 @@
...
+    if (cert_size_out)
+        *cert_size_out = 0;
+
+    if (private_key_out)
+        *private_key_out = NULL;
+
+    if (private_key_size_out)
+        *private_key_size_out = 0;
+
+    if (!cert_out || !cert_size_out || !private_key_out ||
```
Pengpeng-Microsoft, line 175: Not clear about the logics here. 

From lines 163-173, all the values pointed by these out parameter pointers are assigned NULL or 0 before the 'if' logic checks pointers and sends the flow to 'done'. Who will be manage the memory pointed by these pointers before nullifying them?

src/enclave/gencreds.c
```diff
@@ -0,0 +1,415 @@
...
+        oe_free_key(private_key, private_key_size, NULL, 0);
+
+    if (cert)
+        oe_free_attestation_certificate(cert);
+
+    return ret;
+}
+
+
+static int _write_file(const char* path, const void* data, size_t size)
```
Pengpeng-Microsoft, line 242: Both this (_write_file) and _load_file are pretty general utility functions. Do we have a more common place for them than gencreds.c?

Pengpeng-Microsoft (commented):
> [ Duplicating our private discussion here ]
> 
> We need to address the layering here before it can be merged. We should not be introducing new code into `src/enclave` that has nontrivial dependencies on libc. Probably the best structure for this is to:
> 
> * Provide functionality in src/enclave for getting the raw attestation information, without mbedTLS dependencies.  This can talk to OE APIs directly.
> * Modify LKL to allow exposing that somewhere in sysfs (or possibly as a device node) by calling the APIs added in src/enclave.
@davidchisnall , which flow triggers sysfs here?
> * Add the userspace component that reads the raw data from the kernel-owned file, uses mbedTLS to generate the certificates, and puts them in a well-known location.
Is it still triggered by startmain in enclave_init.c as in the current change?

PR #112, opened 2020-04-30
Install sgx-lkl-gdb and make more dev friendly

letmaik:
This PR does the following:
- Installs sgx-lkl-gdb into bin/. Installs gdb helper files into lib/gdb.
- Removes setup.sh, being a common pain point and not really necessary.
- Falls back to /opt/openenclave (if available) if pkg-config could not find OE.

I think this improves the out-of-the-box dev experience. Getting rid of the script generation step also makes it simpler to modify the script.


prp (approved):
LGTM

PR #113, opened 2020-04-30
Add SGXLKL_TRACE_DISK

letmaik:
This PR adds SGXLKL_TRACE_DISK which is currently only used to enable trace output from cryptsetup. It may be used in more/different places in the future.


prp (approved):
LGTM

PR #114, opened 2020-04-30
Add test for "environment" key in app config

letmaik:
This test currently fails with "KeyError: 'FOO'", meaning the environment variable specified in the app config does not propagate to the application.

After the bug is fixed, this PR can be merged.


letmaik:
@prp Should we do the same here and disable it for CI even though its not an intermittent failure? I think that makes it easier for local testing when someone tries to fix the bug.

prp:
> @prp Should we do the same here and disable it for CI even though its not an intermittent failure? I think that makes it easier for local testing when someone tries to fix the bug.

Agreed, thanks.

prp (approved):
LGTM

PR #115, opened 2020-04-30
Contiguous argv/envp in app_config. Fixes #103.

### LINKED ISSUE #103: sgx-lkl fails to start nodejs app with an assert
SGX-LKL code that sets up the environment for the node process throws the assert.

Node expects the argv memory to be adjacent. The code that builds argv needs to be fixed. 

Error:
[[ SGX-LKL ]] libc_start_main_stage2(): Calling app main: usr/local/bin/node
Assertion failed: process_title.len + 1 == size (../deps/uv/src/unix/proctitle.c: uv_setup_args: 65)
Aborted
[[ SGX-LKL ]] lkl_terminate(): terminating LKL (exit_status=134)

Issue seems here: https://github.com/lsds/sgx-lkl/blob/oe_port/src/enclave/enclave_init.c#L38-L64

wintersteiger:
Signed-off-by: Christoph M. Wintersteiger <cwinter@microsoft.com>


letmaik:
Do you mean #114?

wintersteiger:
No, this is a fix for #103. If it accidentally fixes #114 too, that's fine with me :-)

src/enclave/enclave_init.c
```diff
@@ -39,24 +39,40 @@ static void __sgxlkl_enclave_copy_app_config(
...
-        app_config->envp[j] = strdup(envp[j]);
-    app_config->envp[j] = NULL;
+    char* buf = malloc((total_size + 1) * sizeof(char*));
+    app_config->argv = malloc((sgxlkl_config->argc + envc + 2) * sizeof(char*));
+
+    char* p = buf;
+    for (i = 0; i < app_config->argc; i++)
+    {
+        app_config->argv[i] = p;
+        p += sprintf(p, "%s", sgxlkl_config->argv[i]) + 1;
```
davidchisnall: This looks as if it's a TOCTOU vulnerability.  On line 54, you check the length, then an attacker removes the null terminator, and then sprintf copies the data.

There is no reason for using sprintf here, please record the lengths and then use memcpy and assert that you have a null terminator.

src/enclave/enclave_init.c
```diff
@@ -39,24 +39,40 @@ static void __sgxlkl_enclave_copy_app_config(
...
     char** envp = NULL;
+    size_t total_size = 0;
 
     app_config->argc = sgxlkl_config->argc;
-    app_config->argv = malloc((app_config->argc + 1) * sizeof(char*));
 
     for (i = 0; i < app_config->argc; i++)
-        app_config->argv[i] = strdup(sgxlkl_config->argv[i]);
-    app_config->argv[i] = NULL;
+        total_size += strlen(sgxlkl_config->argv[i]) + 1;
```
davidchisnall: Missing a check that `sgxlkl_config->argv[i]` and `sgxlkl_config->argv[i] + strlen(`sgxlkl_config->argv[i])` are both outside the enclave (and not spanning the enclave).

src/enclave/enclave_init.c
```diff
@@ -39,24 +39,40 @@ static void __sgxlkl_enclave_copy_app_config(
...
 
     envp = sgxlkl_config->argv + sgxlkl_config->argc + 1;
-    for (; envp[j++] != NULL;)
-        ;
+    while (envp[envc] != NULL)
+    {
+        total_size += strlen(envp[envc]) + 1;
```
davidchisnall: Missing the same checks as above.

src/enclave/enclave_init.c
```diff
@@ -39,24 +39,40 @@ static void __sgxlkl_enclave_copy_app_config(
...
+        envc++;
+    }
 
-    app_config->envp = malloc((j + 1) * sizeof(char*));
-    for (j = 0; envp[j] != NULL; j++)
-        app_config->envp[j] = strdup(envp[j]);
-    app_config->envp[j] = NULL;
+    char* buf = malloc((total_size + 1) * sizeof(char*));
+    app_config->argv = malloc((sgxlkl_config->argc + envc + 2) * sizeof(char*));
```
davidchisnall: It's not clear to me that this is sufficient.  We should probably mirror the layout mandated by the amd64 ELF psABI here (ideally, we'd have that structure on the stack on userspace process start, because that's what the ABI actually requires).  The layout there is (from the lowest allocated stack address upwards):

- 64-bit `argc`
- `argv` as an array of 64-bit pointers.
- 64-bit 0
- Environment pointers as an array of 64-bit pointers.
- 64-bit 0
- ELF auxiliary argument vector (array of pairs of 64-bit values)
- 64-bit 0.
- Platform-specific things, including all of the things that the pointer arrays point to.

If we have software that depends on this exact layout, it will remain broken with your code.

davidchisnall (changes requested):
We should probably simplify the in-enclave code by modifying the init structure to contain, for each of `argv` and `environ`:

- A null-separated string.
- The length of the string.
- The number of entries in the string.

We can then do a contiguous copy of the strings once and initialise the arrays of pointers to the fields by scanning the arrays.  Given that the ABI requires that we have these things contiguous in memory, it might be simpler to have a single contiguous block of strings and then the offset of the environment.  For example, with: `argv = {"arg1, "arg2", "arg3" }` and `environ = {"env1=val", "env2=val"}`, have a structure with fields something like this:
```C
const char *strings = "arg1\0arg2\0\arg3\0env1=val\0env2=val";
uint64_t argv_length = 12;
uint64_t argc = 3;
uint64_t environ_length = 18
uint64_t environ_count = 2
```

We then have one check that `strings` to `strings + argv_length + environ_length` is within the enclave and copy that in.  We allocate a buffer that is `argv_length  + environ_length + (sizeof(char*) * (argc + environ_count + 2)` bytes long (plus one if we want to have an empty ELF aux arguments vector), then copy the entire string in at the top, and iterate over it finding null pointers and initialise the elements in our `argv` and `environ` with pointers, asserting if we find more nulls than we expect to find arguments or environment variables.

src/enclave/enclave_init.c
wintersteiger: All this string are in enclave memory (see sgxlkl_config.c).

src/enclave/enclave_init.c
wintersteiger: See above, all those string are in enclave memory.

wintersteiger:
"We should probably...": Isn't this exactly what my fix does? Anyways, full ELF compatibility is not on my work plan :-)

`app_config->envp` is currently not merged into `sgxlkl_enclave->envp` (root cause of #114). 

Our own code also depends on the argv+envp layout here: https://github.com/lsds/sgx-lkl/blob/a0f6dd41c5c44a8cfe7d08a99a6da4d548d36ae6/src/enclave/enclave_init.c#L193

davidchisnall:
> "We should probably...": Isn't this exactly what my fix does?

I may be misreading you PR, but I thought it was just changing how these are payed out in a copy (which you have now explained is a second copy of the originally). I am suggesting a cleaner layout in the original version provided by the host, which would remove the unsafe strip use in the first copy.

If this is going to be simplified by the move to JSON for the confit then it might make sense to defer fixing until then.

> Anyways, full ELF compatibility is not on my work plan :-)

It’s a shame to make a change that gets most of the way to ABI-compliant layout and not do the rest.

wintersteiger:
Oh yes, you're right, the host-provided config should receive the same fix (if it has the same problem). They are currently not merged, which would have to be done if we decide to keep this path of providing the configuration to the enclave.

Re: ELF ABI: in my dev branch I removed the auxv, because it wasn't used, and I can't remember seeing any additional platform-specific stuff, is that available somewhere?

hukoyu:
@wintersteiger when this PR will be merged? I am waiting this PR to submit my PR for adding NodeJs test scenario. Are you waiting @prp 's review?

wintersteiger:
No, I wanted to take another look at those `sprintf`s. @davidchisnall I've replaced them with `snprintf`s now, so that we can't run past the end of the memory. I think this is an OK fix for now, until we replace it with the new configs.

letmaik:
@davidchisnall Could you check if this PR is OK now? It's blocking other PRs.

davidchisnall:
It's probably not a problem to merge.  It isn't where we want to be (with a proper initial stack layout on entry to the program), but I don't think it makes doing the right thing any harder.

wintersteiger:
Yeah, this is mainly a temporary fix to unblock others. All of this will change anyways, so there's no reason to fix the stack layout here and now.

src/enclave/enclave_init.c
davidchisnall: Please open a PR for this and mark this as resolved.

davidchisnall (approved):

src/enclave/enclave_init.c
wintersteiger: Which "this" is this? I assume you mean merge the current PR and open a new PR/issue for full ELF support?

hukoyu:
@wintersteiger  I thought you will merge this PR to unblock NodeJS apps and create a new PR for the remaining tasks/changes as @davidchisnall  recommended. Did the plan change? This PR still not merged.

wintersteiger:
No, the plan didn't change, I was just busy with other stuff.

src/enclave/enclave_init.c
```diff
@@ -41,24 +41,44 @@ static void __sgxlkl_enclave_copy_app_config(
...
     char** envp = NULL;
+    size_t total_size = 0;
 
     app_config->argc = sgxlkl_config->argc;
-    app_config->argv = oe_malloc((app_config->argc + 1) * sizeof(char*));
 
     for (i = 0; i < app_config->argc; i++)
-        app_config->argv[i] = oe_strdup(sgxlkl_config->argv[i]);
-    app_config->argv[i] = NULL;
+        total_size += strlen(sgxlkl_config->argv[i]) + 1;
```
SeanTAllen: we are removing libc dependencies from src/enclave

Can you change this to oe_strlen ?

Same about apply elsewhere.

src/enclave/enclave_init.c
```diff
@@ -41,24 +41,44 @@ static void __sgxlkl_enclave_copy_app_config(
...
+    char* buf = oe_malloc((total_size + 1) * sizeof(char*));
+    app_config->argv =
+        oe_malloc((sgxlkl_config->argc + envc + 2) * sizeof(char*));
+    size_t remaining = total_size + 1;
+
+    char* p = buf;
+    for (i = 0; i < app_config->argc; i++)
+    {
+        app_config->argv[i] = p;
+        p += snprintf(p, remaining, "%s", sgxlkl_config->argv[i]) + 1;
```
SeanTAllen: oe_snprintf please (there's a second one to change as well)

src/enclave/enclave_init.c
wintersteiger: Sure, no problem. This particular piece of code will likely vanish soon, but other bits (like the new json parser) will have to work on the host and in the enclave, so explicitly calling `oe_*` will be a problem on the host side. On my branch I added a compatibility header that `#defines` the normal names to the `oe_` names, which allows me to compile with and without that header.

wintersteiger: (done)

SeanTAllen: sounds good.

src/enclave/enclave_init.c
wintersteiger: (done)

wintersteiger:
Are those ltp test failures in the CI run the current "normal" or do they indicate that I broke something that wasn't already broken?

SeanTAllen:
@wintersteiger a couple look new. in particular the "sgx-lkl-oe" is still running. I've never seen that one.

there is something going on in CI that I haven't been able to reproduce locally of ltp tests occasionally timing out.

wintersteiger:
Thanks for the info @SeanTAllen! I've restarted the failed tasks to see whether anything changes. 

I've never seen the "still running" problem before either, but I suspect it's just another flavor of timeout.

hukoyu:
> @wintersteiger a couple look new. in particular the "sgx-lkl-oe" is still running. I've never seen that one.
> 
> there is something going on in CI that I haven't been able to reproduce locally of ltp tests occasionally timing out.

I also see random timeout failures in the LTP tests. These are regression. 
cc @SeanTAllen @wintersteiger

wintersteiger:
CI passed, so I'm merging.

PR #124, opened 2020-04-30
Fix condition for linking dcap lib

letmaik:
The sgx_dcap_ql library was linked depending on whether the host system has SGX and FLC support. This breaks if you want to build on a non-SGX machine. The right check is whether OE was built with DCAP enabled or not. This PR changes the detection logic.


prp (approved):
LGTM

PR #133, opened 2020-05-01
Remove musl-gcc dependency from tests

letmaik:
This PR prepares the tests to be run from an SGX-LKL installation instead of requiring a build tree by removing the dependency to musl-gcc and instead using Docker images.

Note that I used multi-stage Dockerfiles as otherwise the disk images would be quite large (due to the gcc and musl-dev packages).

Also, note that this PR does not intend to cleanup tests in any way. If that happened it is a coincidence.

cc @hukoyu


letmaik:
Could someone approve this? It blocks me from working on further improvements.

prp (approved):
LGTM

PR #137, opened 2020-05-01
Remove extraneous libc dependencies from lthread.c.

davidchisnall:
The dependencies that are part of the lack of separation between pthread
and lthread implementations are still there, and we still depend on
free-standing C functions (memset, memcpy, and so on), but this is
sufficient until we do the cleanup to decouple the pthread and lthread
implementations.

### DIFF
--- a/src/include/enclave/lthread.h
+++ b/src/include/enclave/lthread.h
@@ -181,7 +181,6 @@ struct lthread_sched
     void* stack;
     size_t stack_size;
     uint64_t default_timeout;
-    int page_size;
     /* convenience data maintained by lthread_resume */
     struct lthread* current_lthread;
 };
--- a/src/include/openenclave/corelibc/oemalloc.h
+++ b/src/include/openenclave/corelibc/oemalloc.h
@@ -1,0 +1,9 @@
+#ifndef __OE_MALLOC_INCLUDED__
+#define __OE_MALLOC_INCLUDED__
+
+void *oe_malloc(size_t size);
+void oe_free(void *ptr);
+void *oe_calloc(size_t nmemb, size_t size);
+void *oe_realloc(void *ptr, size_t size);
+
+#endif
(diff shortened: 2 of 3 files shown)


prp (approved):
LGTM. We should ensure that the OE heap is large enough for the additional allocations that we are now doing. Currently we are running with a small OE heap IIRC.

letmaik:
Don't we have 1 GiB currently?

prp:
> Don't we have 1 GiB currently?

No, it's 32 MB at the moment: https://github.com/lsds/sgx-lkl/blob/f0d4b2a54e45478d358a4c4ec922c81b8133446c/src/include/host/sgxlkl_params.h#L65

letmaik:
So there's LKL kernel memory size, which is 32M by default, then there's 1G heap size as defined in config/params.conf, and then there's `DEFAULT_SGXLKL_OE_HEAP_PAGE_COUNT`. What's the difference between params.conf and the latter?

prp:
> So there's LKL kernel memory size, which is 32M by default, then there's 1G heap size as defined in config/params.conf, and then there's `DEFAULT_SGXLKL_OE_HEAP_PAGE_COUNT`. What's the difference between params.conf and the latter?

`params.conf` specifies the full heap size that is allocated by OE; `DEFAULT_SGXLKL_OE_HEAP_PAGE_COUNT` reserves part of this heap for OE, so it can do its own allocations (through `oe_malloc()`) without interfering with the rest of the heap, which is now managed by the SGX-LKL enclave mmap allocator. The enclave mmap allocator gives 32 MB to LKL, which are then visible to the kernel. The rest of the heap memory can only be allocated by the application.

davidchisnall:
Note that this is a temporary limit.  Once OE gains pluggable allocator support (due in a month or so), we can provide an oe_malloc that sits on top of enclave_mmap, at which point we will be back down to two memory allocators.  At some point, we should figure out how to integrate our memory allocator with the kernel one...

Issue #139, opened 2020-05-01
The CI should also run tests with different ethread counts


hukoyu:
The CI should also run tests with different thread settings, like number of ethreads.
Pick one of the big application tests, e.g. JVM and run it with ethreads={1, 2, 3, 4, 8}.

SGXLKL_ETHREAD is used to pass thread count to sgx-lkl enclave

hukoyu:
This is addressed in this PR: https://github.com/lsds/sgx-lkl/pull/241

PR #140, opened 2020-05-01
SGXLKL_ETHREAD: Run JVM test with different ethread counts

hukoyu:
This PR addresses the issue: https://github.com/lsds/sgx-lkl/issues/139
ethread count is passed with SGXLKL_ETHREAD env var to sgx-lkl enclave. 
If this is not passed default value is the number of cores available.

### DIFF
--- a/tests/languages/java/Makefile
+++ b/tests/languages/java/Makefile
@@ -28,12 +28,28 @@ $(DISK_IMAGE): $(PROG)
 run: run-hw run-sw
 
 run-hw: $(DISK_IMAGE)
-	@echo "sgx-lkl-java --hw-debug ${DISK_IMAGE} HelloWorld"
-	@${SGXLKL_ENV} ${SGXLKL_JAVA_RUN} --hw-debug ${DISK_IMAGE} HelloWorld
+       @echo "SGXLKL_ETHREAD=1 sgx-lkl-java --hw-debug ${DISK_IMAGE} HelloWorld"
+       @SGXLKL_ETHREAD=1 ${SGXLKL_ENV} ${SGXLKL_JAVA_RUN} --hw-debug ${DISK_IMAGE} HelloWorld
+       @echo "SGXLKL_ETHREAD=2 sgx-lkl-java --hw-debug ${DISK_IMAGE} HelloWorld"
+       @SGXLKL_ETHREAD=2 ${SGXLKL_ENV} ${SGXLKL_JAVA_RUN} --hw-debug ${DISK_IMAGE} HelloWorld
+       @echo "SGXLKL_ETHREAD=3 sgx-lkl-java --hw-debug ${DISK_IMAGE} HelloWorld"
+       @SGXLKL_ETHREAD=3 ${SGXLKL_ENV} ${SGXLKL_JAVA_RUN} --hw-debug ${DISK_IMAGE} HelloWorld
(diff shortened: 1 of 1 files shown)


davidchisnall:
Thanks, this looks like really useful testing, but we probably want to do it in a slightly more generic way. Once #138 is merged, we will have a common include file for all of the tests, it would then be good to try to provide some generic infrastructure for running tests with multiple ethereal options.

It’s also not clear how failures are reported here. It looks as if this still appears to be a single test. It would be good if failures with 4 threads and passes with 2 threads (for example) were reported accurately.

hukoyu:
@davidchisnall running every test with different ethread counts will multiply tests and pipeline test run will be very long. @prp agreed that running JVM test with different ethread values were good coverage for now. You are right now we report run-hw and run-sw separately but different ethread runs will be reported together. If it fails, logs will clearly show which ETHREAD value failed. I don't see a value in reporting each ethread test result separately.

davidchisnall:
> @davidchisnall running every test with different ethread counts will multiply tests and pipeline test run will be very long. 

I agree, but that isn't an argument against having some generic infrastructure.  TensorFlow, OpenMP, and .NET would also benefit from running with different ethread counts.  Not necessarily on every commit or PR, but as part of a nightly test run.  

> @prp agreed that running JVM test with different ethread values were good coverage for now. 

I also agree, but not that we should implement it with custom logic in a single Makefile.

> You are right now we report run-hw and run-sw separately but different ethread runs will be reported together. If it fails, logs will clearly show which ETHREAD value failed. I don't see a value in reporting each ethread test result separately.

The more information that I can gain without having to read the logs, the better.  Having a report that the Java tests failed with 4 ethreads but passed with other values immediately tells me something is problematic with concurrency.  If one of the tests fails in the middle, then I will have to search in a long log full of working things to find it, whereas with a single failure I should be able to look directly at the log for the failed test.

If this were implemented with some generic infrastructure, with this Makefile picking up the ethread values from the environment provided by the test runner, then any job that we wanted to run with multiple tests would just be a separate invocation and would be reported separately.

hukoyu:
@davidchisnall just for the record you don't have to search in long logs. Now we add each test's log to its test result it is not a log file for all tests.

hukoyu:
@letmaik @davidchisnall I strongly believe we should merge this PR to have test coverage for different ETHREAD values. Creating an infrastructure for running tests with difrerent ETHREAD values I think out of scope of this PR.  You can create an issue for that and we can prioritize and work on that infra if needed.

letmaik:
@hukoyu I get your point, but I'd like to avoid creating a precedent. Let's take this as an opportunity to set up long-running "nightly" CI jobs. We need them anyway for testing things like Tensorflow which currently doesn't happen at all and is only part of the samples, not tests, due to its long build time.

Given that we can easily set SGXLKL_ETHREAD as an environment variable we don't even have to change any of the tests, we just have to set it globally at pipeline level. The only thing that's important is to add the number of ethreads somehow to the test labels. And of course figuring out whether we need to create a separate yaml file for the nightly pipeline or whether we can re-use the existing one and somehow parameterize it for scheduled runs. Do you want to have a look at that? I'm happy to have a look myself, just let me know.

letmaik:
I had a go myself, see #241. If everyone is happy with that I would close this PR.

hukoyu:
I aggree with @letmaik. PR #241 covers different ETHREAD combinations and will run daily. This PR not needed anymore.

PR #141, opened 2020-05-01
Fix incorrect timespec_diff comment

SeanTAllen:


davidchisnall:
I think the comment now reflects the code, which is an improvement, but it doesn’t alter the fact that the code is deeply confusing. I think that the correct fix is to swap the names of the two arguments and have a comment include something like:

```
diff = endtime - starttime
```
Expecting the end to be before the star is odd. I am not very good at time but I am reliably informed that most humans expect the past to happen before the future.

SeanTAllen:
I agree. I would worry about switching the arguments given they are the same type and ending up with bugs. We could update all callers at the same time but that doesn't account for new callers on branches, etc.

davidchisnall:
I doubt there are new callers on branches.  Having a function with arguments in obviously the wrong order seems more likely to cause bugs in new code.  That said, the fix that I'd actually like to see is removing the indirection and making it:

```c
struct timespec timespec_diff(struct timespec end, struct timespec start);
```

Fixing it like that would cause argument type mismatches with out-of-tree callers and force them to update their code before they were able to build it.

letmaik:
Could we get this merged or abandoned? I'm not a big fan of stale PRs :-)

PR #145, opened 2020-05-03
[CI] Remove *.lock files in .git folder before submodule update

letmaik:
I saw this once in CI where a VM was essentially "broken" because there was a lock file from git which prevented running some git commands. This file was left-over from a previous build for unknown reasons. This is not an issue when fresh VMs are used for each build but this is not the case (yet).
This PR simply cleans up old git lock files. Note that I tested this PR when the situation occurred and it resolved the block.

### DIFF
--- a/.azure-pipelines/scripts/checkout_submodules.sh
+++ b/.azure-pipelines/scripts/checkout_submodules.sh
@@ -1,8 +1,11 @@
 #!/bin/bash
 
+set -ex
+
 mkdir -p ~/.ssh > /dev/null
 chmod 700 ~/.ssh
 ssh-keyscan -t rsa github.com >> ~/.ssh/known_hosts
+find .git -type f -name '*.lock' -exec rm -v {} +
 git reset --hard
 git submodule sync
 git submodule update --init --force --recursive


.azure-pipelines/scripts/checkout_submodules.sh
```diff
@@ -1,8 +1,11 @@
 #!/bin/bash
 
+set -ex
+
 mkdir -p ~/.ssh > /dev/null
 chmod 700 ~/.ssh
 ssh-keyscan -t rsa github.com >> ~/.ssh/known_hosts
+find .git -type f -name '*.lock' -exec rm -v {} +
```
prp, line 8: Would it also be a good idea to kill any leftover git processes on the machine?

prp (approved):
LGTM. I left one comment.

.azure-pipelines/scripts/checkout_submodules.sh
letmaik: No, in my experience git processes never "hang" or similar, so there is no need to check if any are still running. But if something goes bad then they may leave lock files behind.

PR #149, opened 2020-05-04
Fix for umount01

ashokchandran:
This PR addresses issue 67 in LTP tests:
Issue Description: Currently the test tries to test the umount api. For this it was using a test device using loop device and make a filesystem and then unmount. But this was failing due to the memory limit of SGX at 32 MB.
Solution: With this PR we are mounting the root file system and verifying the mount functionality.
Reference:https://dev.azure.com/ConfidentialContainers/ConfidentialLinuxContainers/_workitems/edit/67/


hukoyu (approved):
It looks good to me. @letmaik  can you also review and merge if looks good. Thanks.

letmaik:
Can you merge latest oe_port?

ashokchandran:
tracked in https://github.com/lsds/sgx-lkl/pull/201

Issue #12, opened 2018-09-22
What is `src/libsgx/libs/{lible.a, lible.so}`?


mssun:
I guess it is the launch enclave, but I didn't find any source code for generating the `lible`.

cpriebe:
Hi Bob,

thanks for opening this issue! They are just wrappers around (an older version of) the Intel-provided launch enclave. We should switch to simply use the LE that comes with the Intel PSW. We'll look into that and push a corresponding change soon.

mssun:
Got it. Thank you!
-- 
Best,
Mingshen
www.cse.cuhk.edu.hk/~mssun

Issue #11, opened 2018-08-13
SGX-LKL documentation (paper?)


zavalyshyn:
Hi,

I've tried sgx-lkl (jvm example) and it seems to run properly. There are few things that I'd like to clarify though. 

First of all, is there any documentation available for this software? The Wiki section leads to the repo front page. There is no link to any documentation in the README file. Was there any paper published on sgx-lkl?

I mainly look for documentation to clarify my other concerns. 

- For instance, how I can verify that the HelloWorld example is indeed running in the SGX enclave? 
- What are the security guarantees SGX-LKL provides? 
- How can I prove those?

Finally, seems like the developers of SGX-LKL come from the same group that worked on SCONE, which I also tested previously. What system is more actively developed and should be thus used? 

Thanks

prp:
Hi,

Since SGX-LKL is an ongoing research project, we unfortunately don't have documentation beyond the README file and the source code. We are currently in the process of writing a research paper about it.

Regarding your questions, all of the source code of SGX-LKL is available, and you can verify that the code indeed is launched inside of the enclave. In general, the security guarantees are similar to what Graphene-SGX or SCONE provide. In ongoing work, we aim to give stronger guarantees regarding possible data leakage across the interface to the host OS.

SGX-LKL is an early fork of SCONE, done at the time before the SCONE paper was published, and has since evolved independently. SCONE is a separate commercial offering.

zavalyshyn:
Great. Thanks for your explanation. I will stick to SCONE then.

kanakraju73:
Any update on this?

Issue #10, opened 2018-05-08
New hello world sample fails


lshacham:
**Summary**: new hellow world sample fails
**Details**: return value is 255
I run: 
`sgx-lkl/apps/helloworld$ SGXLKL_TRACE_HOST_SYSCALL=1 SGXLKL_TRACE_LKL_SYSCALL=1 SGXLKL_HD_RW=1 ../../build/sgx-lkl-run sgxlkl-disk.img  /app/helloworld | grep -v clock_gettime `
And get: 
```
[ HOST SYSCALL ] [tid=???] mmap	9	(0, 4096, 3, 34, 0, 0) = 139642381250560
[ HOST SYSCALL ] [tid=???] mprotect	10	(0, 4096, 0, 34, 0, 0) = 0
[ HOST SYSCALL ] [tid=???] mprotect	10	(84160512, 524288, 3, 34, 0, 0) = 0
[ HOST SYSCALL ] [tid=???] mprotect	10	(84156416, 4096, 3, 34, 0, 0) = 0
[ HOST SYSCALL ] [tid=???] mprotect	10	(9969664, 4096, 1, 34, 0, 0) = 0
[ HOST SYSCALL ] [tid=???] mprotect	10	(84156416, 4096, 3, 34, 0, 0) = 0
[ HOST SYSCALL ] [tid=???] mmap	9	(0, 4096, 3, 34, 0, 0) = 139642381246464
[ HOST SYSCALL ] [tid=???] lseek	8	(5, 0, 2, 0, 0, 0) = 104857600
[ HOST SYSCALL ] [tid=???] mprotect	10	(83632128, 524288, 3, 0, 0, 0) = 0
[ HOST SYSCALL ] [tid=???] mprotect	10	(83628032, 4096, 3, 0, 0, 0) = 0
[ HOST SYSCALL ] [tid=???] mprotect	10	(9969664, 4096, 1, 0, 0, 0) = 0
[ HOST SYSCALL ] [tid=???] mprotect	10	(83628032, 4096, 3, 0, 0, 0) = 0
[ HOST SYSCALL ] [tid=???] mmap	9	(0, 4096, 3, 34, 0, 0) = 139642381242368
[ HOST SYSCALL ] [tid=???] mprotect	10	(71041024, 12587008, 3, 0, 0, 0) = 0
[ HOST SYSCALL ] [tid=???] mprotect	10	(71036928, 4096, 3, 0, 0, 0) = 0
[...]

PR #152, opened 2020-05-04
Fix for umount02

ashokchandran:
Issue Description: This test was failing with panic due to lack of space. This was a limitation in SGX with 32MB memory.
Solution: The test now uses root file system


hukoyu (approved):
I already reviewed this PR and @ashokchandran addressed my feedback in private repo. This PR is resubmitted to public repo. It looks good to me. @letmaik can you also review and merge if looks good. Thanks.

tests/ltp/buildenv.sh
```diff
@@ -45,6 +46,10 @@ if [[ "$mode" == "build" ]]; then
         git apply $FORK_DISABLE_PATCH
     fi
 
+    if [ -f $UMOUNT02_PATCH ];then
```
letmaik: Why would the patch file be missing?

tests/ltp/umount02.patch
```diff
@@ -0,0 +1,81 @@
...
++const char* opts = "mode=0777";  
+ 
+ static char long_path[PATH_MAX + 2];
+ static int mount_flag;
+@@ -31,7 +36,7 @@ static struct tcase {
+ 	int exp_errno;
+ } tcases[] = {
+ 	{"Already mounted/busy", MNTPOINT, EBUSY},
+-	{"Invalid address", NULL, EFAULT},
++/*	{"Invalid address", NULL, EFAULT}, This test is commented because in SGX accessing NULL results in enclave exit */
```
letmaik: Is that a bug in SGX-LKL? Should it be recorded in an issue?

tests/ltp/umount02.patch
```diff
@@ -0,0 +1,81 @@
...
+ 
+ 	if (TST_RET != -1) {
+-		tst_res(TFAIL, "umount() succeeds unexpectedly");
++		tst_res(TFAIL, "umount() succeeds unexpectedly. TEST_FAILED");
+ 		return;
+ 	}
+ 
+ 	if (tc->exp_errno != TST_ERR) {
+-		tst_res(TFAIL | TTERRNO, "umount() should fail with %s",
++		tst_res(TFAIL | TTERRNO, "umount() should fail with %s. TEST_FAILED",
```
letmaik: Why is this needed? Is the exit code of the test not enough?

tests/ltp/umount02.patch
```diff
@@ -0,0 +1,81 @@
...
++
++	const char* src  = "/dev/vda";
++	const char* type = "ext4";
++	const unsigned long mntflags = 0;
++	const char* opts = "mode=0777";  
+ 	memset(long_path, 'a', PATH_MAX + 1);
+ 
+ 	SAFE_MKDIR(MNTPOINT, 0775);
+-	SAFE_MOUNT(tst_device->dev, MNTPOINT, tst_device->fs_type, 0, NULL);
++/*	SAFE_MOUNT(tst_device->dev, MNTPOINT, tst_device->fs_type, 0, NULL);*/
```
letmaik: Use `//` when only commenting out a single line. And maybe also add a comment why the patch to mount is necessary.

tests/ltp/umount02.patch
```diff
@@ -0,0 +1,81 @@
+diff --git a/testcases/kernel/syscalls/umount/umount02.c b/testcases/kernel/syscalls/umount/umount02.c
+index cd67aa6d2..8822eab89 100644
+--- a/testcases/kernel/syscalls/umount/umount02.c
++++ b/testcases/kernel/syscalls/umount/umount02.c
+@@ -20,6 +20,11 @@
+ #include "tst_test.h"
+ 
+ #define MNTPOINT	"mntpoint"
++const char* src  = "/dev/vda";
```
letmaik: Add a comment somewhere that explains why this patch is necessary so that it can be understood in the future.

letmaik (changes requested):

tests/ltp/buildenv.sh
ashokchandran: Chance is less but if the patch file is not copied to the proper path we can find it here during the build.

letmaik: Sorry I don't understand, could you explain it a bit more?

tests/ltp/umount02.patch
ashokchandran: This was for a pipe line script to find result as per Huseyn.

tests/ltp/umount02.patch
ashokchandran: Yes  will raise this as an issue..  in umount02 when it test umount with null, it causes enclave to abort. Normally application is expected to get EFAULT. But the back trace is showing that in lkl it calls oe_abort when address is invalid.
 
#0  lkl_access_ok (addr=0, size=1) at lkl/setup.c:1499
#1  0x00007fff0010aaff in __lkl_access_ok (size=<optimized out>, addr=<optimized out>) at ./arch/lkl/include/asm/uaccess.h:14
#2  strncpy_from_user (count=<optimized out>, src=<optimized out>, dst=<optimized out>) at ./include/asm-generic/uaccess.h:233
#3  getname_flags (filename=0x0, flags=0, empty=0x0) at fs/namei.c:149
#4  0x00007fff0010ad2d in getname (filename=<optimized out>) at fs/namei.c:209
#5  0x00007fff0010da1d in user_path_mountpoint_at (dfd=0, name=<optimized out>, flags=1049890816, path=0x7fff3e941020)
    at fs/namei.c:2745
#6  0x00007fff0012003d in ksys_umount (name=0x7fff3e941000 " \020\224>\377\177", flags=0) at fs/namespace.c:1680
#7  0x00007fff00120479 in __do_sys_umount (flags=<optimized out>, name=<optimized out>) at fs/namespace.c:1706
#8  __se_sys_umount (name=<optimized out>, flags=<optimized out>) at fs/namespace.c:1704
#9  0x00007fff00088076 in run_syscall (params=<optimized out>, no=<optimized out>) at arch/lkl/kernel/syscalls.c:44
#10 lkl_syscall (no=140734243278880, params=0x0) at arch/lkl/kernel/syscalls.c:183
#11 0x00007fff0054901a in __filter_syscall2 (n=39, a1=0, a2=0) at ./src/internal/syscall.h:161
#12 0x00007fff0054924a in umount (special=0x0) at src/linux/mount.c:11
    if (!ret)
    {
        sgxlkl_fail("lkl_access_check failed: %p\n", (void*)addr);
    }
    return ret;

tests/ltp/umount02.patch
letmaik: @hukoyu Can you comment?

tests/ltp/buildenv.sh
ashokchandran: I meant copying patch to mount point is done in Makefile, So thought its better to check here that is there. 
vamsi@vm-gshanker:~/ashok_latest/fork3/sgx-lkl/tests/ltp$ grep patch Makefile
        $(ESCALATE_CMD) cp ltp_fork_disable.patch $(MOUNTPOINT)/
        $(ESCALATE_CMD) cp patches/umount03.patch $(MOUNTPOINT)/

tests/ltp/umount02.patch
ashokchandran: will do

tests/ltp/umount02.patch
ashokchandran: will do

tests/ltp/buildenv.sh
letmaik: So if the patch is not there then you don't apply it and continue happily as if everything is ok? Doesn't make much sense to me. Or am I missing something?

ashokchandran: Agree that it is not really required. Will remove the check.

ashokchandran:
@letmaik  please review and approve if all ok now.

letmaik (approved):

PR #159, opened 2020-05-04
Split CI into build/test/publish

letmaik:
- Build & sw tests are run on non-SGX VM.
- Only hw tests are run on SGX VM.
- LKL tests are now also running in sw mode.
- All test variants (run mode, build mode, suite) are separate jobs and run in parallel.
- Removed old "Debian" pipeline in favour of a single one.
- Added new CC test, previously hidden in test_standalone_deb_pkg.sh.
- Added `--mode=debug|nondebug` to `sgx-lkl-docker build-cc`.
- Other minor fixes, cleanup and improvements.

Note that this PR does not change the existing build modes. There's still debug and nondebug. Any change to that is a future PR.

Also, note that an individual LTP test job only takes 3-5min. It may not be worth having separate ltp1 and ltp2 test suites going forward. Separate PR though.

cc @hukoyu


davidchisnall:
Shiny!  We should probably think about splitting the non-LTP tests in a future PR.

davidchisnall (approved):

PR #160, opened 2020-05-04
LTP: Fix failure in flock syscall test flock03

shaikshavali1:
Issue No: 214
Issue Description:
Test case is waiting for the parent process to wake up check point and encountered timeout error after 10000ms.

Solution:
sgxlkl is a single process environment so modified the test case to use posix threads.


hukoyu (approved):
Looks good to me. @letmaik can you also review and merge if looks good? Thanks.

tests/ltp/buildenv.sh
```diff
@@ -45,6 +46,11 @@ if [[ "$mode" == "build" ]]; then
         git apply $FORK_DISABLE_PATCH
     fi
 
+    if [ -f $FIX_FLOCK_PATCH ];then
```
letmaik: Is this check necessary?

letmaik (changes requested):

tests/ltp/buildenv.sh
shaikshavali1: Removed the check and updated the buildenv.sh file

letmaik (approved):

PR #161, opened 2020-05-04
Enclave pie

davidchisnall:
Compile our enclave code and LKL with -fPIE.  This generates position-independent code without GOT relocations where possible, so most things will work without dynamic relocations.  The Linux kernel contains a number of weak reference (529), which still need relocations.


letmaik:
What is this fixing?

davidchisnall:
The need to run musl code to relocate the enclave before we run anything.

prp (approved):
LGTM

PR #162, opened 2020-05-04
In enclave time source

SeanTAllen:
@davidchisnall per our conversation from Friday. Here's the PR with ugliness to get feedback. 

The ltp tests are passing. This includes your lthread fix you pointed out on Friday. There's some TO-DO's I left for myself to clean up and there's another test I want to add to make sure that the system clock gets set on startup.

I'm going to start in on my clean up and tests and will address your feedback as it comes in.

Not seen in these changes but required: removing the sgx-lkl-musl overrides of SYS_clock_gettime and SYS_futex.


docs/HostInterface.md
```diff
@@ -154,7 +155,35 @@ Note that the guest cannot guarantee anything about the amount of time that this
...
+```
+
+`timer_dev` has a field `nanos` that indicates the number of monotonic nanoseconds that have passed since startup.
+An instance of `timer_dev` is initialized on startup and updated periodically from outside the enclave to reflect the passage of time.
+`nanos` can be accessed from within the enclave to create a monotonic source of time.
+
+Inside the enclave, an internal "source of truth" for monotonic passage of time is also kept.
+When an enclave caller needs access to the latest monotonic time, the following general logic occurs:
+
+- If monotonic time outside the enclave is greater than monotonic time inside the enclave, then internal time is set to external time.
```
davidchisnall: A bulleted list with a single bullet is odd...

src/enclave/sgxlkl_config.c
```diff
@@ -55,7 +55,7 @@ int sgxlkl_copy_config(const sgxlkl_config_t* from, sgxlkl_config_t** to)
     sgxlkl_config_t* cfg = *to;
 
     // Catch modifications to sgxlkl_config_t early.
-    SGXLKL_ASSERT(sizeof(sgxlkl_config_t) == 464);
+    SGXLKL_ASSERT(sizeof(sgxlkl_config_t) == 472);
```
davidchisnall: Can this be `_Static_assert`?  We may need to tweak the makefile to build as c11 instead of c99.

src/include/lkl/syscall-overrides-futex.h
```diff
@@ -0,0 +1,14 @@
+#ifndef _LKL_SYSCALL_OVERRIDES_FUTEX_H
+#define _LKL_SYSCALL_OVERRIDES_FUTEX_H
+
+#include <time.h>
```
davidchisnall: Please can you make sure that this is an LKL header and not a musl one?

src/include/enclave/enclave_mem.h
```diff
@@ -42,7 +42,15 @@ void* syscall_SYS_mremap(
...
-int syscall_SYS_futex(
+int syscall_SYS_enclave_futex(
+    int* uaddr,
+    int op,
+    int val,
+    const struct timespec* timeout,
+    int* uaddr2,
+    int val3);
+
+long syscall_SYS_futex_override(
```
davidchisnall: Why is this declared in two headers?

src/include/openenclave/corelibc/oemalloc.h
```diff
@@ -0,0 +1,9 @@
+#ifndef __OE_MALLOC_INCLUDED__
```
davidchisnall: This file is in oe_port, I'm not sure why it's showing up here...

src/include/sched/futex.h
```diff
@@ -0,0 +1,14 @@
+#ifndef _SCHED_FUTEX_H
+#define _SCHED_FUTEX_H
+
+#include <time.h>
+
+int syscall_SYS_enclave_futex(
```
davidchisnall: Since we're restructuring things, this should probably just be called `enclave_futex`

src/lkl/setup.c
```diff
@@ -1400,13 +1441,17 @@ void lkl_start_init()
     SGXLKL_VERBOSE("creating LKL termination thread\n");
     create_lkl_termination_thread();
 
+    register_lkl_syscall_overrides();
```
davidchisnall: Why are we doing this twice?

src/lkl/syscall-overrides-futex.c
```diff
@@ -0,0 +1,69 @@
...
+    if ((op & FUTEX_WAIT_BITSET) && timeout != NULL)
+    {
+        // adjust absolute timeout to a relative one
+        clock_t clock =
+            op & FUTEX_CLOCK_REALTIME ? CLOCK_REALTIME : CLOCK_MONOTONIC;
+
+        struct lkl_timespec now;
+        if (lkl_sys_clock_gettime(clock, &now) != 0)
+        {
+            // SEAN-TODO: fail here
```
davidchisnall: `sgxlkl_fail` is fine here.  The world is broken, give up.

src/sched/lthread.c
```diff
@@ -403,7 +394,7 @@ void _lthread_free(struct lthread* lt)
         lthread_rundestructors(lt);
     if (lt->itls != 0)
     {
-        munmap(lt->itls, lt->itlssz);
+        enclave_munmap(lt->itls, lt->itlssz);
```
davidchisnall: These look like upstream changes too. Can you rebase?

davidchisnall (commented):
Looks mostly fine.  Some small tweaks and please rebase.

src/lkl/setup.c
SeanTAllen: left over from my "what is going on here" debugging. it will be removed.

src/sched/lthread.c
SeanTAllen: yup

src/lkl/syscall-overrides-futex.c
SeanTAllen: yup

src/include/sched/futex.h
SeanTAllen: yeah, renaming is part of the cleanup.

src/include/openenclave/corelibc/oemalloc.h
SeanTAllen: it came in as part of rebasing on your now closed PR.

src/enclave/sgxlkl_config.c
SeanTAllen: That is mostly part of the config attestation work, I was trying to step on it as little as possible. @wintersteiger any issue with my making that change?

wintersteiger: Not at all, this is intended to be a static check. Definitely don't tweak any Makefiles just for this.

src/include/enclave/enclave_mem.h
SeanTAllen: left over. gone.

docs/HostInterface.md
```diff
@@ -154,7 +155,33 @@ Note that the guest cannot guarantee anything about the amount of time that this
...
+    uint64_t init_walltime_nsec;
+};
+```
+
+`timer_dev` has a field `nanos` that indicates the number of monotonic nanoseconds that have passed since startup.
+An instance of `timer_dev` is initialized on startup and updated periodically from outside the enclave to reflect the passage of time.
+`nanos` can be accessed from within the enclave to create a monotonic source of time.
+
+Inside the enclave, an internal "source of truth" for monotonic passage of time is also kept.
+When an enclave caller needs access to the latest monotonic time, the following general logic occurs if monotonic time outside the enclave is greater than monotonic time inside the enclave, then internal time is set to external time, otherwise, internal time is increased and returned.
```
jxyang, line 176: Not clear to me how much the `internal time` is increased.

docs/HostInterface.md
```diff
@@ -154,7 +155,33 @@ Note that the guest cannot guarantee anything about the amount of time that this
...
+
+Wallclock time within the enclave is provided by setting the wallclock on enclave startup to a time approximately the same as the host.
+Wallclock time inside the enclave can then move based on the corresponding movement of the monotonic nanoseconds as described earlier in this section.
+`init_walltime_sec` and `init_walltime_nsec` are used to convey **on startup** the hosts perceived wallclock time.
+Note, with the current version of this interface, it should be assumed that `init_walltime_*` fields are set only once.
+Software running inside the enclave can use standard system calls such as `clock_settime` to change the wallclock time.
+So for example, NTP would be able to function within the enclave.
+However, the `init_walltime_*` fields are not updated by changes to wallclock time within the enclave.
```
jxyang, line 184: Not clear to me how would `clock_settime` work without changing `init_walltime_*` fields. Are we changing `nanos` instead?

src/lkl/syscall-overrides-futex.c
```diff
@@ -0,0 +1,72 @@
...
+ * If endtime < starttime, difference will be considered to be "0".
+ */
+static void timespec_diff(
+    const struct timespec* starttime,
+    const struct timespec* endtime,
+    struct timespec* diff)
+{
+    if (starttime->tv_sec > endtime->tv_sec ||
+        (starttime->tv_sec == endtime->tv_sec &&
+         starttime->tv_nsec >= endtime->tv_nsec))
```
jxyang, line 19: Is it easier to calculate and diff the values of `(tv_sec * 1000000 + tv_nsec)`, and return 0 for negative results?

src/lkl/posix-host.c
```diff
@@ -612,7 +610,7 @@ struct lkl_host_operations sgxlkl_host_ops = {
     .tls_free = tls_free,
     .tls_set = tls_set,
     .tls_get = tls_get,
-    .time = time_ns,
+    .time = enclave_nanos,
```
jxyang, line 605: The system `time` function should produces the seconds elapsed since epoch. With the current implementation, the `time` function returns the seconds since we launch sgx-lkl.

jxyang (dismissed):

docs/HostInterface.md
SeanTAllen: the documentation here is limited to the host/enclave interface. the how is documented with the code.

src/lkl/posix-host.c
SeanTAllen: What current implementation?

SeanTAllen: With these changes, time(NULL) returns seconds since epoch, not seconds since the launch of sgx-lkl.

davidchisnall: This does not matter.  The time function is a time source provided to Linux's hrtimer infrastructure.  It starts at an arbitrary point and increments monotonically, modelling a hardware time source.  The kernel records the initial value as the offset for uptime and a different (mutable) offset to use for the wall-clock time.  The fact that the old time source started at an arbitrary point in the past and this one starts at a different arbitrary point in the past does not matter to the consumers.

docs/HostInterface.md
davidchisnall: `clock_settime` is a Linux system call that adjusts the kernel's internal idea of the displacement between the hardware time source and a specific time counter (for example, the real-time clock is implemented as the difference between the hardware time source and the UNIX epoch time).  This is the interface is for the hardware time source.  It is completely independent of the any of the clocks that the kernel chooses to model on top of it.

src/lkl/setup.c
```diff
@@ -31,6 +42,7 @@
 #include "enclave/sgxlkl_t.h"
 #include "enclave/wireguard.h"
 #include "enclave/wireguard_util.h"
+#include "pthread.h"
```
davidchisnall: What is this for?  We shouldn't need pthreads here...

SeanTAllen: Not sure. I don't remember adding it. I'll remove.

davidchisnall (approved):

src/lkl/syscall-overrides-futex.c
davidchisnall: Doing that without hitting undefined behaviour from signed integer overflow / underflow is nontrivial.  If you want to try it, please submit a separate PR.

PR #174, opened 2020-05-04
Remove obsolete test_miniroot.sh

### LINKED ISSUE #163: test_miniroot.sh
.azure-pipelines/scripts/test_miniroot.sh is broken and not invoked from anywhere. Should this test be made into a proper test and moved to tests/ or is it obsolete?

letmaik:
Closes #163. 
This test (whenever it did work in the past) checked whether `whoami` returns `root`. Once the user becomes configurable in the app config we should re-add a similar test.


prp (approved):
LGTM

PR #176, opened 2020-05-05
Test networking in CC test

letmaik:


davidchisnall:
Is this likely to hit the same intermittent failures as the curl test?

letmaik:
Probably. At least then we know it's not curl specific.

prp (approved):
LGTM

PR #192, opened 2020-05-05
Fix for mount issues

ashokchandran:
Issue: Mount related test cases in mount folder of ltp were failing. 
Description: Test cases were failing as there was not sufficient memory to use for test image creation for a loop device for test purpose. 
Solution: Test cases are modified to use /dev/vda the root file system device. In mount02 test accessing fault address is commented. This is a known issue in sgx that it causes enclave abort and is being tracked in Issue 169. mount03 issue will be addressed later as it has other problems.


tests/ltp/patches/mount.patch
```diff
@@ -0,0 +1,409 @@
...
+--- a/testcases/kernel/syscalls/mount/mount01.c
++++ b/testcases/kernel/syscalls/mount/mount01.c
+@@ -36,8 +36,13 @@ int TST_TOTAL = 1;
+ #define DIR_MODE (S_IRWXU | S_IRUSR | S_IXUSR | S_IRGRP | S_IXGRP)
+ #define MNTPOINT "mntpoint"
+ 
+-static const char *device;
+-static const char *fs_type;
++/*
++	patch to use root file system as sgx has 32 MB limit in using in
```
letmaik, line 12: This comment is not accurate, it's not a limitation of SGX but rather our choice of setting the default LKL kernel memory to 32M.

tests/ltp/patches/mount.patch
```diff
@@ -0,0 +1,409 @@
...
++static const char *fs_type = "ext4";
+ 
+ int main(int ac, char **av)
+ {
+@@ -54,13 +59,13 @@ int main(int ac, char **av)
+ 		TEST(mount(device, MNTPOINT, fs_type, 0, NULL));
+ 
+ 		if (TEST_RETURN != 0) {
+-			tst_resm(TFAIL | TTERRNO, "mount(2) failed");
++			tst_resm(TFAIL | TTERRNO, "mount(2) TEST_FAILED");
```
letmaik, line 26: Is this really necessary? We cannot patch all LTP tests like that. @hukoyu

letmaik (changes requested):
I think this needs some work, in particular I'm worried of patching the success/failure console outputs, which doesn't seem sustainable. It shouldn't be necessary.

ashokchandran:
tracked in https://github.com/lsds/sgx-lkl/pull/201

PR #199, opened 2020-05-06
Remove sgx-lkl-docker build-sgxlkl

letmaik:
Building SGX-LKL inside Docker and using the binaries on the host as-is is brittle and not something we should support. We may re-introduce a fully reproducible Docker build in the future which would produce a self-contained Debian package, but until then this has to go. Also, sgx-lkl-docker is installed and "building SGX-LKL" doesn't make sense in the context of an installation.


prp (approved):
LGTM

PR #206, opened 2020-05-06
LTP:Fixed issue setfsuid syscall test setfsuid04

vamsikrishna935:
Testcase met success criteria but due to lack of child process, it is
not reporting test status, so replaced exit with tst_resm to report
the status and removed wait for child in parent.

### DIFF
--- a/tests/ltp/ltp-batch2/ltp_disabled_tests.txt
+++ b/tests/ltp/ltp-batch2/ltp_disabled_tests.txt
@@ -822,7 +822,7 @@
 #/ltp/testcases/kernel/syscalls/setfsuid/setfsuid01
 #/ltp/testcases/kernel/syscalls/setfsuid/setfsuid02
 /ltp/testcases/kernel/syscalls/setfsuid/setfsuid03
-/ltp/testcases/kernel/syscalls/setfsuid/setfsuid04
+#/ltp/testcases/kernel/syscalls/setfsuid/setfsuid04
 #/ltp/testcases/kernel/syscalls/setgid/setgid01
 /ltp/testcases/kernel/syscalls/setgid/setgid02
 #/ltp/testcases/kernel/syscalls/setgid/setgid03
--- a/tests/ltp/patches/setfsuid04.patch
+++ b/tests/ltp/patches/setfsuid04.patch
@@ -1,0 +1,74 @@
+diff --git a/testcases/kernel/syscalls/setfsuid/setfsuid04.c b/testcases/kernel/syscalls/setfsuid/setfsuid04.c
+index 8585d6207..75621d652 100644
+--- a/testcases/kernel/syscalls/setfsuid/setfsuid04.c
++++ b/testcases/kernel/syscalls/setfsuid/setfsuid04.c
+@@ -25,7 +25,7 @@
+  * The same test is done in a fork to check if new UIDs are correctly
+  * passed to the son.
+  */
+-
++// Patch Description : Modified to report PASS/FAIL status by replacing exit, as child process is not supported.
+ #include <sys/types.h>
+ #include <sys/stat.h>
+ #include <sys/wait.h>
+@@ -69,7 +69,7 @@ int main(int ac, char **av)
+        if (pid == 0)
+                do_master_child();
+
+-       tst_record_childstatus(cleanup, pid);
++       //tst_record_childstatus(cleanup, pid);
+
+        cleanup();
+        tst_exit();
+@@ -92,15 +92,13 @@ static void do_master_child(void)
+
+        if (TEST_RETURN != -1) {
+                close(TEST_RETURN);
+-               printf("open succeeded unexpectedly\n");
+-               exit(TFAIL);
++               tst_resm(TFAIL, "open succeeded unexpectedly TEST_FAILED\n");
+        }
+
+        if (TEST_ERRNO == EACCES) {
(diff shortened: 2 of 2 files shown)


letmaik:
Can you explain why exit() doesn't work?

vamsikrishna935:
> 
> 
> Can you explain why exit() doesn't work?

exit() will return the status to parent process here we don't have 2 process, so in the single process environment it is exiting improperly, so modified the tests to work for single process environment.

letmaik:
I see, but what does "exiting improperly" mean?

vamsikrishna935:
> 
> 
> I see, but what does "exiting improperly" mean?

The status reporting from exit is not happening.

letmaik:
What does that mean? There's a printf before the exit, and the exit with non-zero code should mark the test as failed. What am I missing?

vamsikrishna935:
> 
> 
> What does that mean? There's a printf before the exit, and the exit with non-zero code should mark the test as failed. What am I missing?

Printf statement will print the message, usually exit() method sends the status to parent process, but we have only one process so the status from exit will not be reported. Because of this test result is showing as skipped. So as we cannot report the status from printf and in order to represent test status , i changed it to tst_resm.

vamsikrishna935:
> 
> 
> > What does that mean? There's a printf before the exit, and the exit with non-zero code should mark the test as failed. What am I missing?
> 
> Printf statement will print the message, usually exit() method sends the status to parent process, but we have only one process so the status from exit will not be reported. Because of this test result is showing as skipped. So as we cannot report the status from printf and in order to represent test status , i changed it to tst_resm.

Result of this testcase from CI Run : 'kernel-syscalls-setfsuid-setfsuid04' **skipped**. Failure Count = 0, Pass Count = 0.

Our reporting mechanism uses Pass Count and Failure count to report the test status as passed / failed. Here both counts are zero , so it will be marked as skipped. So i have modified it to get pass/fail counts.

letmaik:
So do I understand correctly that you do not look at the exit code at all but instead only scan for keywords in the output?

hukoyu:
@vamsikrishna935  if you are talking about run_ltp_test.sh (https://github.com/lsds/sgx-lkl/blob/oe_port/tests/ltp/run_ltp_test.sh) there is no more skipped tests. There are failures or passes.

vamsikrishna935:
> 
> 
> So do I understand correctly that you do not look at the exit code at all but instead only scan for keywords in the output?

The reporting framework is designed to look at the output and it is generalized for all the tests.

> 
> 
> @vamsikrishna935 if you are talking about run_ltp_test.sh (https://github.com/lsds/sgx-lkl/blob/oe_port/tests/ltp/run_ltp_test.sh) there is no more skipped tests. There are failures or passes.

But we need pass count/ fail count to report the status right ?

letmaik:
> > So do I understand correctly that you do not look at the exit code at all but instead only scan for keywords in the output?
> 
> The reporting framework is designed to look at the output and it is generalized for all the tests.

If it was general then you wouldn't have to patch output strings in the test cases. Why can't we additionally look at the exit code as well to avoid patching strings?

hukoyu:
> > > So do I understand correctly that you do not look at the exit code at all but instead only scan for keywords in the output?
> > 
> > 
> > The reporting framework is designed to look at the output and it is generalized for all the tests.
> 
> If it was general then you wouldn't have to patch output strings in the test cases. Why can't we additionally look at the exit code as well to avoid patching strings?

@letmaik current LTP test runner just looks at pass failure keywords and doesn't look at the exit code. My guess is relying on exit code for these tests doesn't work in all cases like this one since there is child process/parent process cases. I strongly believe using exit code is out of scope of enabling these test cases. Let's stick with current architecture and logic and I will try and submit a PR on run_ltp_test.sh using exit code and will see if we will have test failures due to that,

hukoyu:
@letmaik with current test pass and fail logic I asked PASS/TEST_FAILED outputs in the tests from PR createors.

vamsikrishna935:
> 
> 
> > > So do I understand correctly that you do not look at the exit code at all but instead only scan for keywords in the output?
> > 
> > 
> > The reporting framework is designed to look at the output and it is generalized for all the tests.
> 
> If it was general then you wouldn't have to patch output strings in the test cases. Why can't we additionally look at the exit code as well to avoid patching strings?


> 
> 
> > > So do I understand correctly that you do not look at the exit code at all but instead only scan for keywords in the output?
> > 
> > 
> > The reporting framework is designed to look at the output and it is generalized for all the tests.
> 
> If it was general then you wouldn't have to patch output strings in the test cases. Why can't we additionally look at the exit code as well to avoid patching strings?

@letmaik 
Let me brief you the changes i have made:
The current framework will not look exit() code for the status.
We are having PASSED/PASS, TEST_FAILED keywords to state the status.
Changes :
The testcase is implemented using parent and child processes. All the tests were happening in child block.
Usually exit() method will report the status to the parent process. But we have only one process, so the status reporting from exit will not work here. So removed the exit() code.
Also the test does not have PASSED/TEST_FAILED keywords, so i have added those keywords in respective places. Instead of printf , i have used tst_resm function because it is a function designed in framework for status report using (TFAIL/TPASS).
Also commented parent waiting for the child status.

vamsikrishna935:
@letmaik : I have removed PASSED/TEST_FAILED keywords, could you please review .

letmaik (approved):

PR #218, opened 2020-05-06
Use OE's default exception handler

jhand2:
OE now supports making ocalls from the default exception handler. Use this instead of the custom exception continuation hook.

### DIFF
--- a/src/enclave/enclave_signal.c
+++ b/src/enclave/enclave_signal.c
@@ -40,9 +40,6 @@ static struct oe_hw_exception_map exception_map[] = {
     {OE_EXCEPTION_SIMD_FLOAT_POINT, X86_TRAP_XF, SIGFPE, true},
 };
 
-extern void (*oe_continue_execution_hook)(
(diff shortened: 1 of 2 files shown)


jhand2:
Unclear to me why this failed: https://dev.azure.com/sgx-lkl/ff25f828-9f87-48e4-94f0-7449609f7e8f/_apis/build/builds/221/logs/115

Looks unrelated to these changes

letmaik:
I had a look at the test and yes, it's a mystery to me. It outputs:
```
fopen failed errno(2)
TEST FAILED
test: 1 failed
```
The above means that test 1 failed. But test 1 succeeds if the errno is 2, which it is (see first line). I have no idea why it failed, and I can't reproduce it locally. Cosmic radiation?? @prp

davidchisnall (approved):

jhand2:
CI looks good now. @letmaik @prp @davidchisnall can this be merged?

PR #222, opened 2020-05-07
Print host-side non-errors to stdout instead of stderr

letmaik:


prp:
The rationale here was to have only application output use stdout, so we can easily ignore any output from SGX-LKL.

letmaik:
I don't think this makes sense as application output may also use stderr (though currently this is all merged into stdout). If you don't want SGX-LKL output then you should adjust the log level. Isn't that how it's normally done?

prp:
I don't have a strong opinion on this.

davidchisnall:
The normal convention for UNIX programs is to provide all diagnostic messages on stderr and messages that form part of the normal output on stdout. The rationale is that control and diagnostic messages should not affect the next process in a pipe chain. This change seems to contradict that model. None of the SGX-LKL messages should be consumed in normal use by the program receiving the pipe output.

Note that we lose this distinction for code running inside the enclave. Ideally, we’d provide a second VirtIO console channel and send that to stderr on the host and use it for stderr in the enclave. That’s not a very big change to make, but it’s low priority unless we have an important use case that involves using SGX-LKL with pipes (inference, maybe, with a confidential model but public inputs and outputs?).

letmaik:
OK, makes sense. Thanks for the explanation! I'll close this PR.

PR #228, opened 2020-05-11
Resolve relative Docker entrypoints

### LINKED ISSUE #212: Support relative path for app executable
If a Dockerfile has `ENTRYPOINT ["python3"]` then this will work if the base image (some Ubuntu) includes `/usr/bin` in the `PATH` environment variable (I'm assuming the Docker daemon replicates the usual shell search logic). Similary, if the Dockerfile has `WORKDIR /foo` and `ENTRYPOINT ./myapp` then it will run `/foo/myapp`.

SGX-LKL takes neither the working directory nor `PATH` into account and will only accept absolute executable paths. This means automatic conversion with `sgx-lkl-disk` + `sgx-lkl-cfg` will produce an invalid app config. Even if `sgx-lkl-cfg` is not used and instead the app config is hand-crafted it may still be confusing/surprising.

letmaik:
Fixes #212.

This handles the following cases:
```
ENTRYPOINT ["python3"]
```
```
CMD ["python3"]
```
```
WORKDIR /app
ENTRYPOINT ["./myapp"]
```
```
WORKDIR /app
CMD ["./myapp"]
```

If the entrypoint/cmd is given in shell form like
```
ENTRYPOINT python3
```
then Docker translates this into `/bin/sh -c python3` which results in `/bin/sh` being extracted as entrypoint. This is annoying but OK. The `sgx-lkl-cfg` tool prints a helpful advice message in that case.

### DIFF
--- a/tools/sgx-lkl-cfg
+++ b/tools/sgx-lkl-cfg
@@ -64,6 +64,7 @@ def create(args):
         img_folder = img_path.parent
         img_filename = img_path.name
         docker_meta_path = img_folder / (img_filename + '.docker')
+        docker_entrypoint_path = img_folder / (img_filename + '.docker_entrypoint')
         img_roothash_path = img_folder / (img_filename + '.roothash')
         img_roothash_offset_path = img_folder / (img_filename + '.hashoffset')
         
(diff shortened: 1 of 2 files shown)


davidchisnall (approved):

prp (approved):
LGTM

Issue #151, opened 2020-05-04
Remove libc dependencies from src/enclave


davidchisnall:
As part of the relayering work, we should ensure that nothing in `src/enclave` depends on libc init.  Some of these changes are simple and mechanical:

 * Change `malloc` to `oe_malloc` so that we don't need musl's version of malloc to be working before we start.  
 * Remove use of C asserts and use `sgxlkl_fail`.

Others are more complex.

We can't remove a small set of things that we expect to exist in a freestanding C environment (e.g. `memset` / `memcpy`), because the compiler will emit these even if we don't.

Checklist:
- [x] `enclave_event_channel.c`
  - [x] `calloc` / `free`
  - [x] `__assert_fail`
- [ ] `enclave_init.c`
  - [ ] `__init_libc` / `__init_tls` / `__libc` / `__libc_start_init`.  This can't be fixed until we can start LKL without any other libc dependencies.
  - [x] `malloc`.  
  - [x] `strdup`
  - [ ] `wg_get_device` / `wg_key_to_base64` / `wgu_add_peers` / `wgu_list_devices`
- [x] `enclave_mem.c`
  - [x] `___errno_location`.  We should not be using `errno` anywhere at this layer.
  - [x] `mprotect` should be an explicit call to the host mprotect.
  - [x] `lseek` / `read` are both used in the syscall replacements.  These functions should be moved into `src/lkl` and should call the relevant LKL functions directly, not go via libc.
- [ ] `enclave_oe.c`
  - [ ] `__init_tls` / `_dlstart_c` should be removed after we no longer depend on libc for this layer.
- [x] `enclave_util.c`
  - [x] `snprintf` This is used only for tracing.  The tracing is LKL-specific and so should be in `src/lkl`.  This particular usage can be replaced by `strcpy` / `memcpy` because it's just appending to a buffer.
- [x] `sgxlkl_app_config.c`
  - [x] `___errno_location`.  We should not be using `errno` anywhere at this layer.
  - [x] `fprintf` / `__stderr_FILE` should be `sgxlkl_verbose` / `sgxlkl_fail`.
  - [x] `malloc` / `free`
  - [x] `strdup`
  - [x] `snprintf` 
  - [x] `strerror` (We shouldn't be calling anything that sets `errno`, so this should go away once the other fixes happen)
- [x] `mpmc_queue.c`
  - [x] `assert` (Replaced with `SGXLKL_ASSERT`)

The full list of dependencies can be seen by running the following command:

```
[...]

PR #229, opened 2020-05-11
Replace libc memory calls in enclave_event_channel

SeanTAllen:
Part of the relayering work for #151

### DIFF
--- a/src/enclave/enclave_event_channel.c
+++ b/src/enclave/enclave_event_channel.c
@@ -12,6 +12,8 @@
 #include "enclave/ticketlock.h"
 #include "enclave/vio_enclave_event_channel.h"
 
+#include "openenclave/corelibc/oemalloc.h"
+
 static struct ticketlock** evt_chn_lock;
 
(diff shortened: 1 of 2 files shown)


davidchisnall:
@SeanTAllen, enclave_event_channel.o still contains undefined references to calloc.  The following lines still use it:

```c
    evt_chn_lock = (struct ticketlock**)calloc(
        evt_channel_num, sizeof(struct ticketlock*));
```

And:
```c
        evt_chn_lock[i] =
            (struct ticketlock*)calloc(1, sizeof(struct ticketlock));
```

Please can you validate these changes with the nm command from #151?  We should not see any underfined symbols for malloc / calloc / strdup, and so on.

SeanTAllen:
@davidchisnall yes, those were intentionally skipped for now as switching them causes a couple test applications to segfault in HW mode. I'm tracking that down. This PR was only for the ones that are allocated and freed in the file and passed CI when changed.

davidchisnall:
Thanks.

PR #232, opened 2020-05-11
Cleanup Dockerfile in TF sample

letmaik:
I also removed the old 1.11 Dockerfile which shouldn't be used anymore and is just confusing to have.

I would appreciate if someone could test this locally as I merely ported changes from another project and haven't tested this PR as it is yet due to the long build times of TF.


letmaik:
OK found two bugs that I fixed now. All working.

prp (approved):
LGTM

PR #233, opened 2020-05-11
Tiny contained change to oe_ memory functions

SeanTAllen:


letmaik:
@SeanTAllen This looks like it's not ready for review yet. Could you make it a draft PR?

SeanTAllen:
@davidchisnall can you have a look at this set of changes for #151? There are additional malloc/calloc/strdup/free calls to replace however, doing so causes some of the test suite to segfault. 

I spent some time looking at it and haven't found any obvious cause yet.

davidchisnall (approved):

PR #235, opened 2020-05-12
Adding java scenario test for reading file

hukoyu:
Adding java scenario test for reading file

MainApp class uses FileReadSample class to read file input.txt and verifies content of the file to pass the java scenario.

@letmaik please review


tests/languages/java-file/app/MainApp.java
```diff
@@ -0,0 +1,19 @@
...
+		System.out.println("Hello SGX world from MainApp Java!");
+		FileReadSample frs = new FileReadSample();
+		System.out.println("Trying to read file /app/input.txt");
+		String data = frs.readFile("/app/input.txt");
+
+		System.out.println("Content of the file:");
+		System.out.println(data);
+		
+		if (data.compareTo("This is the content of file\n") == 0) {
+		    System.out.println("TEST_PASSED");
```
letmaik: Let's exit with exit code 1 in case of failure, or just throw an exception which has the same effect. Identifiers should only be a last resort in tests.

tests/languages/java-file/app/FileReadSample.java
```diff
@@ -0,0 +1,19 @@
+import java.io.*;
+import java.nio.file.*;
+import java.util.stream.*;
+import java.nio.charset.*;
+
+public class FileReadSample {
+	public  String readFile(String filePath) {
+	StringBuilder contentBuilder = new StringBuilder();
+		try (Stream<String> stream = Files.lines( Paths.get(filePath), StandardCharsets.UTF_8))
```
letmaik: Inconsistent indentation.

letmaik: I would remove the try-catch and let it crash if something goes wrong.

tests/languages/java-file/Makefile
```diff
@@ -0,0 +1,36 @@
...
+endif
+
+.DELETE_ON_ERROR:
+.PHONY: run run-hw run-sw clean
+
+$(APP_ROOT)/%.class: $(APP_ROOT)/%.java
+	javac -source 1.8 -target 1.8 $(APP_ROOT)/*.java
+
+$(DISK_IMAGE): $(APP_ROOT)/$(PROG).class
+	${SGXLKL_DISK_TOOL} create --size=${IMAGE_SIZE} --alpine="iputils iproute2 unzip libstdc++ openjdk8-jre nss" --copy=./${APP_ROOT}/ ${DISK_IMAGE}
```
letmaik: Could you use a Dockerfile instead of `--alpine`? With a Dockerfile you can then compile inside Docker. I'd like to move over to Docker images for all new tests and gradually for existing ones as well.

tests/languages/java-file/app/input.txt
```diff
@@ -0,0 +1 @@
+This is the content of file
```
letmaik: Either `This is the content of the file` or `This is the content of input.txt`.

letmaik (changes requested):

tests/languages/java-file/app/FileReadSample.java
hukoyu: Done

hukoyu: try-catch removed per below feedback. Indentation is fixed.

tests/languages/java-file/app/MainApp.java
hukoyu: Done. Added exit()

tests/languages/java-file/app/input.txt
hukoyu: Fixed the grammar of data and made it two lines.

tests/languages/java-file/Makefile
hukoyu: Implemented using Dockerfile and verified it is working properly.

hukoyu:
@letmaik applied all of your feedback. Ready to be reveiewed.

tests/languages/java-file/Makefile
```diff
@@ -0,0 +1,35 @@
...
+   SGXLKL_VERBOSE=1 SGXLKL_KERNEL_VERBOSE=1 SGXLKL_TRACE_LKL_SYSCALL=0 SGXLKL_TRACE_SIGNAL=0
+else
+SGXLKL_ENV=
+endif
+
+.DELETE_ON_ERROR:
+.PHONY: run run-hw run-sw clean
+
+$(DISK_IMAGE): $(MAINAPP_JAVA) $(FILEREADER_JAVA)
+	docker build -t java-app:dev .
```
letmaik: This works, but in tests we should use the Dockerfile directly in sgx-lkl-disk to avoid creating tagged/named Docker images on the CI or dev machines. You can use `--docker=Dockerfile` and remove the docker command.

tests/languages/java-file/app/MainApp.java
```diff
@@ -0,0 +1,22 @@
...
+	public static void main(String[] args) throws IOException {
+		System.out.println("Hello SGX world from MainApp Java!");
+		FileReadSample frs = new FileReadSample();
+		System.out.println("Trying to read file /app/input.txt");
+		String data = frs.readFile("/app/input.txt");
+
+		System.out.println("Content of the file:");
+		System.out.println(data);
+		
+		if (data.compareTo("This is the first line of the file\nThis is the second line of the file\n") == 0) {
```
letmaik: I was thinking of this:
```
	        if (data.compareTo("This is the first line of the file\nThis is the second line of the file\n") != 0) {
		    System.exit(1);
		}
```

Would that not be enough?

letmaik (changes requested):
Looks great, nearly done!

tests/languages/java-file/app/MainApp.java
hukoyu: Sure will do the necessary change.

hukoyu: Done

tests/languages/java-file/Makefile
hukoyu: Done

hukoyu:
@letmaik I applied latest feedback. Ready to be merged.

tests/languages/java-file/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM openjdk:8-alpine
```
Pengpeng-Microsoft: Please change file structure as java / java_file.

Pengpeng-Microsoft (changes requested):

tests/languages/java-file/Dockerfile
hukoyu: Done. Created java_hello_world and java_file under java folder. Will add new java-* folders under here

hukoyu:
@Pengpeng-Microsoft I applied your feedback. Could you please review and approve if looks good.

Pengpeng-Microsoft (approved):

letmaik:
I think it would be better to remove the `java_` prefix in the subfolders since the test names include the whole folder hierarchy. While you're at it I would rename "file" to "read-file" to make it absolutely clear what the test is doing.

davidchisnall:
What does this add to our test coverage?  We already have a test that reading a file works (the hello world test does this) and a test that a JVM can start.  The JVM is reading a file via the same libc interfaces as the other test, so it's not obvious to me that this adds anything to our tests.  What is the motivation for adding this test?

hukoyu:
> What does this add to our test coverage? We already have a test that reading a file works (the hello world test does this) and a test that a JVM can start. The JVM is reading a file via the same libc interfaces as the other test, so it's not obvious to me that this adds anything to our tests. What is the motivation for adding this test?

@davidchisnall I was adding more test scenarios with java language to have more complex scenarios than existing hello world java app. This one reads a file, I was planning to add for multithreading, networking and so on to have different more complext samples. You mean we don't need to add more java scenario tests? Hello world is good enough?

hukoyu:
> I think it would be better to remove the `java_` prefix in the subfolders since the test names include the whole folder hierarchy. While you're at it I would rename "file" to "read-file" to make it absolutely clear what the test is doing.

Thanks @letmaik . Before test result title was just including the most inner folder name. Thanks fixing to have whole folder structure. I didn't notice that. 
- moved tests/language/java/java_file to tests/language/java/read_file
- moved tests/language/java/java_hello_world to tests/language/java/hello_world

letmaik (approved):
Looks good. @davidchisnall I agree with you that the new test probably won't exercise more of the guts of SGX-LKL compared to the existing tests, but I'm still in favour of more real-life application tests just to create more confidence. We can consolidate tests later on again if we think we have too many that are not testing distinctive things.

davidchisnall:
I don't think it's a good idea to add tests without understanding what they are testing and why they are in the test suite.

Is this an integration test or a unit test?  We have an integration test for starting a JVM and unit tests for I/O system calls / libc library calls.  If it's intended as an integration test, then the same purpose would be better served by a more complex off-the-shelf Java application (or, ideally, something like the OpenJDK test suite).  

Increasing the things that we do in Java without understanding their impact on SGX-LKL coverage will bloat our test suite, slowing our test cycles, but not deliver any benefit in exchange for this cost.  Spawning threads, for example, is not something I would expect to have any benefit: a hello world Java program that uses Java's printf is already testing threading, the class loader, and GC.  OpenJDK implements an N:M threading model, so we won't necessarily see any extra SGX-LKL-visible concurrency from more threads.  

We want more (nightly) integration tests that check that end-to-end workloads are working.

We want more (pre-merge) unit tests and regression that check that specific features are working.

This does not seem to be either of them.

letmaik:
@davidchisnall I opened https://github.com/lsds/sgx-lkl/issues/247 for that.

hukoyu:
@davidchisnall @letmaik @prp I definitely want to work on adding scenario tests that will add value, I don't want to waste my time on useless scenarios per your explanation. Then my suggestion is please create issues for test scenarios for different languages that will add value to be implemented so that we can implement them and add to scenario tests.

jxyang:
@davidchisnall Since different language runtimes exercise musl/lkl in different ways, and the way to increase language runtime coverages is to add more programs written the particular language, I do see the value of this particular test case. 

Do we need to understand what a test case is testing before we write it? Absolutely. In this case, we are testing Java language features, with little regard to how the language runtime is implemented in sgx-lkl. Ultimately, that's what the users of sgx-lkl care most (if they plan to write confidential applications in Java).

PR #240, opened 2020-05-13
Fixed utime/utime06 testcase

vamsikrishna935:
Patch Description: Test cases were failing as there was not sufficient memory
 to use for test image creation for a loop device for test purpose.
 Test cases are modified to use /dev/vda the root file system device.


hukoyu (approved):
LGTM. cc @letmaik

letmaik (approved):

PR #241, opened 2020-05-13
CI: Slow nightly builds/tests

letmaik:
This is an alternative to https://github.com/lsds/sgx-lkl/pull/140. It introduces a new nightly pipeline by refactoring the existing one into a template. It also adds an `ethreads` template parameter that defines a list of ethread values to test, which is set to 8 for the default pipeline and to 1, 4, 8 in the nightly pipeline. The nightly pipeline runs every day midnight UTC from `oe_port`. We can change the cron schedule if we want.

You can see a manually triggered nightly run here: https://dev.azure.com/sgx-lkl/sgx-lkl/_build/results?buildId=302&view=results.

Note that the nightly pipeline only appears in the GitHub checks below because I manually triggered it.

For reviewing, please have a look at how the test names have changed and whether that's ok.

cc @hukoyu


letmaik:
Interesting, a bunch of stuff is failing for 1 ethread.

.azure-pipelines/scripts/run_test.sh
```diff
@@ -21,7 +21,7 @@ fi
 tests_dir=$SGXLKL_ROOT/tests
 test_name="$(realpath --relative-to="$tests_dir" "$(pwd)")"
 test_name="${test_name//\//-}"
-test_name+="-($build_mode)-($run_mode)"
+test_name+="-($build_mode)-($run_mode)-($SGXLKL_ETHREADS)"
```
hukoyu: This will just a number to the test name. Adding also "threads" suffix will be more clear.

tests/ltp/run_ltp_test.sh
```diff
@@ -65,7 +65,7 @@ for file in ${ltp_tests[@]}; do
     # Initialize the variables.
     test_name=${final_test_name#"-ltp-testcases-"}
     ltp_testcase_name=$(echo ${test_name%.*.*} | sed 's/-/\//g; s/_/\//g')
-    test_name="${test_name}-($build_mode)-($test_mode)"
+    test_name="${test_name}-($build_mode)-($test_mode)-($SGXLKL_ETHREADS)"
```
hukoyu: This will just a number to the test name. Adding also "threads" suffix will be more clear.

.azure-pipelines/azure-pipelines-nightly.yml
```diff
@@ -0,0 +1,19 @@
...
+  branches:
+    include:
+    - oe_port
+  always: true
+
+extends:
+  template: template.yml
+  parameters:
+    skip_publish: true
+    test_timeout_minutes: 120 # ethread==1 takes longer to run
```
hukoyu, line 15: If LTP tests are timing out due to ethread change you might need to increase LTP test timeout as well https://github.com/lsds/sgx-lkl/blob/oe_port/tests/ltp/run_ltp_test.sh#L46

hukoyu (approved):
A few minor comments. Other than that LGTM.

.azure-pipelines/azure-pipelines-nightly.yml
letmaik: Good point, I think the additional timeouts in CI that are still happening are actual deadlocks/hangs, so for now I think 60s per ltp test should work.

tests/ltp/run_ltp_test.sh
letmaik: Yeah, I agree. I was torn between keeping short test names and still including enough information. I wonder if we can abbreviate "ethreads" somehow. Maybe... "kernel-syscalls-fcntl-fcntl34-(nondebug)-(run-hw)-(et=8)"?

letmaik: I went for the full name, it's clearer.

letmaik:
Note to self: grep for SGXLKL_ETHREADS and remove from tests/

letmaik:
@davidchisnall @prp This is ready to be reviewed now.

prp (approved):
LGTM

PR #243, opened 2020-05-13
Java scenario test for threading and synchronization

hukoyu:
This is a java scenario test for threading and syncroization.
- Creates a shared object which has a counter value and this shared object is accessed by all threads 
- Creates 10 threads and passes shared object
- Each thread just increases counter by 1 on shared object synchronized
- Main thread waits all threads to finish
- Then checks the counter value on the shared object to be 10 otherwise returns exit code 1

cc @letmaik

### DIFF
--- a/tests/languages/java/thread/Dockerfile
+++ b/tests/languages/java/thread/Dockerfile
@@ -1,0 +1,4 @@
+FROM openjdk:8-alpine
+COPY ./app /app
+WORKDIR /app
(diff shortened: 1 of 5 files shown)


hukoyu:
@letmaik I checked the pipeline failure and I see that all non-LTP tests and java tests are passing. But there are other test failures and they failed due to Docker daemaon issues.

letmaik (approved):

PR #246, opened 2020-05-14
CI: adapt to Azure Pipelines job user changes

letmaik:
With scale set agents, jobs would run as root in the past. Azure Pipelines rolled out an update to change that and run jobs as non-root user with password-less sudo, like the regular hosted agents do it. This PR adapt to that change.


letmaik:
The user issue is fixed now. CI failures are unrelated.

davidchisnall (approved):

PR #248, opened 2020-05-14
Rewrite Debian packaging script to be based on lddtree only

letmaik:
@davidchisnall You will like this.


davidchisnall (approved):
Yay, much less code!

SeanTAllen:
Nice.

PR #257, opened 2020-05-15
Add test for disabled FSGSBASE

letmaik:
Tests in CI are run with FSGSBASE enabled, both for hw and sw. This test specifically disables FSGSBASE to test the fall-back emulation.


letmaik:
@prp @jhand2 This test is failing. Apparently the fall-back isn't working in sw mode:

```
[   SGX-LKL  ] SGX-LKL (OE) Git version 90d2004-dirty LKL version 5.3.0 [DEBUG build (-O0)] [SOFTWARE DEBUG]
[   SGX-LKL  ] nproc=8 ETHREADS=8 CMDLINE="mem=32M"
[   SGX-LKL  ] HW TLS support: conf->fsgsbase=0
[   SGX-LKL  ] Registering disk 0 (path='sgx-lkl-rootfs.img', mnt='/', [RW   ])
[   SGX-LKL  ] get_signed_libsgxlkl_path... result=/opt/sgx-lkl-debug/bin/../lib/libsgxlkl.so.signed
[   SGX-LKL  ] oe_create_enclave... result=0 (OE_OK)
[[  SGX-LKL ]] wg0 has public key LeYJAb9pj5nWXwUuv739AQXI8aI8NBe/SQAnbFx7kU8=
[[  SGX-LKL ]] FAIL: Encountered an illegal instruction inside enclave (opcode=0x48f3)
```

I can reproduce the same locally.

jhand2:
@letmaik yes, SGX-LKL does not use OE's exception handler in software mode. @anakrish informed me that OE's exception handler doesn't work in simulation mode, although I don't have a good handle on why.

anakrish:
Yes, OE's exception handler does not yet work in simulation mode.
However, based on past discussion, we did not need to add fsgsbase emulation for simulation mode.

@prp From what I remember, LKL does not use wrfsbase in simulation mode. Is that accurate?

prp:
@anakrish yes, I don't understand why we would see `WRFSBASE` used in software mode.

@letmaik could you run this with gdb, identify the illegal instruction and provide a stack trace? Thanks.

letmaik:
```c
[[  SGX-LKL ]] libc_start_main_stage2(): Calling app main: /bin/echo

Thread 8 "ENCLAVE" received signal SIGILL, Illegal instruction.
[Switching to Thread 0x7fff416b9000 (LWP 8814)]
0x00007fff005a8268 in oe_ocall (func=32768, arg_in=140737188055296, arg_out=0x0) at ../../enclave/core/sgx/calls.c:709
(gdb) display/i $pc
1: x/i $pc
=> 0x7fff005a8268 <oe_ocall+288>:       wrfsbase %rax
(gdb) bt
#0  0x00007fff005a8268 in oe_ocall (func=32768, arg_in=140737188055296, arg_out=0x0)
    at ../../enclave/core/sgx/calls.c:709
#1  0x00007fff005a8552 in oe_call_host_function_by_table_id (table_id=0, function_id=1, input_buffer=0x7fffe4000b90, 
    input_buffer_size=112, output_buffer=0x7fffe4000c00, output_buffer_size=32, output_bytes_written=0x7fff40a9fa00, 
    switchless=false) at ../../enclave/core/sgx/calls.c:795
#2  0x00007fff005a43c3 in _call_host_function (function_id=1, input_buffer=0x7fffe4000b90, input_buffer_size=112, 
    output_buffer=0x7fffe4000c00, output_buffer_size=32, output_bytes_written=0x7fff40a9fa00)
    at ../../enclave/core/core_t_wrapper.c:28
#3  0x00007fff005a538a in oe_write_ocall (device=0, 
    str=0x7fff40a9faf0 "[[  SGX-LKL ]] libc_start_main_stage2(): Calling app main: /bin/echo\n", 
    maxlen=18446744073709551615) at enclave/core/core_t.c:435
#4  0x00007fff005a1cae in oe_host_write (device=0, 
    str=0x7fff40a9faf0 "[[  SGX-LKL ]] libc_start_main_stage2(): Calling app main: /bin/echo\n", 
    len=18446744073709551615) at ../../enclave/core/hostcalls.c:175
#5  0x00007fff005a1ab8 in oe_host_vfprintf (device=0, fmt=0x7fff00671a88 "[[  SGX-LKL ]] %s(): Calling app main: %s\n", 
    ap_=0x7fff40a9fc30) at ../../enclave/core/hostcalls.c:124
#6  0x00007fff005a1b87 in oe_host_printf (fmt=0x7fff00671a88 "[[  SGX-LKL ]] %s(): Calling app main: %s\n")
    at ../../enclave/core/hostcalls.c:135
#7  0x00007fff0040f486 in libc_start_main_stage2 (main=0x7fff3dbdddfa, argc=3, argv=0x7fff40a9fda8)
    at src/env/__libc_start_main.c:161
#8  0x00007fff0040f421 in __libc_start_main (main=0x7fff3dbdddfa, argc=3, argv=0x7fff40a9fda8)
    at src/env/__libc_start_main.c:153
#9  0x00007fff3dbdd733 in ?? ()
#10 0x0000000000000003 in ?? ()
(gdb) c
Continuing.
[[  SGX-LKL ]] FAIL: Encountered an illegal instruction inside enclave (opcode=0x48f3)
2020-05-16T08:33:19.000000Z [(H)ERROR] tid(0x7fffee19d700) | :OE_ENCLAVE_ABORTING [../../host/calls.c:oe_call_enclave_function_by_table_id:91]
[   SGX-LKL  ] ethread (2: 19) [   SGX-LKL  ] FAIL: sgxlkl_ethread_init() failed (id=2 result=19 (OE_ENCLAVE_ABORTING))
```

prp:
Ok, @jhand2 @anakrish, so this appears to be an OE issue. It shouldn't be using `wrfsbase` in software mode.

letmaik:
I will disable this test for CI so that it can be merged.

tests/basic/fsgsbase_unset/Makefile
```diff
@@ -0,0 +1,59 @@
+include ../../common.mk
+
+PROG=/bin/echo Hello World
```
prp: Hmm, wouldn't it be better to run something that actually uses TLS?

prp (approved):
LGTM, although perhaps we could rather test with a simple Python etc. program that uses TLS?

tests/basic/fsgsbase_unset/Makefile
letmaik: I guess so, I don't know what a minimal test for TLS that reliably fails if FSGSBASE emulation is not working would look like though.

anakrish:
@jhand2 You could store a pointer to `oe_set_fs_register_base` in `oe_ecall_context_t` and have the enclave call that function to set fs in simulation mode.

jhand2:
I did as @anakrish suggested here https://github.com/openenclave/openenclave/commit/2d92ced21169b0fa9628d60216ac8922f913bc37

@letmaik feel free to rerun CI or do any local testing you need to make sure this test works.

letmaik:
@jhand2 It's working. Thanks!
@prp I'm not sure how to write a proper test for TLS, should we merge this one as it is for now?

prp:
@letmaik your call, but you could do this: https://stackoverflow.com/questions/1408171/thread-local-storage-in-python

letmaik:
That would be an option but I'm not too sure it's a great test since under the hood it may not actually use real thread local storage: https://github.com/python/cpython/blob/master/Lib/_threading_local.py It depends on the distribution. Probably something written in C would be more straight-forward.

letmaik:
@prp I added a C test using TLS. Could you have a look if this makes sense or whether I've forgotten to test something obvious?

prp (approved):
LGTM

Issue #154, opened 2020-05-04
Intercept `clone` to handle `CLONE_CHILD_CLEARTID`


davidchisnall:
The `CLONE_CHILD_CLEARTID` flag to `clone` turns the return value into a futex that is signalled on thread exit.  We should intercept this and register the futex location with the lthread library so that it can be zeroed and sent a wakeup when the underlying lthread exits.  We need to also intercept `set_tid_address` in the same way.

davidchisnall:
Once #259 is merged, this should be possible.  We need to:

- [ ] Add a field on the lthread structure to store the exit futex location.
- [ ] On `thread_destroy_host`, check that field and, if it is non-null, zero the value at that address and wake the futex.
- [ ] Add an interceptor for the `clone` system call that sets this field in the newly created futex.
- [ ] Add an interceptor for `set_tid_address`, which does the same thing.

Carrying the value between the clone interceptor and `thread_create_host` is non-trivial.  The simplest thing to do is to store the value in the stack.  The distance between the `sp` value and `sp-128` is the red zone, but anything after that is fair game according to the ABI.  We can store the futex address (or 0) at `sp-136` and then check this in `thread_create_host`.

davidchisnall:
This is no longer required.  If we move to using the LKL futex implementation for all of userspace, this works already.

PR #259, opened 2020-05-18
Support the clone system call.

### LINKED ISSUE #155: LKL does not support clone
To fix the layering, we need to return to musl creating threads via the `clone` system call.  Currently, [LKL does not implement `clone` at all](https://github.com/lsds/lkl/blob/21cd7e192cf2a97ea69230b9c0c34da15d157579/arch/lkl/include/asm/syscalls.h#L21).

We need to provide an implementation that handles a the flags required for `pthread_create`.  The correct change in LKL may simply be to provide a host_ops hook that handles clone entirely in the LKL consumer.  The musl implementation of `pthread_create` depends on the following `clone` flags:

 - `CLONE_VM`:  Share address space with the parent. In a single address space world, we cannot support anything other than this.
 - `CLONE_FS`: Share a filesystem namespace with the parent. In a single-process world, this is the obvious thing to do.
 - `CLONE_FILES`: Share a file descriptor table with the parent.  We probably want to support not having this so that our init process can have a separate FD table.
 - `CLONE_SIGHAND`. Share signal handlers. We probably want to support not having this so that our init process can have separate signal handlers.
 - `CLONE_THREAD`. New thread shares PID and has a separate TGID from the parent.  It would be nice to support both variations of this so that we can have a distinct PID for init.
 - `CLONE_SYSVSEM`. Shares ownership of SysV semaphores with the parent.  It doesn't matter too much if we support this because our init process shouldn't use SysV IPC.
 - `CLONE_SETTLS`. Sets the TLS pointer.  Should simply set the %fs base value.
 - `CLONE_PARENT_SETTID`.  Sets the child thread ID in the parent.  Should be easy to support.
 - `CLONE_CHILD_CLEARTID`. Uses the return thread ID as a futex.  See #154.
 - `CLONE_DETACHED`. Has no effect in modern Linux, safe to ignore.

davidchisnall:
This implements two new LKL hooks. The first one to create an lthread
with a specific initial register state (to capture the returns-twice
behaviour of clone, along with the caller's ability to define the stack
and TLS addresses).  The new thread is immediately associated with the
Linux task structure (normally, lthreads are associated with Linux tasks
lazily when they perform a system call).

The second hook destroys a thread.  This is done in response to an exit
system call.  This is somewhat complicated, because LKL never returns to
this thread and the thread's stack may be deallocated by the time we
exit it.

The lthread scheduler does not have an easy way of adding a mechanism to
kill a thread without that thread running.  We can add one eventually,
but for now create a temporary stack that lthreads can use during
teardown and make them run the teardown from there.

Fixes #155

### DIFF
--- a/src/include/enclave/lthread.h
+++ b/src/include/enclave/lthread.h
@@ -196,6 +196,23 @@ extern "C"
         size_t sleeptime_ns,
         size_t futex_wake_spins);
 
+    /**
+     * Create a new thread where the caller manages the initial thread state.
+     * The newly created thread is returned via `new_lt`.  The newly created
+     * thread will begin executing from `pc`, with its stack pointer set to
+     * `sp` and its TLS area set to `tls`.  It is the caller's responsibility
+     * to ensure that the stack and TLS area allocated by this thread are
+     * cleaned up.
+     *
+     * The thread is not scheduled by this call and must be explicitly
+     * scheduled by the caller.
+     */
+    int lthread_create_primitive(
+        struct lthread** new_lt,
+        void* pc,
+        void* sp,
+        void* tls);
+
     int lthread_create(
         struct lthread** new_lt,
         struct lthread_attr* attrp,
@@ -238,9 +255,38 @@ extern "C"
 
     int lthread_key_delete(long key);
 
-    void* lthread_getspecific(long key);
+    /**
+     * Access a thread-local variable corresponding to the key given by `key`,
+     * in the thread specified by `lt`.  This function is not safe to call
+     * while `lt` is running or concurrently with a call to
+     * `lthread_setspecific_remote` on the same lthread.  It is the caller's
+     * responsibility to ensure that this does not happen, for example after
+     * the thread has been removed from the scheduler during tear-down or by
+     * explicitly descheduling it.
+     */
+    void* lthread_getspecific_remote(struct lthread* lt, long key);
     
-    int lthread_setspecific(long key, const void* value);
+    /**
+     * Sets a thread-local variable corresponding to the key given by `key` to
+     * `value`, in the thread specified by `lt`.  This function is not safe to
+     * call while `lt` is running or concurrently with a call to
+     * `lthread_setspecific_remote` on the same lthread.  It is the caller's
+     * responsibility to ensure that this does not happen.  The most common use
+     * for this is between a call to `lthread_create_primitive` and
+     * `__scheduler_enqueue`, to initialise a thread-local variable before a
+     * thread starts.
+     */
+    int lthread_setspecific_remote(struct lthread* lt, long key, const void* value);
+
+    static void* lthread_getspecific(long key)
+    {
+        return lthread_getspecific_remote(lthread_current(), key);
+    }
+
+    static int lthread_setspecific(long key, const void* value)
+    {
+        return lthread_setspecific_remote(lthread_current(), key, value);
+    }
 
     static inline void __scheduler_enqueue(struct lthread* lt)
     {
--- a/src/lkl/posix-host.c
+++ b/src/lkl/posix-host.c
@@ -324,6 +324,70 @@ static lkl_thread_t thread_create(void (*fn)(void*), void* arg)
     return (lkl_thread_t)thread;
 }
 
+/**
+ * Create an lthread to back a Linux task, created with a clone-family call
+ * into the kernel.
+ */
+static lkl_thread_t thread_create_host(void* pc, void* sp, void* tls, struct lkl_tls_key* task_key, void* task_value)
+{
+    struct lthread* thread;
+    // Create the thread.  The lthread layer will set up the threading data
+    // structures and prepare the lthread to run with the specified instruction
+    // and stack addresses.
+    int ret = lthread_create_primitive(&thread, pc, sp, tls);
+    if (ret)
+    {
+        sgxlkl_fail("lthread_create failed\n");
+    }
+    // Store the host task pointer.  LKL normally sets this lazily the first
+    // time that a thread calls into the LKL.  Threads created via this
+    // mechanism begin life in the kernel and so need to be associated with the
+    // kernel task that created them.
+    lthread_setspecific_remote(thread, task_key->key, task_value);
+    // Mark the thread as runnable.  This must be done *after* the
+    // `lthread_setspecific_remote` call, to ensure that the thread does not
+    // run while we are modifying its TLS.
+    __scheduler_enqueue(thread);
+    return (lkl_thread_t)thread;
+}
+
+static void host_thread_exit(void)
+{
+    LKL_TRACE("enter");
+    lthread_exit(0);
+}
+
+/**
+ * Destroy the lthread backing a host task created with a clone-family call.
+ * This is called after an `exit` system call.  The system call does not return
+ * and the lthread backing the LKL thread that issued the task will not be
+ * invoked again.
+ */
+static void thread_destroy_host(lkl_thread_t tid, struct lkl_tls_key* task_key)
+{
+    // Somewhat arbitrary size of the stack for teardown.  This must be big
+    // enough that `host_thread_exit` can run and call any remaining TLS
+    // destructors.  This can be removed once there is a clean mechanism for
+    // destroying a not-running lthread without scheduling it.
+    static const size_t teardown_stack_size = 8192;
+    struct lthread *thr = (struct lthread*)tid;
+    // The thread is currently blocking on the LKL scheduler semaphore, remove
+    // it from the sleeping list.
+    _lthread_desched_sleep(thr);
+    // Delete its task reference in TLS.  Without this, the thread's destructor
+    // will call back into LKL and deadlock.
+    lthread_setspecific_remote(thr, task_key->key, NULL);
+    // Give the thread a stack to use during lthread teardown.
+    thr->attr.stack_size = teardown_stack_size;
+    thr->attr.stack = enclave_mmap(0, teardown_stack_size, 0, PROT_READ | PROT_WRITE, 1);
+    // Set up the state so that this will call into the host_thread_exit function and 
+    thr->ctx.eip = host_thread_exit;
+    thr->ctx.esp = thr->attr.stack + teardown_stack_size;
+    // Schedule the thread again.  It will exit when it is next scheduled.
+    __scheduler_enqueue(thr);
+}
+
+
 static void thread_detach(void)
 {
     LKL_TRACE("enter\n");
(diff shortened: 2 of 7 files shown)


davidchisnall:
Don't merge this yet: the LKL commit will change once lsds/lkl#1 is merged.  I'll force-push to this branch to update the LKL submodule once that's done.

davidchisnall:
Note that this doesn't yet fix #154.

src/include/enclave/lthread.h
```diff
@@ -196,6 +196,12 @@ extern "C"
         size_t sleeptime_ns,
         size_t futex_wake_spins);
 
+    int lthread_create_primitive(
```
prp, line 210: Comment?

src/include/enclave/lthread.h
```diff
@@ -238,9 +244,19 @@ extern "C"
 
     int lthread_key_delete(long key);
 
-    void* lthread_getspecific(long key);
+    void* lthread_getspecific_remote(struct lthread* lt, long key);
```
prp, line 267: Comment?

src/include/enclave/lthread.h
```diff
@@ -238,9 +244,19 @@ extern "C"
...
     
-    int lthread_setspecific(long key, const void* value);
+    int lthread_setspecific_remote(struct lthread* lt, long key, const void* value);
```
prp, line 279: Comment?

src/lkl/posix-host.c
```diff
@@ -324,6 +324,69 @@ static lkl_thread_t thread_create(void (*fn)(void*), void* arg)
     return (lkl_thread_t)thread;
 }
 
+/**
+ * Create an lthread to back a Linux task, created with a clone-family call
+ * into the kernel.
```
prp, line 329: It would be nice to document the parameters, either here or in the header file.

src/lkl/posix-host.c
```diff
@@ -324,6 +324,69 @@ static lkl_thread_t thread_create(void (*fn)(void*), void* arg)
...
+
+static void host_thread_exit(void)
+{
+    LKL_TRACE("enter");
+    lthread_exit(0);
+}
+
+ 
+
+
```
prp:
```suggestion
```

src/lkl/posix-host.c
```diff
@@ -324,6 +324,69 @@ static lkl_thread_t thread_create(void (*fn)(void*), void* arg)
...
+/**
+ * Destroy the lthread backing a host task created with a clone-family call.
+ * This is called after an `exit` system call.  The system call does not return
+ * and the lthread backing the LKL thread that issued the task will not be
+ * invoked again.
+ */
+static void thread_destroy_host(lkl_thread_t tid, struct lkl_tls_key* task_key)
+{
+    static const size_t teardown_stack_size = 8192;
```
prp, line 372: A comment would be useful here. Will we ever have to increase this? Perhaps define this as a constant?

src/lkl/posix-host.c
```diff
@@ -324,6 +324,69 @@ static lkl_thread_t thread_create(void (*fn)(void*), void* arg)
...
+    thr->attr.stack_size = teardown_stack_size;
+    thr->attr.stack = enclave_mmap(0, teardown_stack_size, 0, PROT_READ | PROT_WRITE, 1);
+    // Set up the state so that this will call into the host_thread_exit function and 
+    thr->ctx.eip = host_thread_exit;
+    thr->ctx.esp = thr->attr.stack + teardown_stack_size;
+    // Schedule the thread again.  It will exit when it is next scheduled.
+    __scheduler_enqueue(thr);
+}
+
+
```
prp, line 390:
```suggestion

```

src/sched/lthread.c
```diff
@@ -627,6 +628,98 @@ static void init_file_lock(FILE* f)
         f->lock = 0;
 }
 
+int lthread_create_primitive(
```
prp, line 631: As above, it would be good to document these parameters somewhere.

tests/basic/clone/clone.s
```diff
@@ -0,0 +1,25 @@
+.text
```
prp: Out of curiosity, why do you need assembly to invoke the clone syscall instead of using lkl_syscall directly?

prp (approved):
LGTM, some minor comments. I see CI failures though...

src/lkl/posix-host.c
davidchisnall: This is a constant, but I agree on the comment.

tests/basic/clone/clone.s
davidchisnall: It needs to set up the stack explicitly.  The system call returns twice to the same instruction but with the new stack pointer.  You need to then pop the real call address from the stack somehow, which is easy in assembly and impossible to do in C without the compiler being allowed to break it.  Musl and glibc use assembly to call the clone system call directly (rather than via the `syscall` wrapper) for the same reason.

davidchisnall:
CI is failing because the weak symbol for `lkl_syscall` isn't found.  It does for me, so I'm not entirely sure what's going on here...

tests/basic/clone/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM alpine:3.6 AS builder
```
SeanTAllen, line 1: out of curiosity, given that 3.6 is rather dated. is there a "no later than X" dependency for sgx-lkl that I'm not aware of?

tests/basic/clone/clone.c
```diff
@@ -0,0 +1,99 @@
...
+	void **stack = (void**)child_stack;
+	stack[-1] = fn;
+	stack[-2] = arg;
+	return do_clone(220, args);
+}
+
+static void sigsegv(int signo, siginfo_t *si, void *addr)
+{
+	fprintf(stderr, "Fault: %p (%p)\n", si->si_addr, addr);
+	_Exit(0);
```
SeanTAllen: why 0 here?

tests/basic/clone/Dockerfile
prp: We should use a consistent Alpine version in all tests. @letmaik?

tests/basic/clone/clone.s
```diff
@@ -0,0 +1,25 @@
...
+.global do_clone
+.hidden do_clone
+.type   do_clone,@function
+do_clone:
+	push %rax
```
davidchisnall: Explain why this is needed.

tests/basic/clone/Dockerfile
letmaik: I opened https://github.com/lsds/sgx-lkl/issues/260. For this PR, it's fine to leave it at an arbitrary version. The [maximum supported](https://github.com/lsds/sgx-lkl/issues/200) currently is 3.10.

tests/basic/clone/clone.c
davidchisnall: Dead code from when I was getting segfaults in the kernel because the stack was misaligned in the clone assembly wrapper.

tests/basic/clone/clone.s
davidchisnall: This code is now in musl, with the relevant comment.

SeanTAllen:
I started looking into the failing `access02` ltp test. Here's how to replicate.

- Start from fresh repo of the `clone` branch.
- make DEBUG=true
- cd tests/ltp/ltp-batch1
- edit ../batch.mk
- comment out the commands for run-hw and run-sw commands (those commands run all tests)

```
run-hw: $(ROOT_FS)
        #@${LTP_TEST_SCRIPT} run-hw

run-sw: $(ROOT_FS)
        #@${LTP_TEST_SCRIPT} run-sw
```

- make clean
- make DEBUG=true
- make run

At this point, you are prepped to run the single test for HW mode:

- SGXLKL_VERBOSE=1 SGXLKL_KERNEL_VERBOSE=0   ../../../build/sgx-lkl-run-oe   --hw-debug   sgxlkl-miniroot-fs.img   /ltp/testcases/kernel/syscalls/access/access02

for SW mode:

SGXLKL_VERBOSE=1 SGXLKL_KERNEL_VERBOSE=0   ../../../build/sgx-lkl-run-oe   --sw-debug   sgxlkl-miniroot-fs.img   /ltp/testcases/kernel/syscalls/access/access02

Result I get
--------------

HW:

tst_test.c:1106: INFO: Timeout per run is 0h 05m 00s
tst_test.c:1125: INFO: No fork support
access02.c:144: PASS: access(file_f, F_OK) as root behaviour is correct.
access02.c:144: PASS: access(file_f, F_OK) as nobody behaviour is correct.
access02.c:144: PASS: access(file_r, R_OK) as root behaviour is correct.
access02.c:144: PASS: access(file_r, R_OK) as nobody behaviour is correct.
access02.c:144: PASS: access(file_w, W_OK) as root behaviour is correct.
access02.c:144: PASS: access(file_w, W_OK) as nobody behaviour is correct.
bad count while changing owner
[[  SGX-LKL ]] FAIL: Kernel panic! Run DEBUG build with SGXLKL_KERNEL_VERBOSE=1 for more information. Aborting...
2020-05-19T19:52:36.000000Z [(H)ERROR] tid(0x7fe8a8ff1700) | :OE_ENCLAVE_ABORTING [/home/sean/openenclave-sgxlkl.git/host/calls.c:oe_call_enclave_function_by_table_id:91]
[   SGX-LKL  ] ethread (4: 19) [   SGX-LKL  ] FAIL: sgxlkl_ethread_init() failed (id=4 result=19 (OE_ENCLAVE_ABORTING))

SW:

It hangs after printing access02.c:144: PASS: access(file_w, W_OK) as nobody behaviour is correct.

----

This is somewhat different than what I am seeing in CI, so, before I proceed any further, @davidchisnall can you try recreating the above?

----

The source of the failing test is https://github.com/linux-test-project/ltp/blob/master/testcases/kernel/syscalls/access/access02.c

davidchisnall:
Thanks @SeanTAllen.  I actually see something more interesting when I try to reproduce this locally:

```
tst_test.c:1106: INFO: Timeout per run is 0h 05m 00s
tst_test.c:1125: INFO: No fork support
access02.c:144: PASS: access(file_f, F_OK) as root behaviour is correct.
access02.c:144: PASS: access(file_f, F_OK) as nobody behaviour is correct.
access02.c:144: PASS: access(file_r, R_OK) as root behaviour is correct.
access02.c:144: PASS: access(file_r, R_OK) as nobody behaviour is correct.
access02.c:144: PASS: access(file_w, W_OK) as root behaviour is correct.
access02.c:144: PASS: access(file_w, W_OK) as nobody behaviour is correct.
Created new host task 7f2fc02f4540 (for 7f2fbeada4c0)
```


So it appears that we're hitting the code path for creating a new clone'd task (for a test that shouldn't be calling clone).  I'll take a look.

PR #261, opened 2020-05-18
Test oe_get_report & fix CC test

letmaik:
This PR adds a dummy `oe_get_report_v2` call at enclave start-up as a placeholder for future use. Doing that exercises the Azure DCAP Client library and associated host-side networking and will provide insight into issues appearing in CI over time. The confidential container test was fixed to mount the missing SSL certificates loaded by libcurl (via the Azure DCAP Client library) from the host via `-v /etc/ssl/certs:/etc/ssl/certs:ro`.

### DIFF
--- a/src/enclave/enclave_init.c
+++ b/src/enclave/enclave_init.c
@@ -128,6 +128,24 @@ static int startmain(void* args)
     /* Get the application configuration & disk param from remote server */
     enclave_get_app_config(&app_config);
 
+    /* Retrieve remote attestation report to exercise Azure DCAP Client (for testing only) */
+    // TODO replace this later on
+    uint8_t* remote_report;
+    size_t remote_report_size;
+    oe_result_t result = OE_UNEXPECTED;
+    result = oe_get_report_v2(
+        OE_REPORT_FLAGS_REMOTE_ATTESTATION,
+        NULL,
+        0,
+        NULL,
+        0,
+        &remote_report,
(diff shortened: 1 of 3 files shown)


letmaik:
Note that an alternative to `-v /etc/ssl/certs:/etc/ssl/certs:ro` could be to simply install `ca-certificates` (=`/etc/ssl/certs`) into the CC, which shouldn't be an issue. This is similar to https://github.com/GoogleContainerTools/distroless/blob/master/base/README.md.

prp:
> Note that an alternative to `-v /etc/ssl/certs:/etc/ssl/certs:ro` could be to simply install `ca-certificates` (=`/etc/ssl/certs`) into the CC, which shouldn't be an issue. This is similar to https://github.com/GoogleContainerTools/distroless/blob/master/base/README.md.

Shouldn't the certificates be part of the attestation?

letmaik:
These are just the host-side certificates used by Azure DCAP Client, not the ones inside the enclave.

letmaik:
I've changed it so that /etc/ssl/certs/ca-certificates.crt is bundled in the CC, avoiding any extra mounting.

letmaik:
Can this be merged?

prp:
I've added @jxyang as a reviewer to get more folks involved in this, thanks.

davidchisnall (approved):

jxyang (approved):

Issue #121, opened 2020-04-30
Enclave Timing is Incorrect


praenubilus:
We have found time issue in sgxlkl oe_port branch.
The symptom: when starting enclave, the system time in the enclave will be set to 
**Thu Jan  1 00:00:00 UTC 1970**
@jxyang can verify that.

SeanTAllen:
This is probably because https://github.com/lsds/sgx-lkl-musl-oe/pull/30 was moved over and the sgx-lkl work required to make that a reasonable change is currently hung up on some issue (I'm currently working on it).

Adding that change back in the syscall.h should fix the issue.

SeanTAllen:
@jxyang is currently confirming the fix for this.

jxyang:
Yes, the fix enabled the ocall workaround and the timing is correct now.

SeanTAllen:
@prp can you close this issue? I don't have permissions to close.

hukoyu:
How is this fixed? I don't see the related PR.

jxyang:
This is the PR: https://github.com/lsds/sgx-lkl/pull/126

hukoyu:
This issue still exists. Issue https://github.com/lsds/sgx-lkl/issues/157 created since we cannot reopen this issue.

PR #270, opened 2020-05-19
Enabling ltp/kernel/syscalls/stime/stime01

vamsikrishna935:
Testcase is passing now after https://github.com/lsds/sgx-lkl/issues/121 is
fixed. Hence enabling the testcase


hukoyu:
LGTM @letmaik

letmaik (approved):

PR #273, opened 2020-05-19
Confidential Container docs

letmaik:
This extends the main README to include steps on creating a confidential container.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,8 +58,9 @@ echo "deb [trusted=yes] https://clcpackages.blob.core.windows.net/apt/1fa5fb889b
 Now, install with:
 ```sh
 sudo apt update
-sudo apt install clc
+sudo apt install clc-debug
 ```
+To install the non-release variant, use `clc-nonrelease`. A release variant is not available yet.
 
 The FSGSBASE DKMS driver can be installed with:
 ```sh
@@ -270,14 +271,21 @@ sgx-lkl-run-oe --hw-debug ./sgxlkl-disk.img /usr/bin/python /app/myapp.py
 
 #### Creating Docker-based disk images
 
-The `sgx-lkl-disk` tool can also build disk images from Dockerfiles with the `--docker`
-flag, e.g. when an application needs to be compiled manually. Note that SGX-LKL 
-applications still need to be linked against musl libc, so a good starting 
-point is an Alpine Docker base image.
+The `sgx-lkl-disk` tool can also build disk images from Dockerfiles or Docker images
+with the `--docker` flag, e.g. when an application needs to be compiled manually.
+Note that SGX-LKL applications still need to be linked against musl libc,
+so a good starting point is an Alpine Docker base image.
 
 To build an SGX-LKL disk image from a Dockerfile, run:
+```sh
+# This assumes a MyDockerfile file exists in the current directory.
+sgx-lkl-disk create --size=100M --docker=./MyDockerfile sgxlkl-disk.img
 ```
-sgx-lkl-disk create --size=100M --docker=MyDockerfile sgxlkl-disk.img
+
+To build an SGX-LKL disk image from an existing Docker image, run:
+```sh
+# This assumes a Docker image tagged 'my-app' exists on the system.
+sgx-lkl-disk create --size=100M --docker=my-app sgxlkl-disk.img
 ```
 
 #### Creating plain disk images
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -438,22 +446,78 @@ _To be added_
...
-Note that for the debugging options to have an effect, SGX-LKL must be built
-with `DEBUG=true`.
+The recommended way to configure SGX-LKL is through host config and app config JSON files.
+Using JSON files is also required when creating Confidential Containers (see next section).
+To create initial host and app config files in the current directory, run:
+```
+sgx-lkl-cfg create --disk sgxlkl-disk.img
+```
+Note that you may have to adapt the generated files to your needs.
+See the `tests/containers/cc` folder for example host and app config files.
```
prp:
```suggestion
Note that you may have to adapt the generated files to your needs
(see the `tests/containers/cc` folder for sample host and app config files).
```

README.md
```diff
@@ -438,22 +446,78 @@ _To be added_
...
+You can then run SGX-LKL using the host and app config files alone:
+```
+sgx-lkl-run-oe --hw-debug --host-config=host-config.json --app-config=app-config.json
+```
+Note that environment variables can still be used to specify additional options.
+This is especially useful for host configuration options that are only known
+at runtime in the target system where SGX-LKL will run, for example the name
+of the network tap device (`SGXLKL_TAP`).
```
prp, line 477: Do we want to keep environment variables that refer to the host_config in a RELEASE build?

README.md
```diff
@@ -438,22 +446,78 @@ _To be added_
...
+
+F. Confidential Container bundling for deployment
+-------------------------------------------------
+
+A Confidential Container is a Docker image that only contains the disk images
+and the host and app config files.
```
prp, line 483: Why would the CC contain the host_config? By definition, the host_config should be provided by the deployment environment, no?

README.md
```diff
@@ -438,22 +446,78 @@ _To be added_
...
+It is used for deploying an SGX-LKL application to a target system that has
+Docker as well as SGX-LKL installed on the host.
+This allows to update SGX-LKL without rebuilding all Docker images.
```
prp:
```suggestion
This allows to update SGX-LKL without rebuilding all Confidential Containers.
```

README.md
```diff
@@ -438,22 +446,78 @@ _To be added_
...
+Like Docker, SGX-LKL can be considered part of the host runtime environment.
```
prp: Docker daemon? Docker engine?

README.md
```diff
@@ -438,22 +446,78 @@ _To be added_
...
+
+To create a Confidential Container, run:
+```
+sgx-lkl-docker build-cc --name=my-app-cc --mode=debug --host-cfg=host-config.json --app-cfg=app-config.json
```
prp, line 491: I wonder if it is right that we have to specify the execution mode at CC build time. Why is that necessary?

README.md
```diff
@@ -438,22 +446,78 @@ _To be added_
...
+On the target system where the Confidential Container is run, install SGX-LKL
+from the published Debian packages (see A. Installing SGX-LKL-OE).
+Note that an SGX-LKL installation built from source (see B. Building SGX-LKL-OE from source)
+cannot be used directly as it is not a self-contained installation 
+(bundling required library dependencies).
+
+To run the Confidential Container, run:
+```
+docker run --rm --privileged --network=host \
+  -v /opt/sgx-lkl-debug:/opt/sgx-lkl-debug \
```
prp, line 507: See my comment above -- since we're specifying the SGX-LKL build, the mode above seems unnecessary.

Due to the app_config, attestation will fail if the host picks the wrong SGX-LKL build type.

prp (commented):
Overall this looks good -- I've left some comments, thanks.

README.md
letmaik: Yeah, I think this may be useful for infrastructure injecting information (like the tap device to use).

README.md
letmaik: Some things are known at CC build time, like the paths of the disks.

prp: I wonder if it wouldn't be simpler if we always just called the disks, `disk0.img`, `disk1.img` etc. and the CC provided all of them in that order. Is there a benefit of giving the container builder control over this? One benefit is that the host_config gives the container builder a mechanism to configure the host environment, and we may need this in the future.

README.md
letmaik: And since there is no difference in security for the host config I wouldn't make the behaviour different between build types, simply for consistency.

README.md
letmaik: Yeah, I wasn't sure which word fits best. Docker engine makes more sense.

README.md
letmaik: This determines the ENTRYPOINT that's used. E.g. `--mode=debug` results in `ENTRYPOINT ["/opt/sgx-lkl-debug/bin/sgx-lkl-run-oe", "--host-config=/host-cfg.json", "--app-config=/app-cfg.json"]`. In principle you can also leave out the entrypoint and specify it at runtime but that gets a bit unwieldy. Unfortunately it's not possible to do `-v /opt/sgx-lkl-debug:/opt/sgx-lkl` because the installation prefix must stay the same as it's baked in to the executable's interpreter path which must be absolute (`/opt/sgx-lkl-debug/lib/external/ld-linux-x86-64.so.2`). One thing we could simplify is to change the install prefixes to `/opt/sgx-lkl/<build-type>` so that at least the mount flag is always the same regardless of build type: `-v /opt/sgx-lkl:/opt/sgx-lkl`.

README.md
letmaik: For CCs this may work but I'd like to keep the flexibility of specifying explicit paths in SGX-LKL itself since it may be used in cases that are not strictly CCs, like in our Azure ML pipeline demo at //build where the disks were downloaded automatically at certain paths and would have to be copied around to fulfil the `disk<n>.img` naming scheme, which is suboptimal.

README.md
prp: Hmm, I don't think that the above is a great justification to have the `--mode` flag. You could argue that an entrypoint that simply refers to the SGX-LKL launcher is something that should come from the host environment and not be part of the CC. What would it take to provide the entrypoint at launch time?

letmaik:
```
docker run --rm --privileged --network=host \
  -v /opt/sgx-lkl-debug:/opt/sgx-lkl-debug \
  --entrypoint=/opt/sgx-lkl-debug/bin/sgx-lkl-run-oe
  my-app-cc \
  --host-config=/host-cfg.json --app-config=/app-cfg.json \
  --hw-debug
```
We could get rid of `--host-config` and `--app-config` if we extend SGX-LKL to pick up the paths from environment variables, then we can specify them in the CC Dockerfile via `ENV SGXLKL_APP_CONFIG_PATH=/app-cfg.json` etc.

letmaik: Note that I agree with you, it would be better to get rid of the `--mode` flag. If we need to add it in the future for other purposes again we can bring it back.

letmaik: Alternatively, instead of new environment variables, we could have SGX-LKL look for default paths in the current directory, like `host-config.json` and `app-config.json`.
The default ENTRYPOINT would be `/opt/sgx-lkl-release/bin/sgx-lkl-run-oe` and the default CMD would be `--hw-release`. Then the Docker run would be:
```
docker run --rm --privileged --network=host \
  -v /opt/sgx-lkl-release:/opt/sgx-lkl-release \
  my-app-cc
```

If you want to use something non-default, you can specify the entrypoint and SGX-LKL arguments, e.g here for running a debug build in sw mode:
```
docker run --rm --privileged --network=host \
  -v /opt/sgx-lkl-debug:/opt/sgx-lkl-debug \
  --entrypoint=/opt/sgx-lkl-debug/bin/sgx-lkl-run-oe \
  my-app-cc --sw-debug
```

letmaik: Sorry for the spam, but I'm a bit torn here. The only reason for running a CC in a different mode is for debugging/testing, and in that case you would very likely not use the same CC anyway as you would not want attestation-backed disk encryption, and maybe you also compile your app differently if its a native app (e.g. with debug symbols etc.). What `--mode` then does is two things: first, it makes running the CC easier by hard-coding some things, and second it could check for common mistakes, e.g. warning if the disks are not encrypted when using `--mode=release`. Are you sure we want to get rid of that?

README.md
letmaik: How about we change the install prefixes to `/opt/sgx-lkl/<mode>` so that the mode is not specified again when running the container, since you could then mount the whole parent folder?

letmaik:
I'll hold off on this PR until we established the right terminology.

letmaik:
Closing in favour of https://github.com/lsds/sgx-lkl/pull/453 and further PRs.

PR #278, opened 2020-05-20
Output message for how to run individual ltp test on failure

SeanTAllen:
Changes the Makefile target from run-single-hw to run-hw-single
as it makes it much easier to construct the informational message.


SeanTAllen:
This is failing due to the issues that #279 fixes.

prp (approved):
LGTM

PR #285, opened 2020-05-20
JSON Parser

mikbras:
This PR introduces **libjson**, a tiny JSON parser that was derived from the **odatalite** project I worked on a few years back. We reworked the code to be a standalone library whose footprint is less than 8 kilobytes.

Also, we reworked the library so that it no longer has any libc dependencies and it now accepts a user-defined allocator, which is passed in during initialization (thank you Christoph for all your changes and feedback that made this possible). This makes **libjson** usable in both enclave and host environments.

There is an additional library (**libjsonprint.a**) and header (**jsonprint.h**) for host applications that want to print JSON to standard output.


src/libjson/Makefile
```diff
@@ -0,0 +1,33 @@
...
+
+INCLUDES += -Iinclude
+
+SOURCES += $(wildcard *.c)
+
+OBJECTS = $(SOURCES:.c=.o)
+
+HEADERS += include/json.h include/jsonprint.h
+
+all: $(OBJECTS)
```
wintersteiger: This works fine, but it may be a little less effort to just add the .c/.h files to the existing build scripts to avoid issues with CFLAGS not being passed correctly or the wrong compiler being used (see e.g., reference to `gcc` below).

src/libjson/include/json.h
```diff
@@ -0,0 +1,140 @@
...
+** LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+** OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+** THE SOFTWARE.
+**
+**==============================================================================
+*/
+
+#ifndef _JSON_H
+#define _JSON_H
+
```
wintersteiger, line 32: No `#include`s, awesome!

src/libjson/json.c
```diff
@@ -0,0 +1,1217 @@
...
+    while ((*s1 && *s2) && (*s1 == *s2))
+    {
+        s1++;
+        s2++;
+    }
+
+    return *s1 - *s2;
+}
+
+static size_t _strlcpy(char* dest, const char* src, size_t size)
```
wintersteiger, line 207: Thanks for re-implementing these functions to cut the dependencies, that makes everything a bit easier!

src/libjson/json.c
```diff
@@ -0,0 +1,1217 @@
...
+            CHECK(_invoke_callback(
+                parser,
+                JSON_REASON_BEGIN_ARRAY,
+                JSON_TYPE_NULL,
+                NULL));
+
+            if (_get_array(parser) != JSON_OK)
+                RAISE(JSON_BAD_SYNTAX);
+
+            CHECK(_invoke_callback(
```
wintersteiger, line 1119: It would be awesome of the parser did a forward-scan to determine the length of the array, but that may be more effort than we're willing to invest.

src/libjson/json.c
```diff
@@ -0,0 +1,1217 @@
...
+        result = JSON_NO_MATCH;
+        goto done;
+    }
+
+    /* Compare the elements */
+    for (size_t i = 0; i < pattern_depth; i++)
+    {
+        if (_strcmp(pattern_path[i], "#") == 0)
+        {
+            if (_strtou64(&n, parser->path[i]) != 0)
```
wintersteiger: It would be nice if we could get the index/`n` for arrays too and I think it wouldn't be too hard to add. I think we'd just need an additional `size_t` for every path element (or a separate vector with the same nesting as the path).

src/libjson/jsonprint.c
```diff
@@ -0,0 +1,295 @@
...
+#include "json.h"
+#include "jsonprint.h"
+#include "raise.h"
+
+static void _Indent(FILE* os, size_t depth)
+{
+    size_t i;
+
+    for (i = 0; i < depth; i++)
+        fprintf(os, "  ");
```
wintersteiger: Cherry on top: if we made `fprintf` configurable, I think we could hook it up to SGX-LKL's logging facilities (which will change soon too). Would be nice for error messages, etc.

wintersteiger:
This looks great, I think I'll be able to use this immediately, as is!

For the attested configuration, we'll need a canonical version of json, so that it's as predictable as possible to compute enclave hashes before the enclave or container is started, as well as to avoid malleability problems. We've discussed some criteria, but I think the main ones that we want are no whitespace, sorted keys, and no duplicate keys. I think we could check those properties during parsing and reject anything that violates them immediately, but you likely have a better idea of how much effort this would take.

src/libjson/sample/main.c
```diff
@@ -0,0 +1,830 @@
...
+        {
+            break;
+        }
+        case JSON_REASON_VALUE:
+        {
+            const char** path = parser->path;
+            size_t depth = parser->depth;
+            unsigned long i;
+
+            if (json_match(parser, "keyslots.#.type", &i) == JSON_OK)
```
letmaik: The code below in this file seems like a lot of boiler plate. No way to cut this down? I'm used to nlohmann's C++ library which makes it nearly fun to use JSON.

src/libjson/Makefile
mikbras: Oh, good point Christoph! I added that change to the new **libjson2** branch and will create a new PR from that one (I failed to successfully merge with oe_port).

src/libjson/include/json.h
mikbras: Thanks!

src/libjson/json.c
mikbras: Thank you. I learned my lesson this time. Hopefully this is 100% standalone now. :)

src/libjson/json.c
mikbras: Good suggestion. Done. The paths array is now called nodes and includes the name of the element, the size of the array (or zero if none), the index of the array (or zero if none).

src/libjson/json.c
mikbras: Good idea. Done. The index is now included in the **parser->nodes[]** array.

src/libjson/jsonprint.c
mikbras: Nice suggestion. Done. This also eliminated the need for a second **libjsonprint.a** library. Now you can pass in a **write** callback function with a callback argument. This really cleaned things up but required implement a function to convert double-precision to string.

mikbras:
> This looks great, I think I'll be able to use this immediately, as is!
> 
> For the attested configuration, we'll need a canonical version of json, so that it's as predictable as possible to compute enclave hashes before the enclave or container is started, as well as to avoid malleability problems. We've discussed some criteria, but I think the main ones that we want are no whitespace, sorted keys, and no duplicate keys. I think we could check those properties during parsing and reject anything that violates them immediately, but you likely have a better idea of how much effort this would take.

The parser doesn't check for these now. Are these captured in an issue already or should I create one to track?

src/libjson/sample/main.c
mikbras: Not that I'm aware of without much more effort. What does nlohmann's parser do to simplify this?

mikbras:
Unfortunately I'm going to have to create a new PR since I have been unsuccessful at merging **oe_port** into my **libjson** branch. The following failed to merge all changes:

```
$ git merge oe_port
```

It did merge many changes but left out others. Sigh!

src/libjson/sample/main.c
letmaik: I guess, the fact that it uses C++? No worries. If Christoph is happy with it, it's fine.

wintersteiger: Yeah, it's C++ and it follows a completely different usage pattern, i.e. the parser internals are not exposed at all during parsing, you just get a nice map of keys to values. This parser just calls the callback and we have to do all the mapping and storing ourselves, which in C is a bit cumbersome, but faster and requires less memory.

src/libjson/json.c
wintersteiger: Awesome, thanks so much, this will make the client code much easier to write and read!

wintersteiger:
Shouldn't it be relatively easy to rebase this onto the current oe_port? I can't imagine that a rebase would throw many conflicts, except in the Makefiles perhaps. 

For the canonical json: We had an issue in the old repo but it hasn't been moved yet I think, I'll check and migrate it if so.

src/libjson/sample/main.c
mikbras: Thank you. I see. There is a back end for the **odatalite JSON parser** that does create parse trees. We might consider bringing that part over as well at some point. It is not C++ but at least you get a fully-formed tree. I'm not sure how helpful that would be.

mikbras:
> Shouldn't it be relatively easy to rebase this onto the current oe_port? I can't imagine that a rebase would throw many conflicts, except in the Makefiles perhaps.
> 
> For the canonical json: We had an issue in the old repo but it hasn't been moved yet I think, I'll check and migrate it if so.

Ah, yes, I will try a rebase. :)

mikbras:
I opened a new PR that references this one. I attempted a rebase but corrupted the branch further.

PR #294, opened 2020-05-21
Use exit code of LTP tests for pass/fail decision

hukoyu:
- Use exit code to decide LTP test pass/failure
- Remove failure identifier logic
- Disabling 23 failing tests after enabling exit code which were false positives.  We will investigate further these 23 tests and will fix the test issues and enable later (https://github.com/lsds/sgx-lkl/issues/296)
- LTP test count dropped from 461 downto 438 due to these 23 disabled tests.


hukoyu:
@letmaik this PR is for using LTP exit code for deciding test pass/failure instead of text match. Please review this and let's merge this first before merging any other LTP test patch PR.
  
cc @davidchisnall @prp @jxyang

hukoyu:
@davidchisnall @prp I think @letmaik is not available. could you please review and merge this PR if looks good. We want to merge this PR before merging any other LTP test enable PR.

davidchisnall (approved):

PR #295, opened 2020-05-21
Make timespec_diff usage less error-prone

SeanTAllen:
The signature is very odd and likely to result in an error. If you
want to get the time elapsed between two periods of time, the standard
for an API is to do end, start not start, end.


SeanTAllen:
@davidchisnall this is the change that came up in previous PR for changing the comment on the confusing previous version.

https://github.com/lsds/sgx-lkl/pull/141

davidchisnall (approved):

PR #298, opened 2020-05-22
Fix null-pointer dereference.

davidchisnall:
If we get a null pointer dereference in instruction fetch, OE injects a
trap for us and we try to give a helpful error message by disassembling
the instruction.  Doing so, we dereference a null pointer and get a
trap.  Bad things ensue.


davidchisnall:
@hukoyu, would you mind taking a look at this one?  CI isn't happy, but this change should be a no-op for anything that isn't crashing with an illegal instruction at address 0.

prp (approved):
LGTM

PR #299, opened 2020-05-22
Fix memory management issues in clone host support

davidchisnall:
When an LKL task does an exit system call, the kernel leaves it blocking
on the scheduler semaphore then frees the semaphore.  This leaves the
corresponding lthread on the enclave futex's sleeping queue.

This fixes the root cause of the intermittent failure in the clone test
if we deleted the thread immediately, so we can now simplify this code
and delete the thread, without pivoting the stack.


prp (approved):
LGTM

Issue #283, opened 2020-05-20
Investigate pipeline-local run gap for java read file test


hukoyu:
@davidchisnall says this test passes locally everytime and only fails on pipeline randomly. 
Investigate if there is gap between pipeline and local run that causing these flaky failures or this is a sgx-lkl product bug.

cc @hukoyu

SeanTAllen:
I set up a new dev machine the other day.

On my previous machine, I was unable to recreate any problems with JVM tests. However, on the new box, all of them have problems with being very slow. However, the problem is intermittent. I can do many runs in a row of SW mode without issue before hitting on that hangs for a long period of time. It looks like it is deadlocked but when I am patient enough, it sometimes continues. Other times, it has hung long enough that I give up and cancel it.

In [enclave_init.c](https://github.com/lsds/sgx-lkl/blob/oe_port/src/enclave/enclave_init.c#L209) hardcoding libc.user_tls_enabled to 0 makes the problem much less severe.

Locally with that change, all the JVM tests were snappy, however even with that change, I did get a JVM test timing out in CI. 

I'm not sure what the cause of this is my I wanted to report the symptom.

Note this problem applies to all JVM tests. Including the Hello World that was turned off for CI because it would also time out after 5 minutes sometimes. 

It should be noted that HW mode is much slower for me than SW mode without the above tls_enabled change. However in CI, both SW and HW sometimes time out and without the change, SW mode still feels sluggish. 

The sluggish nature of the run is not confined to a particular point in time during the applications lifecycle. My first thought when looking at it was that due to the low amount of memory allocated to various JVM bits in sgx-lkl-java, that it was GCing. Running with -verbose:gc as one of the JVM options did not show an excessive amount of GC, although I did see 1 `GC (Allocation Failure)`.

For me locally, doing, seems to correct the problem with HW mode without any code changes:

`  cd tools/kmod-set-fsgsbase; make set-cr4-fsgsbase`

From the docs:

```
Support for Thread-Local Storage (TLS) in hardware mode

On x86-64 platforms thread-local storage for applications and their initially
available dependencies is expected to be accessible via fixed offsets from the
current value of the FS segment base. Whenever a switch from one thread to
another occurs, the FS segment base has to be changed accordingly. Typically
this is done by the privileged kernel. However, with SGX the FS segment base is
explicitly set when entering the enclave (OFSBASE field of the TCS page) and
reset when leaving the enclave. SGX-LKL schedules application threads within
[...]

PR #306, opened 2020-05-26
Use the LKL futex implementation in userspace.

### LINKED ISSUE #180: Reove futex system call bypass
This depends on all consumers of futex-related APIs that run before userspace init moving to calling the enclave futex implementation directly, rather than via musl.

### LINKED ISSUE #209: Async signals aren't always delivered as expected
Asynchronous signals such as those that will be delivered by using the system calls `alarm` and `setitimer` are not delivered as expected. 

This can result in unexpected behavior from applications, for example, busybox ping. It will use alarm to wait to send it's ping and during that time isn't executing system calls so it hangs in sgx-lkl after the first ping is sent. 

@letmaik first reported this issue. It can be recreated by changing the `CC_APP` and `CC_APP_CMDLINE` options in the [curl test](https://github.com/lsds/sgx-lkl/blob/oe_port/tests/basic/network/Makefile) to:

```
CC_APP=/bin/ping
CC_APP_CMDLINE=${CC_APP} -c 2 google.com
```

Attached is a zip file with 3 different variations on how ping could be written that can help diagnosing. Those that rely on a timer will fail unless the timer is less than the amount of time it takes to receive a packet back from the pinged host. If a packet is received after the timer timeout, then another ping will be sent. Otherwise it hangs.

This behavior might be related to 

https://github.com/lsds/lkl/blob/21cd7e192cf2a97ea69230b9c0c34da15d157579/arch/lkl/kernel/syscalls.c#L195

where it appears that signals are only handled on syscalls. However, there is more investigation to be done there.
[ping-examples.zip](https://github.com/lsds/sgx-lkl/files/4586698/ping-examples.zip)

My suspicion is that if we checked for signals in [`futex_tick`](https://github.com/lsds/sgx-lkl/blob/oe_port/src/sched/futex.c#L48) or somewhere in the main scheduler loop, that this issue would go away but that hasn't been verified and that might be addressing a symptom rather than underlying cause. More investigation is needed.

### LINKED ISSUE #305: syscall_SYS_futex does not implement FUTEX_WAKE_OP
The FUTEX_WAKE_OP operation has existed since Linux 2.6.14 and is intended for efficient implementations of `pthread_cond_signal` (though it is quite rarely used).  This is probably low priority, but we should track it for Linux compat.

davidchisnall:
We currently have two implementations of futexes, one provided by Linux
(which was not being built) and one in `src/enclave`.  The `src/enclave`
version was incomplete (for example, not implementing `FUTEX_WAKE_OP`)
and was under-tested.  After this change:

 - The `enclave_futex` implementation is used only for the LKL host ops.
 - The Linux futex is used for everything in userspace.

This should now give us parity with Linux for futex operations (and
ensure that we pick up new ones when they are added upstream, for
example the proposed operation to wait on multiple futexes).

This does not fix #209, but hopefully simplifies the fix because there
should now be only one way for a userspace thread to block.  It appears
as if the thread is blocking correctly in userspace and the scheduler is
receiving ticks, but is not waking up the blocking thread.

Fixes #180
Fixes #305


davidchisnall:
Note: The lsds/sgx-lkl-musl#13 should be merged and the submodule commit bumped to the new hash before this PR is merged.

SeanTAllen:
In case you haven't looked yet, the ltp failure was:

2020-05-26T11:26:09.1532304Z /opt/sgx-lkl-nonrelease/bin/sgx-lkl-run-oe --hw-debug sgxlkl-miniroot-fs.img /ltp/testcases/kernel/syscalls/getppid/getppid01 : TIMED OUT after 60 secs
2020-05-26T11:26:09.1540583Z 'kernel-syscalls-getppid-getppid01-(nonrelease)-(run-hw)-(8-ethreads)' failed.
2020-05-26T11:26:09.1548529Z 'make run-hw-single test=/ltp/testcases/kernel/syscalls/getppid/getppid01' can be used to test this individually failing ltp test

SeanTAllen (approved):

SeanTAllen:
@davidchisnall I'm going to see if I can replicate that ltp failure locally.

SeanTAllen:
I can't replicate the failed test locally.

davidchisnall:
It seems to be intermittent.  It passed the first run, failed the second for me, and then didn't fail again.

davidchisnall:
I have now pushed an update to the submodule, so this is ready to merge if LTP is happy...

davidchisnall:
The getppid test is about as simple as hello world and doesn't test anything that this PR would modify.  I wonder if it's intermittently hitting an existing concurrency bug?

SeanTAllen:
Based on my seeing similar failures across other tests in other PRs, I think its either an existing concurrency bug or a problem in the CI environment as I can never recreate locally.

SeanTAllen:
There's a CI failure which is the known JVM issue #283, but some of the JVM tests are still running so it failed in that intermittent way this time.

SeanTAllen:
Different LTP failed with non-release this time:

2020-05-26T13:37:03.4857363Z 'make run-hw-single test=/ltp/testcases/kernel/syscalls/setegid/setegid02' can be used to test this individually failing ltp test

SeanTAllen:
Another set of ltp timeouts. Different this time.

PR #314, opened 2020-05-27
Move the LKL host layer to using OE malloc.

davidchisnall:
I tried this previously, but reverted it because of crashes.  The
crashes were caused by the fact that LKL allocate memory for the kernel
and for `lkl_vprintf` buffers.  The former is larger than the OE heap,
so fails, resulting in LKL refusing to start.  To avoid this, assume
that there is only going to be one allocation > 1MiB and that this is
the kernel heap.  Special case it and provide it via `enclave_mmap`,
forward everything else to `oe_malloc`.  Other structures should be safe
to allocate via `oe_malloc`.

### DIFF
--- a/src/lkl/posix-host.c
+++ b/src/lkl/posix-host.c
@@ -23,6 +23,7 @@
 #include "enclave/sgxlkl_t.h"
 #include "lkl/iomem.h"
 #include "lkl/jmp_buf.h"
+#include "openenclave/corelibc/oemalloc.h"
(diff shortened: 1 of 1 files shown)


davidchisnall:
Without this, the futex change was occasionally causing crashes and probably deadlock when we were trying to go through the kernel to do malloc for LKL trace messages before the kernel was ready.

SeanTAllen (approved):

PR #320, opened 2020-05-27
Remove libc dependencies from enclave_util.c

SeanTAllen:
As part of work on #151, this commit breaks enclave_util.c apart into
the parts only needed by lkl & sgx-lkl-musl and parts that are also
used by the enclave.

All parts that aren't needed for the enclave have been moved to lkl_util.c

This removes the libc dependency on snprintf from enclave_util and advances
towards our goal of removing libc dependencies from src/enclave.


SeanTAllen:
Corresponding sgx-lkl-musl PR: https://github.com/lsds/sgx-lkl-musl/pull/14

davidchisnall (approved):

PR #323, opened 2020-05-27
libjson

mikbras:
This PR supersedes PR #285 (whose branch I corrupted with a failed rebase operation). Please see the original PR for history.


mikbras:
@wintersteiger: I changed the **json_match()** signature because it was only able to match one number ("#"). If there were two numbers, there was no way to get the second one. I hope the change is not too much of an inconvenience. Something like this:

```
if (json_match(parser, "digests.#.digest", &i) == JSON_OK)
{
    ...
}
```

now becomes:

```
if (json_match(parser, "digests.#.digest") == JSON_OK)
{
    /* Get the number for the '#' element */
    uint64_t i = parser->path[1].number;
    ...
}
```

And now we can do things like this:

```
if (json_match(parser, "a.#.b.#.c") == JSON_OK)
{
    uint64_t n = parser->path[1].number;
    uint64_t m = parser->path[3].number;
    ...
}
```

src/libjson/README.md
```diff
@@ -0,0 +1,19 @@
...
+=======
+
+This package defines a simple JSON parser library. Rather than building a
+parse tree, it invokes a user-defined callback as elements are recognized
+during parser.
+
+Samples
+=======
+
+The **sample** directory uses the parser to populate a multi-level C structure.
```
jxyang: This needs to be updated to sample1 and sample2.

src/libjson/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+The **jsonprint** sample formats JSON source and prints it to standard output.
```
jxyang, line 15: It appears to me both `sample2` and `jsonprint` output a `beautified` json representation, except the former uses `json_parser_parse`, and the later uses `json_print`. What are the other differences?

src/libjson/include/json.h
```diff
@@ -0,0 +1,179 @@
...
+    json_allocator_t* allocator;
+    void (*trace)(
+        json_parser_t* parser,
+        const char* file,
+        unsigned int line,
+        const char* func,
+        const char* message);
+};
+
+json_result_t json_parser_init(
```
jxyang, line 152: Some comments that describe the functions in this header would be nice.

src/libjson/jsonprint/main.c
```diff
@@ -0,0 +1,125 @@
...
+#include <json.h>
+#include <sys/stat.h>
+#include <string.h>
+
+static void _write(void* stream, const void* buf, size_t count)
+{
+    fwrite(buf, 1, count, (FILE*)stream);
+}
+
+int _load_file(
```
jxyang: This function is duplicated in several places, can we put it in something like `libjson/util.c`. sgx-lkl could benefit from it too if we expose the API.

wintersteiger:
> I changed the **json_match()** signature ... 

Sure, that's perfectly fine!

src/libjson/json.c
```diff
@@ -0,0 +1,1736 @@
...
+    else if (p[0] == '-')
+    {
+        negative = 1;
+        p++;
+    }
+
+    /* If base is zero, deduce the base from the prefix. */
+    if (base == 0)
+    {
+        if (p[0] == '0' && (p[1] == 'x' || p[1] == 'X'))
```
wintersteiger, line 339: Sorry for the additional complications! Technically I think, JSON does not support hex or octal ints and it allows only one leading zero, if it's before the decimal point of a float. It looks like we'll encode must numbers into strings anyways (also limited to 2^53, malleability issues, etc), but I think we should tighten the parser down to reject everything that's not in the standard.

src/libjson/README.md
mikbras: Oh, thanks! Fixed.

src/libjson/include/json.h
mikbras: :) I added a bunch. Thanks.

src/libjson/jsonprint/main.c
mikbras: I moved it to libjson/common and it is shared by the samples. But I cannot put it in **libjson** because it has no dependencies on a C runtime and so no way to load a file.

src/libjson/json.c
mikbras: Thank you Christoph. We are okay because we only call it with base 10 (so that code is never executed). Non base 10 expressions will fail.

```
un->integer = _strtol(start, &end, 10);
```

mikbras:
Thank you both for your comments. I did the following:
(1) Added code documentation for json.h.
(2) Move load_file into a common directory so there would not be 3 versions.
(3) Verified that **_strtol()** is only called with a base 10 argument.
(4) Fixed README.md to define both samples.

src/libjson/README.md
mikbras: Right. Sample2 shows how to write a parser that beautifies code. The second is a utility for doing the same.

wintersteiger (approved):
This looks good to me. Thanks for the additional changes and checks!

mikbras:
I accidentally merged thinking the squash button was to simply squash the commits.

wintersteiger:
Haha, the button is just too big and too green :-) No worries though, I think this was ready to be merged anyways.

mikbras:
> Haha, the button is just too big and too green :-) No worries though, I think this was ready to be merged anyways.

Hahaha! :) It was irresistible!

Issue #265, opened 2020-05-19
Use JSON schemas for host and app config files


letmaik:
This would make writing those files a lot nicer in IDEs like VS Code.

- [x] Create schemas #324 
- [ ] Validate all configs in tests during CI
- [ ] Allow easy use from VS Code

PR #324, opened 2020-05-27
App and host config JSON schemas & sgx-lkl-cfg validate

letmaik:
First steps towards https://github.com/lsds/sgx-lkl/issues/265.

There are some open issues about versioning, CI integration, and IDE integration but I think this PR is good enough as a first step.

Note that the host config schema is mostly empty on purpose as it will massively change soon.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -135,7 +135,7 @@ fsgsbase-kernel-module:
 	make -C ${TOOLS}/kmod-set-fsgsbase
 
 install:
-	mkdir -p ${PREFIX}/bin ${PREFIX}/lib ${PREFIX}/lib/gdb ${PREFIX}/tools ${PREFIX}/tools/kmod-set-fsgsbase
+	mkdir -p ${PREFIX}/bin ${PREFIX}/lib ${PREFIX}/lib/gdb ${PREFIX}/share ${PREFIX}/share/schemas ${PREFIX}/tools ${PREFIX}/tools/kmod-set-fsgsbase
 	cp $(BUILD_DIR)/$(SGXLKL_LIB_TARGET_SIGNED) $(PREFIX)/lib
 	cp $(BUILD_DIR)/$(SGXLKL_RUN_TARGET) $(PREFIX)/bin
 	cp $(TOOLS)/sgx-lkl-java $(PREFIX)/bin
@@ -146,6 +146,8 @@ install:
 	cp $(TOOLS)/gdb/sgx-lkl-gdb $(PREFIX)/bin
 	cp $(TOOLS)/gdb/gdbcommands.py $(PREFIX)/lib/gdb
 	cp $(TOOLS)/gdb/sgx-lkl-gdb.py $(PREFIX)/lib/gdb
+	cp ${TOOLS}/schemas/app-config.schema.json $(PREFIX)/share/schemas
+	cp ${TOOLS}/schemas/host-config.schema.json $(PREFIX)/share/schemas
 	cp ${TOOLS}/kmod-set-fsgsbase/mod_set_cr4_fsgsbase.ko $(PREFIX)/tools/kmod-set-fsgsbase/
 
 uninstall:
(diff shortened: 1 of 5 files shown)


wintersteiger:
That looks great, it's a good idea to wire that into `*-cfg` directly. 

I'm working on the attested enclave config and I'm almost completely decided that this will follow a new, non-user-readable schema to get all the security properties we want. I currently merge them in the launcher into this new format, which is then passed via EEID. Would the Python `jsonschema` allow us to check some of those properties like sortedness, whitespace freeness, etc, or would we have to add that ourselves anyways?

letmaik:
The validation is not about canonicalization for attestation purposes. The latter is done via `sgx-lkl-cfg finalize` which currently just sorts keys and removes whitespace. I believe the idea was to be able to predict the EEID by relying on this canonical format, without modifying it further when loaded in SGX-LKL, because otherwise it gets too hard. If you want to change that format could you open an issue? Regarding any checking, yes, that has to be done inside SGX-LKL as well because it would be untrusted otherwise. The schema validation is merely to catch user errors and aid in writing the configs in IDEs, not to enforce attestation constraints or similar, since that wouldn't be the right place anyway.

wintersteiger:
Yeah, that's right. I'm not opening an issue for the EEID config format yet because it currently changes about once per hour. I currently use the SGX-LKL launcher to compile that config, but we can obviously do the same in another tool and the launcher can dump it too. 

Of course, we need checks elsewhere, but it would be nice if it could be done in one tool instead of having schema checks in one and attestation constraints in different one.

letmaik:
(EEID config format: https://github.com/lsds/sgx-lkl/issues/327)

Regarding validation, I think currently the only other thing that makes sense in `sgx-lkl-cfg validate` is to check for duplicate keys, just to fail early in the process.

Makefile
```diff
@@ -146,6 +146,8 @@ install:
 	cp $(TOOLS)/gdb/sgx-lkl-gdb $(PREFIX)/bin
 	cp $(TOOLS)/gdb/gdbcommands.py $(PREFIX)/lib/gdb
 	cp $(TOOLS)/gdb/sgx-lkl-gdb.py $(PREFIX)/lib/gdb
+	cp ${TOOLS}/schemas/app-config.schema.json $(PREFIX)/share/schemas
+	cp ${TOOLS}/schemas/host-config.schema.json $(PREFIX)/share/schemas
```
jxyang, line 150: Why in source tree, the schemas are under `tools`, while in the installation, the schemas are under the /opt/sgx-lkl?

tools/schemas/app-config.schema.json
```diff
@@ -0,0 +1,98 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "type": "object",
+    "title": "SGX-LKL app config schema",
+    "required": [
+        "run",
+        "args",
```
jxyang: With a default value, why is `args` required?

tools/schemas/app-config.schema.json
```diff
@@ -0,0 +1,98 @@
...
+            "type": "array",
+            "description": "Application arguments",
+            "default": [],
+            "items": {
+                "type": "string"
+            }
+        },
+        "cwd": {
+            "type": "string",
+            "description": "Absolute path to the working directory",
```
jxyang, line 26: What's the default of cwd if not specified?

Makefile
letmaik: `tools` in the installation is currently only used for the kernel module, and I don't think it belongs there. I think standard installation folders for data is `share/` but let me know if there's a better folder.

tools/schemas/app-config.schema.json
letmaik: Good point. I'll remove it.

tools/schemas/app-config.schema.json
letmaik: That would be `/`, I'll add it.

davidchisnall (approved):

Issue #245, opened 2020-05-14
[Tests] Java Runtime Environment crashes with fatal error with network stream 


hukoyu:
Added new java scenario test with this PR https://github.com/lsds/sgx-lkl/pull/244 and caught this exception. The test just opens a network stream to https://www.microsoft.com address and reads from the stream line by line. 

Please look at the PR for sample code used and Makefile.

Exception:
```[[  SGX-LKL ]] libc_start_main_stage2(): Calling app main: /usr/bin/java
OpenJDK 64-Bit Server VM warning: Can't detect primordial thread stack location - find_vma failed
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x00007f763d035278, pid=36, tid=0x00007f7600b31e00
#
# JRE version: OpenJDK Runtime Environment (8.0_212-b04) (build 1.8.0_212-b04)
# Java VM: OpenJDK 64-Bit Server VM (25.212-b04 mixed mode linux-amd64 compressed oops)
# Derivative: IcedTea 3.12.0
# Distribution: Custom build (Sat May  4 17:33:35 UTC 2019)
# Problematic frame:
# J 668 C2 java.util.Arrays.equals([B[B)Z (54 bytes) @ 0x00007f763d035278 [0x00007f763d035260+0x18]
#
# Failed to write core dump. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
#
bad count while changing owner
2020-05-14T01:32:57.000000Z [(H)ERROR] tid(0x7f76ecb38700) | :OE_ENCLAVE_ABORTING [/home/hukoyu/openenclave/host/calls.c:oe_call_enclave_function_by_table_id:91]
Makefile:26: recipe for target 'run-hw' failed```

cc @prp @davidchisnall @letmaik

hukoyu:
Here is the sample code in the https://github.com/lsds/sgx-lkl/blob/496384f4d6025ce94d13f3fa7825f0c54bc4a360/tests/languages/java/network/app/MainApp.java

hukoyu:
Here is the full log of run-hw for this sample java app
[java-network.log.log](https://github.com/lsds/sgx-lkl/files/4625589/java-network.log.log)

prp:
@hukoyu, could you do the following:
- run this with gdb and include a stack trace at the time when the SIGSEGV has happened?
[...]

PR #330, opened 2020-05-28
Add debugging doc

letmaik:
@prp @davidchisnall @SeanTAllen @jxyang Anything else you would like to see in this doc? Any useful debugging techniques I'm not aware of?

### DIFF
--- a/docs/Debugging.md
+++ b/docs/Debugging.md
@@ -22,10 +22,6 @@ If you used a different directory, run `. <OE_PREFIX>/share/openenclave/openencl
 
 Note that debugging enclave code (SGX-LKL as well as application code) is not possible in RELEASE builds of SGX-LKL.
 
-### Encrypted disks
-
-TODO add note on how to handle situations when disk keys are retrieved from a remote service  (either re-create disks unencrypted, re-create disks encrypted but with debug-enabled key release policy (closest to production), or hard-code key in app config)
-
 ### Application type
 
 Debugging of non-native applications like Python, Java or .NET applications is currently not possible.
--- a/docs/Debugging.md
+++ b/docs/Debugging.md
@@ -22,10 +22,6 @@ If you used a different directory, run `. <OE_PREFIX>/share/openenclave/openencl
 
 Note that debugging enclave code (SGX-LKL as well as application code) is not possible in RELEASE builds of SGX-LKL.
 
-### Application type
-
-Debugging of non-native applications like Python, Java or .NET applications is currently not possible.
-
 ### Docker deployment container
 
 The instructions below assume that disk images and the host/app configurations are available outside of a Docker deployment container. Debugging of deployment containers directly is not supported yet as they do not contain the necessary debugging tools.
--- a/docs/Debugging.md
+++ b/docs/Debugging.md
@@ -20,7 +20,7 @@ If you used a different directory, run `. <OE_PREFIX>/share/openenclave/openencl
 
 ### Build mode
 
-Debugging enclave code (SGX-LKL as well as the application code) is only possible in debug and non-release builds of SGX-LKL.
+Note that debugging enclave code (SGX-LKL as well as application code) is not possible in RELEASE builds of SGX-LKL.
 
 ### Encrypted disks
 
@@ -32,7 +32,7 @@ Debugging of non-native applications like Python, Java or .NET applications is c
 
 ### Docker deployment container
 
-The instructions below assume that disk images and configuration are available outside of a Docker deployment container. Debugging of such deployment containers is not supported currently as they do not contain the necessary debugging tools.
+The instructions below assume that disk images and the host/app configurations are available outside of a Docker deployment container. Debugging of deployment containers directly is not supported yet as they do not contain the necessary debugging tools.
 
 ## Option A: Debugging with an SGX-LKL installation
 
@@ -69,7 +69,7 @@ If possible, debugging should be done in hardware mode to be as close to a produ
 
 ### Ignore SIGILL
 
-In some cases `sgx-lkl-gdb` still incorrectly stops at illegal instructions that are emulated inside the enclave (see e.g. https://github.com/lsds/sgx-lkl/issues/167). To avoid that, run `handle SIGILL pass nostop noprint` as first command in the active `gdb` session.
+In some cases `sgx-lkl-gdb` stops at illegal instructions that are emulated inside the enclave (see e.g. https://github.com/lsds/sgx-lkl/issues/167). To avoid that, run `handle SIGILL pass nostop noprint` as the first command in the active `gdb` session.
 
 ### Optimized code
 
--- a/docs/Debugging.md
+++ b/docs/Debugging.md
@@ -24,17 +24,21 @@ Debugging enclave code (SGX-LKL as well as the application code) is only possibl
 
 ### Encrypted disks
 
-TODO add note on how to handle situations when disk keys are retrieved from a remote service that would not release them to debug/non-release builds of SGX-LKL (either re-create disks unencrypted, or, if available locally, hard-code key in app config for local debugging)
+TODO add note on how to handle situations when disk keys are retrieved from a remote service  (either re-create disks unencrypted, re-create disks encrypted but with debug-enabled key release policy (closest to production), or hard-code key in app config)
 
-### Confidential Containers
+### Application type
 
-The instructions below assume that disk images and configuration are available outside of a Confidential Container. Debugging of Confidential Containers (via Docker) is not supported currently.
+Debugging of non-native applications like Python, Java or .NET applications is currently not possible.
+
+### Docker deployment container
+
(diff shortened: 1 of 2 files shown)


SeanTAllen:
@letmaik nothing comes to mind at the moment

docs/Debugging.md
```diff
@@ -0,0 +1,81 @@
...
+The instructions below assume that Open Enclave was installed in the default directory `/opt/openenclave`.
+If you used a different directory, run `. <OE_PREFIX>/share/openenclave/openenclaverc` first.
+
+### Build mode
+
+Debugging enclave code (SGX-LKL as well as the application code) is only possible in debug and non-release builds of SGX-LKL.
+
+### Encrypted disks
+
+TODO add note on how to handle situations when disk keys are retrieved from a remote service that would not release them to debug/non-release builds of SGX-LKL (either re-create disks unencrypted, or, if available locally, hard-code key in app config for local debugging)
```
jxyang: We should be able to debug the key release process. Ideally, the key vault should have a "debugger friendly" release policy. During dev/debug cycles, the disk encryption keys should be created with such policies.

docs/Debugging.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Prerequisites
+
+### Open Enclave
+
+`sgx-lkl-gdb` relies on the Open Enclave gdb plugin.
+[Install Open Enclave](https://github.com/openenclave/openenclave/blob/master/docs/GettingStartedDocs/install_oe_sdk-Ubuntu_18.04.md) if necessary before continuing.
+
+The instructions below assume that Open Enclave was installed in the default directory `/opt/openenclave`.
+If you used a different directory, run `. <OE_PREFIX>/share/openenclave/openenclaverc` first.
```
jxyang, line 19: The statement implies we don't need to source `/opt/openenclave/share/openenclave/openenclaverc`.

docs/Debugging.md
```diff
@@ -0,0 +1,81 @@
...
+
+The instructions below assume that disk images and configuration are available outside of a Confidential Container. Debugging of Confidential Containers (via Docker) is not supported currently.
+
+## Option A: Debugging with an SGX-LKL installation
+
+The instructions below assume that SGX-LKL was installed in `/opt/sgx-lkl`.
+
+To debug an application, invoke `sgx-lkl-gdb` as follows:
+```
+/opt/sgx-lkl/bin/sgx-lkl-gdb --args /opt/sgx-lkl/bin/sgx-lkl-run-oe --hw-debug root_disk.img /app
```
jxyang, line 35: This is not accurate enough. It should be `<path-to-app>`, plus optionally parameters to `app`

jxyang: The general idea is to prefix `sgx-lkl-gdb --args` to whatever command we used to launch the application in sgx-lkl.

docs/Debugging.md
```diff
@@ -0,0 +1,81 @@
...
+
+### Visual Studio Code
+
+Visual Studio Code allows to use a custom `gdb` command for debugging.
+To use `sgx-lkl-gdb`, add `"miDebuggerPath": "/opt/sgx-lkl/bin/sgx-lkl-gdb"` (change path if necessary) in your `.vscode/launch.json` file.
+Tip: Use the "Visual Studio Code Remote - SSH" extension to debug on a remote VM with SGX support.
+
+### Run mode
+
+Apart from obvious differences (security, performance, ...) there are subtle differences between SGX-LKL software mode (`--sw-debug`) and hardware mode (`--hw-debug`).
```
jxyang: It's probably simpler to say in `sw-debug` mode we simulate the SGX hardware.

docs/Debugging.md
```diff
@@ -0,0 +1,81 @@
...
+### Ignore SIGILL
+
+In some cases `sgx-lkl-gdb` still incorrectly stops at illegal instructions that are emulated inside the enclave (see e.g. https://github.com/lsds/sgx-lkl/issues/167). To avoid that, run `handle SIGILL pass nostop noprint` as first command in the active `gdb` session.
+
+### Optimized code
+
+In general, debugging optimized code is not recommended.
+For an optimal debugging experience, SGX-LKL should be built in debug mode.
+If the application code should be debugged as well and it is native code,
+then the application should be built without optimizations and with debug symbols.
```
jxyang, line 71: We should mention somewhere at the top that debugging non-native applications is currently not possible. Debugging non-native language runtimes is also challenging because they typically are not built with symbols, e.g., https://github.com/lsds/sgx-lkl/issues/245

docs/Debugging.md
letmaik: Agree, good point. I changed the TODO.

docs/Debugging.md
letmaik: Yeah, that's the case. `sgx-lkl-gdb` checks the default directory. Should I mention that?

docs/Debugging.md
letmaik: I changed the text and made it clear that the command is just an example.

docs/Debugging.md
letmaik: Changed, also added a concrete example of what isn't simulated currently.

docs/Debugging.md
letmaik: I added a subsection at the top about non-native applications. I don't think we have to go into language runtimes as that's a very advanced topic and people would probably know what they are doing in that case.

docs/Debugging.md
```diff
@@ -0,0 +1,85 @@
...
+
+`sgx-lkl-gdb` relies on the Open Enclave gdb plugin.
+[Install Open Enclave](https://github.com/openenclave/openenclave/blob/master/docs/GettingStartedDocs/install_oe_sdk-Ubuntu_18.04.md) if necessary before continuing.
+
+The instructions below assume that Open Enclave was installed in the default directory `/opt/openenclave`.
+If you used a different directory, run `. <OE_PREFIX>/share/openenclave/openenclaverc` first.
+
+### Build mode
+
+Debugging enclave code (SGX-LKL as well as the application code) is only possible in debug and non-release builds of SGX-LKL.
```
prp:
```suggestion
Note that debugging enclave code (SGX-LKL as well as application code) is not possible in RELEASE builds of SGX-LKL.
```

docs/Debugging.md
```diff
@@ -0,0 +1,85 @@
...
+
+### Encrypted disks
+
+TODO add note on how to handle situations when disk keys are retrieved from a remote service  (either re-create disks unencrypted, re-create disks encrypted but with debug-enabled key release policy (closest to production), or hard-code key in app config)
+
```
prp: I would remove this, as the details will depend on how key release is handled.

docs/Debugging.md
```diff
@@ -0,0 +1,85 @@
...
+### Application type
+
+Debugging of non-native applications like Python, Java or .NET applications is currently not possible.
```
prp: I don't understand what that means. What are "non-native applications"?

docs/Debugging.md
```diff
@@ -0,0 +1,85 @@
...
+
+### Docker deployment container
+
+The instructions below assume that disk images and configuration are available outside of a Docker deployment container. Debugging of such deployment containers is not supported currently as they do not contain the necessary debugging tools.
```
prp:
```suggestion
The instructions below assume that disk images and the host/app configurations are available outside of a Docker deployment container. Debugging of deployment containers directly is not supported yet as they do not contain the necessary debugging tools.
```

docs/Debugging.md
```diff
@@ -0,0 +1,85 @@
...
+Tip: Use the "Visual Studio Code Remote - SSH" extension to debug on a remote VM with SGX support.
+
+### Run mode
+
+When using `--sw-debug` then SGX hardware is simulated. However, not all aspects are simulated, for example, certain CPU instructions are invalid within SGX, but would execute fine in simulation mode.
+If possible, debugging should be done in hardware mode to be as close to a production environment as possible.
+
+### Ignore SIGILL
+
+In some cases `sgx-lkl-gdb` still incorrectly stops at illegal instructions that are emulated inside the enclave (see e.g. https://github.com/lsds/sgx-lkl/issues/167). To avoid that, run `handle SIGILL pass nostop noprint` as first command in the active `gdb` session.
```
prp:
```suggestion
In some cases `sgx-lkl-gdb` stops at illegal instructions that are emulated inside the enclave (see e.g. https://github.com/lsds/sgx-lkl/issues/167). To avoid that, run `handle SIGILL pass nostop noprint` as the first command in the active `gdb` session.
```

prp (approved):
LGTM, left some minor tweaks.

docs/Debugging.md
letmaik: Anything that runs in a language runtime, like Python scripts.

prp: Ok, in that case, we should remove the statement that debugging of these applications is not possible. It actually depends how a debugger for a specific runtime works, and there is probably no point in going into that much detail here. (For example, debugging functionality that is part of an interpreter (or uses gdb underneath) may actually.) We should just claim `gdb` support and leave it at that.

Issue #332, opened 2020-05-28
Simplify samples


letmaik:
Samples are currently a bit hard to understand because they use makefiles and have sparse READMEs.

The idea is to simplify the samples as much as possible to make them easier to digest for humans. The README of each sample should be re-written to be more verbose, more like tutorials, and list all commands explicitly without relying on automation (except using SGX-LKL tooling). ~At the same time they are still runnable unattended by having two scripts `build.sh` and `run.sh` in each sample folder which we may use as part of crash tests in CI.~ EDIT: CI automation will be looked at later.

prp:
Yes, I agree that the README should show all the commands, so a user understands how they are run. Given that, we can even dispense with having scripts at all?

letmaik:
Scripts are needed to test samples in CI. We don't need to reference the scripts in the READMEs, but they are important. Except if every sample uses the exact same commands of course, but I don't have a good overview yet if that's the case.

prp:
As mentioned before, I would separate samples from tests because they serve very different purposes: samples are for human consumption, but tests need to integrate with the CI set-up.

letmaik:
I'm sorry, but I disagree. Testable samples/documentation is a thing, and we should do it as much as possible.

prp:
I agree with using the CI to test both the tests and the samples, but not with merging both into one.

letmaik:
Merging?

prp:
> Merging?

Not having separate directories (which is not what you suggest :-)

letmaik:
I've merged the first sample rewrite in #333 and am now looking at the other samples. I think we need to rethink them completely, even the hello world sample is not right in my opinion. It should be something much simpler not involving compilation of C code. In general I also don't want to replicate tests. I think it would be much nicer to have samples which focus on certain tooling tasks or goals. Including specific samples for certain application workloads is ok but only if there is something special about them, e.g. in terms of the configuration or Dockerfile. @prp What do you think?

prp:
> I've merged the first sample rewrite in #333 and am now looking at the other samples. I think we need to rethink them completely, even the hello world sample is not right in my opinion. It should be something much simpler not involving compilation of C code. In general I also don't want to replicate tests. I think it would be much nicer to have samples which focus on certain tooling tasks or goals. Including specific samples for certain application workloads is ok but only if there is something special about them, e.g. in terms of the configuration or Dockerfile. @prp What do you think?

[...]

PR #333, opened 2020-05-28
Re-work Python sample as template for all samples

letmaik:
See https://github.com/lsds/sgx-lkl/issues/332 for context.

Feedback welcome. The idea is to re-work all samples along those lines.


samples/languages/python/README.md
```diff
@@ -1,20 +1,43 @@
 SGX-LKL-OE Python and NumPy Sample Application
 ==============================================
 
-1. First build the Python and NumPy file system image:
+In this sample you will run a simple Python application in SGX-LKL and get used to the available tools.
+The aim of this sample is to verify that an application works in SGX-LKL without considering more advanced topics like encryption, attestation, networking, and deployment via Docker.
```
jxyang: an application -> an `Python` application

samples/languages/python/README.md
```diff
@@ -1,20 +1,43 @@
...
 
-```
-make
+The Python application is packaged as a Docker image. Let's first verify that it works outside of SGX-LKL:
+
+```sh
+docker build -t pythonapp .
+docker run --rm pythonapp
+# Output:
+# Confidential Computing using Intel SGX in Python with NumPy...
```
jxyang: Why the message says "using Intel SGX"? The host mode doesn't use SGX, right?

samples/languages/python/README.md
```diff
@@ -1,20 +1,43 @@
...
+#  [9800 9801 9802 ... 9897 9898 9899]
+#  [9900 9901 9902 ... 9997 9998 9999]]
 ```
 
-2. Run the example application using:
+The Python script is located in the `app/` folder, the Docker image is defined in the `Dockerfile`. Have a look at both files to understand what is happening.
 
-```
-make run-hw
+We can now convert the Docker image into a filesystem image which SGX-LKL can use:
```
jxyang: Is this a typo? "which" -> "with"?

samples/languages/python/README.md
```diff
@@ -1,20 +1,43 @@
...
+
+```sh
+sgx-lkl-disk create --docker=pythonapp --size=300M pythonapp.img
 ```
 
-or 
+The filesystem image contains only the files of the Docker image, but not its runtime metadata (like entrypoint, working directory, and environment variables). To run SGX-LKL we need to create configuration files containing these information. Fortunately, during conversion, this runtime metadata was queried from Docker and stored in special files named `pythonapp.img.*`. The `sgx-lkl-cfg` tool can use these files to generate an initial set of SGX-LKL host and application configuration files:
```
jxyang: These are very helpful info. But since they are applicable to all samples, I'd prefer to put in a README under `samples`, rather than in this particular sample.

samples/languages/python/build.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+export PATH="../../../tools:$PATH"
```
jxyang: What's the rational behind replacing Makefile with `build.sh` and `run.sh`?

hukoyu: Should we mention installing sgx-lkl (sudo -E make install) in README and using:
export PATH=/opt//sgx-lkl/bin:$PATH

samples/languages/python/run.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+export PATH="../../../build:$PATH"
```
hukoyu: Should we mention installing sgx-lkl (sudo -E make install) in README and using:
export PATH=/opt//sgx-lkl/bin:$PATH

samples/languages/python/README.md
```diff
@@ -1,20 +1,43 @@
...
 ```
-make run-sw
+Have a look at the generated `host-config.json` and `app-config.json` files and verify that they are as expected.
+
+We are now ready to run the application in SGX-LKL:
+
+```sh
+sgx-lkl-run-oe --host-config=host-config.json --app-config=app-config.json --hw-debug
 ```
+Use `--sw-debug` if your machine does not support SGX.
```
hukoyu, line 43: Should we also add info about how to debug with sgx-lkl-gdb?

hukoyu (approved):
Added a few recommendations.

samples/languages/python/README.md
letmaik: No, alternatively "that SGX-LKL can use". ... or something else altogether, like "We can now convert the Docker image into a filesystem image suitable for SGX-LKL:". Which one do you prefer?

samples/languages/python/build.sh
letmaik: Makefiles are too complicated for samples. `build.sh` and `run.sh` are just for CI purposes (eventually), and maybe we can get rid of them as well for all samples that follow the same pattern.

samples/languages/python/README.md
letmaik: I agree, although I think `samples/README.md` is not the right place either. It should probably go into a new `docs/Configuration.md` file. Do you agree? If so, then I'll do that in a separate PR first and then change it here.

samples/languages/python/README.md
letmaik: Changed to "SGX-LKL".

samples/languages/python/run.sh
letmaik: I think this should go into the installation instructions of the root README.
Created https://github.com/lsds/sgx-lkl/pull/342 for that.

samples/languages/python/README.md
letmaik: I'm adding a separate document about debugging in docs/. I think we don't need to replicate those bits in the samples necessarily.

samples/languages/python/build.sh
letmaik: I think this should go into the installation instructions of the root README.
Created #342 for that.

samples/languages/python/README.md
```diff
@@ -1,20 +1,43 @@
...
+#  [ 100  101  102 ...  197  198  199]
+#  [ 200  201  202 ...  297  298  299]
+#  ...
+#  [9700 9701 9702 ... 9797 9798 9799]
+#  [9800 9801 9802 ... 9897 9898 9899]
+#  [9900 9901 9902 ... 9997 9998 9999]]
 ```
 
-2. Run the example application using:
+The Python script is located in the `app/` folder, the Docker image is defined in the `Dockerfile`. Have a look at both files to understand what is happening.
```
prp: I wonder if we shouldn't change the name of 'app'. Isn't it unusual to put a Python script (or any application) there?

samples/languages/python/README.md
```diff
@@ -1,20 +1,43 @@
 SGX-LKL-OE Python and NumPy Sample Application
 ==============================================
 
-1. First build the Python and NumPy file system image:
```
prp, line 4: Personally, I like if there are clear numbered steps to follow, as this was it's harder to get lost.

samples/languages/python/README.md
prp: I would put it in the main README.md file, as this is fundamental to running applications with SGX-LKL. This should also include a more detailed explanation of what the host and app configs are and why both are needed.

samples/languages/python/build.sh
prp: Yeah, not having these files would be cleaner. If the CI needs them, they shouldn't pollute the sample directory. I think OE uses cmake files for the CI testing of the samples, right?

prp: Yes, the samples should assume that SGX-LKL has been correctly installed and configured on the host.

prp (commented):
Looks good overall. I left some minor discussion points, thanks.

samples/languages/python/README.md
letmaik: I don't have a strong opinion on that. src/ ?

samples/languages/python/README.md
letmaik: OK, tracking in https://github.com/lsds/sgx-lkl/issues/351.

samples/languages/python/build.sh
letmaik: I removed them. We can think about it again once CI for samples is added.

samples/languages/python/README.md
letmaik: Changed it to the latter.

samples/languages/python/README.md
letmaik: Renamed to src/.

samples/languages/python/README.md
letmaik: Done.

samples/languages/python/README.md
prp: Hmm, if you were to install a (Python) application system wide, where would you put it? Perhaps `/opt/app`? (`/src` suggests source code only.)

letmaik: Maybe? But in Docker it's quite common to just dump things in some folder in /. In this case here it's just source code, not an actual application that's installed, so `/src` seemed fine.

Issue #336, opened 2020-05-28
If sgx-lkl-run-oe process hangs after a timedout test test run script exits


hukoyu:
When a core test times out after 5 minutes and sgx-lkl-oe is still hanging around, pipeline test run script doesn't continue and rest of the tests are not run:

AzDevOps  75716  197  0.0      0     0 ?        RN   00:09   9:39 [sgx-lkl-run-oe]
AzDevOps  76186  0.0  0.0  14856  1060 ?        SN   00:14   0:00 grep sgx-lkl-run-oe

https://github.com/lsds/sgx-lkl/blob/oe_port/.azure-pipelines/scripts/test_runner.sh#L27
https://github.com/lsds/sgx-lkl/blob/oe_port/.azure-pipelines/scripts/test_utils.sh#L26

Instead of exiting and not running remaining tests, we should kill hanging sgx-lkl-run-oe process and let the test run script continue to run remaining tests.

hukoyu:
Here is a link to a sample pipeline run job with this issue: 
https://dev.azure.com/sgx-lkl/sgx-lkl/_build/results?buildId=491&view=logs&j=fe3b6e0d-c167-5b41-87de-7abc79e7e1d7&t=e462892c-9357-502e-648c-9a324ddcd30e&l=306

https://tfsproduks1.visualstudio.com/Ab5352d9a-5ed5-48f2-b767-7f273b55b53c/sgx-lkl/_TestManagement/Runs?runId=7028&_a=resultQuery

letmaik:
`timeout` first sends SIGTERM and if that doesn't work SIGKILL. I'm not sure what else you could do.

hukoyu:
Fixed with PR https://github.com/lsds/sgx-lkl/pull/337

PR #337, opened 2020-05-28
Kill hanging sgx-lkl-run-oe process to continue to test run

hukoyu:
Fixing test issue : https://github.com/lsds/sgx-lkl/issues/336
When a test times out, it is possible that sgx-lkl-run-oe process is still running/hanging

Existing test harness logic was exiting if sgx-lkl-run-oe process exists and rest of the tests were not running. The failed/timed out test already reported. There is no point exiting test harness.

We can kill hanging process and output logs and continue running remaining tests. 

cc @letmaik  @jxyang  @SeanTAllen

### DIFF
--- a/.azure-pipelines/scripts/test_utils.sh
+++ b/.azure-pipelines/scripts/test_utils.sh
@@ -25,9 +25,19 @@ function ChangeDirectory()
 
 function CheckNotRunning()
 {
-    if pgrep -x sgx-lkl-run-oe >/dev/null; then
+    process="sgx-lkl-run-oe"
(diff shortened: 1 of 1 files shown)


letmaik:
I'm fine with merging it, but it doubt it will help since all tests are already run with `timeout` which first sends SIGTERM and if that doesn't work SIGKILL. As it's hard to reproduce these situations I would go ahead with merging and see if we ever observe the "Trying to kill hanging" message in the logs for builds with at least one timed out test.

letmaik (approved):

PR #340, opened 2020-05-29
Use oe_malloc for the mpmcq.

davidchisnall:
This is part of lthreads and sits below libc, so should not depend on
malloc.  Soon it will be initialised before libc starts.


prp (approved):
LGTM

SeanTAllen (approved):

Issue #230, opened 2020-05-11
Cannot mount data disk if mount point folder is missing in read-only root disk


letmaik:
Using read-only disks is useful and also a requirement for dm-verity. SGX-LKL creates mount point folders for additional disks if the root disk is writable, but fails if the root disk is readonly and the folder is missing.

When using Docker to create root disk images this is surprising as it prevents simple lift-and-shifting. Instead, the Dockerfile has to be modified to include `RUN mkdir /data` or similar. Docker gets around this issue by adding a writable overlay filesystem.

EDIT: #346 added support for overlayfs. **The only thing missing now is documentation.**

PR #346, opened 2020-05-29
Add overlayfs support for root disk

letmaik:
This PR adds overlayfs support for the root disk. The point is to have the root disk generally read-only to be able to use dm-verity for all application types without worrying that certain apps may want to write to folders that hadn't been specifically mounted writable (via tmpfs or extra disk). overlayfs essentially makes the root disk writable through an additional ephemeral writable layer. In this PR I currently use tmpfs for that, just because it was the easiest and also provides implicit encryption (since it's enclave memory). Persistent overlays are out of scope.

Notes on unmounting:
The overlay involves three mounts: the original disk `/mnt/vda`, the writable tmpfs layer `/mnt/oda-upper`, and the overlay `/mnt/oda` configured with lower layer = original disk, and upper layer = tmpfs. During start-up, pivoting via chroot makes `/mnt/oda` the new `/`. During shutdown, all extra disks and the root disk (`/`) are unmounted, however this would not unmount the original `/mnt/vda` and `/mnt/oda-upper` mounts. In this PR, the original disk must be read-only to have an overlay. Because of that, it's not an issue to not unmount the original disk (as it cannot become dirty).

It is worth nothing that SGX-LKL currently doesn't unmount any special mounts (like /proc etc.) before unmounting `/`. It may be that this happens automatically in LKL. It would be nice to unmount those mounts explicitly, including `/mnt/vda` (in pre-chroot) when using an overlay, however this would require exiting the chroot to gain access to the original environment. It may be possible to do that simply by unmounting `/` in the chroot and cd'ing to `/` again. This could be analysed by printing the contents of `/proc/mounts` before/after. For this PR this is out of scope and can be looked at separately in a general attempt to explicitly unmount *every* mount.

I opened https://github.com/lsds/sgx-lkl/issues/367 to track explicit unmounting of special mounts and the root disk when using an overlay.


tests/basic/overlay/Makefile
```diff
@@ -0,0 +1,40 @@
+include ../../common.mk
+
+PROG=/bin/touch /foo.txt
```
prp, line 7: It would be good to check that the overlayfs semantics is actually working.

tests/basic/overlay/Makefile
```diff
@@ -0,0 +1,40 @@
...
+SGXLKL_HW_PARAMS=--hw-debug
+SGXLKL_SW_PARAMS=--sw-debug
+
+SGXLKL_ROOTFS=sgx-lkl-rootfs.img
+
+.DELETE_ON_ERROR:
+.PHONY: all clean
+
+$(SGXLKL_ROOTFS):
+	${SGXLKL_DISK_TOOL} create --size=${IMAGE_SIZE} --docker=./Dockerfile ${SGXLKL_ROOTFS}
```
prp, line 25: Don't you want to be testing this with dm-verity turned on?

src/lkl/setup.c
```diff
@@ -755,9 +784,15 @@ static void lkl_mount_disk(
...
     int err = 0;
-    char mnt_point[] = {"/mnt/vda"};
-    char new_dev_str[] = {"/mnt/vda/dev/"};
+    const char mnt_point_original[] = {"/mnt/vda"};
+    const char mnt_point_overlay[] = {"/mnt/vda-overlay"};
+    const char mnt_point_overlay_upper[] = { "/mnt/vda-overlay-upper" };
+    const char overlay_upper_dir[] = { "/mnt/vda-overlay-upper/upper" };
+    const char overlay_work_dir[] = { "/mnt/vda-overlay-upper/work" };
+    const char* mnt_point = { use_overlay ? mnt_point_overlay : mnt_point_original };
+    const char* new_dev_str = { use_overlay ? "/mnt/vda-overlay/dev/" : "/mnt/vda/dev/" };
```
prp: Perhaps it would be cleaner to only do some of this if `use_overlay` is set.

prp (commented):
Pretty cool! Left some minor comments.

src/lkl/setup.c
```diff
@@ -755,9 +784,15 @@ static void lkl_mount_disk(
 
 static void lkl_mount_root_disk(struct enclave_disk_config* disk)
 {
+    bool use_overlay = false;
```
hukoyu: use_overlay is set to false and it is always false. It doesn't change. Is this intentional?

src/lkl/setup.c
```diff
@@ -785,7 +820,21 @@ static void lkl_mount_root_disk(struct enclave_disk_config* disk)
         }
     }
 
-    lkl_mount_disk(disk, 'a', mnt_point);
+    lkl_mount_disk(disk, 'a', mnt_point_original);
+
+    if (use_overlay)
```
hukoyu: Always false. Never becomes true.

src/lkl/setup.c
letmaik: See the PR description. I will add an app config option.

tests/basic/overlay/Makefile
letmaik: Well, the semantics is that now it's writable. If it's read-only then touch will fail which I checked.

tests/basic/overlay/Makefile
letmaik: I could, though there's no difference since SGXLKL_HD_RO=1 is set which achieves the same effect.

src/lkl/setup.c
letmaik: I refactored the code, it's now all inside the `if`.

tests/basic/overlay/Makefile
prp: Hmm, I'm not sure if this is indeed entirely equivalent. I think that it would be good to test this with an actual integrity protect device-mapper file system, especially given the upcoming changes to the encrypted file system handling...

letmaik: My issue is that I don't want to create a test that tests too many things at once. And adding dm-crypt + dm-verity seems like a lot if we can get away with just mounting the root disk read-only. I checked that the test fails if readonly=true and overlay=false.

prp: Fair enough. So we should still have another test that checks if overlayfs works with an integrity-protected file system.

letmaik:
I just realized that this will also solve https://github.com/lsds/sgx-lkl/issues/230.

tests/basic/overlay/Makefile
letmaik: I won't add a dm-verity test because that's still badly broken. But I can do dm-crypt + dm-integrity and set the disk as readonly. OK?

letmaik: I take that back... I can do dm-crypt. Getting "bad count while changing owner" left and right with dm-integrity.

letmaik:
@prp I added a second test and also in the first test mount a second disk where the root disk is readonly and the mount point folder is missing, which works with overlay.

Do you have any idea on what to do about shutdown? Because of the chroot I don't have access to the original disk mount anymore and can only unmount the overlay (= `/`). What's the best way to handle this?

prp:
Hmm. do you have evidence that not all file systems are unmounted as part of the shutdown sequence?

letmaik:
No, but since there is code to unmount the rootfs and extra disks I was assuming that you have to unmount at least all non-tmpfs mounts. I guess it's slightly more important for writable disks. I was wondering actually if it makes sense to allow readonly=false and overlay=true at the same time. For example, singularity doesn't allow that because it's considered a user error (why would you need an overlay if you have a writable disk?).

prp:
Hmm, `umount()` has the ability to recursively unmount everything. I wonder if this is already happening? I don't think that there is anything special about the `tmpfs` mounts.

We should open an issue to track this.

prp (approved):
LGTM

Issue #169, opened 2020-05-04
[Tests] system calls with NULL causes abort


ashokchandran:
in umount02 when it test umount with null, it causes enclave to abort. Normally application is expected to get EFAULT. But the back trace is showing that in lkl it calls oe_abort when address is invalid.
```
#0 lkl_access_ok (addr=0, size=1) at lkl/setup.c:1499
#1 0x00007fff0010aaff in __lkl_access_ok (size=, addr=) at ./arch/lkl/include/asm/uaccess.h:14
#2 strncpy_from_user (count=, src=, dst=) at ./include/asm-generic/uaccess.h:233
#3 getname_flags (filename=0x0, flags=0, empty=0x0) at fs/namei.c:149
#4 0x00007fff0010ad2d in getname (filename=) at fs/namei.c:209
#5 0x00007fff0010da1d in user_path_mountpoint_at (dfd=0, name=, flags=1049890816, path=0x7fff3e941020)
at fs/namei.c:2745
#6 0x00007fff0012003d in ksys_umount (name=0x7fff3e941000 " \020\224>\377\177", flags=0) at fs/namespace.c:1680
#7 0x00007fff00120479 in __do_sys_umount (flags=, name=) at fs/namespace.c:1706
#8 __se_sys_umount (name=, flags=) at fs/namespace.c:1704
#9 0x00007fff00088076 in run_syscall (params=, no=) at arch/lkl/kernel/syscalls.c:44
#10 lkl_syscall (no=140734243278880, params=0x0) at arch/lkl/kernel/syscalls.c:183
#11 0x00007fff0054901a in __filter_syscall2 (n=39, a1=0, a2=0) at ./src/internal/syscall.h:161
#12 0x00007fff0054924a in umount (special=0x0) at src/linux/mount.c:11```
if (!ret)
{
sgxlkl_fail("lkl_access_check failed: %p\n", (void*)addr);
}
return ret;

letmaik:
@ashokchandran Please use ` ``` ... ``` ` for stack traces (and anything else like that) in GitHub, otherwise you create links to old issues.

ashokchandran:
Thanks updated.

vamsikrishna935:
Enclave is getting exited when test is trying to access invalid address.
Same issue observed in fstat03 test case.

hukoyu:
Sub test cases of some LTP tests designed to test generation of EFAULT
by passing/accessing invalid address values (“-1”)  to some syscalls.
[...]

PR #348, opened 2020-05-29
LTP: Fix lkl_access_check failed error

shaikshavali1:
In ltp test case (fcntl13, chdir04, getpeername01, getrusage02,
preadv02, readlink03, rt_sigprocmask02, pwritev02),
some of the sub test cases designed to test generation of EFAULT
by passing/accessing invalid address values(ex:-1, 0, 1) to syscall.
Currently sgx behaviour is to call enclave abort, if address is
not within enclave address range. Because of this test program
is causing enclave abort and exiting with non-zero exit code.

issue169 is fixed.

### DIFF
--- a/tests/ltp/patches/fix_chdir_chdir04.patch
+++ b/tests/ltp/patches/fix_chdir_chdir04.patch
@@ -1,0 +1,44 @@
+In this ltp test case (testcases/kernel/syscalls/chdir/chdir04),
+one of the sub test case designed to test generation of EFAULT
+by passing/accessing invalid address values (“-1”) to “chdir” syscall.
+Currently sgx behaviour is to call enclave abort, if address is
+not within enclave address range. Because of this test program
+is causing enclave abort and exiting with non-zero exit code. 
+
+Github issue169 (https://github.com/lsds/sgx-lkl/issues/169)
+is raised to fix this behaviour. The sub test cases which test
+EFAULT error behaviour is commented/disabled until github
+issue169 is fixed.
+
+diff --git a/testcases/kernel/syscalls/chdir/chdir04.c b/testcases/kernel/syscalls/chdir/chdir04.c
+index f0420e4c9..89f460a97 100644
+--- a/testcases/kernel/syscalls/chdir/chdir04.c
++++ b/testcases/kernel/syscalls/chdir/chdir04.c
+@@ -60,6 +60,10 @@
+ #include <sys/mman.h>
+ #include "test.h"
+ 
++//TODO: Remove below line after fixing github issue 169.
++//url: https://github.com/lsds/sgx-lkl/issues/169
++#define FIXED_GITHUB_ISSUE_169 0
++
+ char *TCID = "chdir04";
+ 
+ char bad_dir[] =
+@@ -83,12 +87,16 @@ struct test_case_t {
+ 	     */
+ 	{
+ 	noexist_dir, ENOENT},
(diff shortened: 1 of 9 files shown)


tests/ltp/patches/fix_chdir_chdir04.patch
```diff
@@ -0,0 +1,44 @@
...
+diff --git a/testcases/kernel/syscalls/chdir/chdir04.c b/testcases/kernel/syscalls/chdir/chdir04.c
+index f0420e4c9..89f460a97 100644
+--- a/testcases/kernel/syscalls/chdir/chdir04.c
++++ b/testcases/kernel/syscalls/chdir/chdir04.c
+@@ -60,6 +60,10 @@
+ #include <sys/mman.h>
+ #include "test.h"
+ 
++//TODO: Remove below line after fixing github issue 169.
++//url: https://github.com/lsds/sgx-lkl/issues/169
```
hukoyu: Remove. Already mentioned in patch description

tests/ltp/patches/fix_chdir_chdir04.patch
```diff
@@ -0,0 +1,44 @@
...
++#define FIXED_GITHUB_ISSUE_169 0
```
hukoyu: Please get rid of FIXED_GITHUB_ISSUE_169 approach. It is overcomplicated. Please just comment out affected subtests and add comment like // TODO: Enable back after issue 169 fixed

Please do this for the whole patches, I don't want to give same feedback for all of them

tests/ltp/patches/fix_chdir_chdir04.patch
```diff
@@ -0,0 +1,44 @@
...
++
+ char *TCID = "chdir04";
+ 
+ char bad_dir[] =
+@@ -83,12 +87,16 @@ struct test_case_t {
+ 	     */
+ 	{
+ 	noexist_dir, ENOENT},
++//TODO: Remove '#if FIXED_GITHUB_ISSUE_169' & '#endif'  directive after fixing github issue 169.
++//url: https://github.com/lsds/sgx-lkl/issues/169
```
hukoyu: Remove. No need to mention url over and over again. You already mentioned it in patch description.

hukoyu (changes requested):
@shaikshavali1 I see you put lots of effort but it is too complicated and too many comments. Please keep it simple.
Please get rid of FIXED_GITHUB_ISSUE_169 approach. It is overcomplicated. 
Please just comment out affected subtests and add comment like 
// TODO: Enable back after issue 169 fixed

Please do this for the whole patches, I don't want to give same feedback for all of them
Also don't need to specify issue 169 url over and over again in comments. You already mentioned it in patch description.

ashokchandran:
No other observations other than those already stated by huseyin.

tests/ltp/patches/fix_chdir_chdir04.patch
shaikshavali1: Done

tests/ltp/patches/fix_chdir_chdir04.patch
shaikshavali1: Done

tests/ltp/patches/fix_chdir_chdir04.patch
shaikshavali1: Done

hukoyu (approved):

PR #354, opened 2020-06-01
Make futex_dequeue removal conditional

SeanTAllen:
Prior to this commit, futex_dequeue assumed that the given lthread had an
entry in futex_queues. If that wasn't the case, a segfault would occur during
removal.

Additionally, there's bookkeeping that needs to be done when removing an item
from futex_queues that wasn't being done.

### DIFF
--- a/src/sched/futex.c
+++ b/src/sched/futex.c
@@ -62,7 +62,7 @@ void futex_dequeue(struct lthread *lt)
             fq->futex_lt = NULL;
             a_fetch_add(&futex_sleepers, -1);
             SLIST_REMOVE(&futex_queues, fq, futex_q, entries);
-            lt->err = FUTEX_NONE;
+            break;
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/sched/futex.c
```diff
@@ -49,11 +49,22 @@ static uint32_t to_futex_key(int* uaddr)
...
 
-    SLIST_REMOVE(&futex_queues, &lt->fq, futex_q, entries);
+    SLIST_FOREACH_SAFE(fq, &futex_queues, entries, tmp)
+    {
+        if (lt == fq->futex_lt)
+        {
+            fq->futex_lt = NULL;
+            a_fetch_add(&futex_sleepers, -1);
+            SLIST_REMOVE(&futex_queues, fq, futex_q, entries);
+            lt->err = FUTEX_NONE;
```
davidchisnall:
```suggestion
            break;
```

We don't need to keep traversing the list once we've found the lthread that we're looking for (it only has one `next` pointer, so can't be in the list twice).  We also don't need to clear `errno` here - this is not a libc call.

davidchisnall (approved):

PR #358, opened 2020-06-01
Ltp: fixed chmod03&04 and utime02&05

vamsikrishna935:
Tests were modified to create temporary directories using root
permissions.


hukoyu (approved):
LGTM

PR #360, opened 2020-06-01
Add asserts for yielding while holding ticket locks.

SeanTAllen:
Ticket locks are spinlocks, which do not yield.  Any code that yields
while holding a ticket lock can be a source of deadlocks.

### DIFF
--- a/src/sched/lthread.c
+++ b/src/sched/lthread.c
@@ -343,6 +343,8 @@ void _lthread_yield_cb(struct lthread* lt, void (*f)(void*), void* arg)
 void _lthread_yield(struct lthread* lt)
 {
     struct lthread_sched* sched = lthread_get_sched();
+    // Yielding while holding a ticket lock may cause deadlock
+    SGXLKL_ASSERT(lt->ticketlocks_held == 0);
     _switch(&sched->ctx, &lt->ctx);
 }
 
@@ -492,8 +494,6 @@ int _lthread_resume(struct lthread* lt)
             sched->current_lthread->tp = fs_ptr;
         }
     }
-    sched->current_lthread = NULL;
-    reset_tls_tp(lt);
 
     if (lt->attr.state & BIT(LT_ST_EXITED))
     {
@@ -517,7 +517,12 @@ int _lthread_resume(struct lthread* lt)
     if (lt->yield_cb)
     {
         lt->yield_cb(lt->yield_cbarg);
+        // Yielding while holding a ticket lock may cause deadlock
+        SGXLKL_ASSERT(lt->ticketlocks_held == 0);
     }
+
+    sched->current_lthread = NULL;
+    reset_tls_tp(lt);
 
     return (0);
 }
--- a/src/include/enclave/lthread.h
+++ b/src/include/enclave/lthread.h
@@ -155,6 +155,9 @@ struct lthread
     void (*yield_cb)(void*);
     void* yield_cbarg;
     struct futex_q fq;
+#ifndef NDEBUG
+    int ticketlocks_held;
+#endif
 };
 
 struct lthread_queue
--- a/src/include/enclave/ticketlock.h
+++ b/src/include/enclave/ticketlock.h
@@ -29,6 +29,11 @@ static inline void ticket_lock(struct ticketlock* t)
         a_spin();
 #if DEBUG
     t->lt = NULL;
+    struct lthread *lt = lthread_self();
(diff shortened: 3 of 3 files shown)


SeanTAllen:
This is your code David. Even though it hasn't found anything yet, I think it is reasonable to merge these changes as they would help prevent a hard to track down regression in the future.

SeanTAllen:
I've never seen this before:

```
2020-06-01T19:20:43.8737882Z -----------stderr-start-------------
2020-06-01T19:20:43.8738616Z [   SGX-LKL  ] SGX-LKL (OE) Git version cc0764d-dirty LKL version 5.3.0 [DEBUG build (-O0)] [SOFTWARE DEBUG]
2020-06-01T19:20:43.8739508Z [   SGX-LKL  ] nproc=8 ETHREADS=8 CMDLINE="mem=32M"
2020-06-01T19:20:43.8740143Z [   SGX-LKL  ] HW TLS support: conf->fsgsbase=0
2020-06-01T19:20:43.8741060Z [   SGX-LKL  ] Registering disk 0 (path='sgx-lkl-rootfs.img', mnt='/', [RW   ])
2020-06-01T19:20:43.8742104Z [   SGX-LKL  ] No tap device specified, networking will not be available.
2020-06-01T19:20:43.8743367Z [   SGX-LKL  ] get_signed_libsgxlkl_path... result=/opt/sgx-lkl-debug/bin/../lib/libsgxlkl.so.signed
2020-06-01T19:20:43.8744489Z [   SGX-LKL  ] oe_create_enclave... result=0 (OE_OK)
2020-06-01T19:20:43.8745457Z [   SGX-LKL  ] sgxlkl_enclave_init(ethread_id=0)
2020-06-01T19:20:43.8746520Z [[  SGX-LKL ]] wg0 has public key 8bfpz0XjrxwxVfIkrGSTAKvGXyN+uZ7VHS1LhSF9Fmk=
2020-06-01T19:20:43.8747199Z free(): invalid pointer
2020-06-01T19:20:43.8747701Z free(): invalid pointer
2020-06-01T19:20:43.8748226Z free(): invalid pointer
2020-06-01T19:20:43.8748701Z free(): invalid pointer
2020-06-01T19:20:43.8749280Z free(): invalid pointer
2020-06-01T19:20:43.8750468Z [   SGX-LKL  ] ethread (4: 0) free(): invalid pointer
2020-06-01T19:20:43.8750867Z Aborted (core dumped)
2020-06-01T19:20:43.8751378Z make: *** [run-sw] Error 134
```

src/sched/lthread.c
```diff
@@ -567,8 +567,13 @@ int _lthread_resume(struct lthread* lt)
     if (lt->yield_cb)
     {
         lt->yield_cb(lt->yield_cbarg);
+        // Yielding while holding a ticket lock may cause deadlock
+        SGXLKL_ASSERT(lt->ticketlocks_held == 0);
     }
 
+    sched->current_lthread = NULL;
```
SeanTAllen, line 524: did you intend to move these two lines?

davidchisnall: Yes, without that any code run in the callback is not able to get the current thread. That makes it impossible for the unlock to decrement the counter.

I did not send this as a PR because I’m uncertain if that has any other negative side effects. I don’t believe it should be possible for the lthread strict to be deallocated before here, but I’m not completely sure. There may be a race condition in the cloned thread deallocation if there are multiple ethreads. That said, the resume function expects the lthread object to be still live because it accesses a field, so this shouldn’t make things worse...

SeanTAllen: Bad things definitely happen with this here. See:

https://dev.azure.com/sgx-lkl/sgx-lkl/_build/results?buildId=1549&view=logs&j=4dfb742b-8e24-5ebe-1c3b-99535daead16

for some segfaults in addition to the assertions being triggered.

SeanTAllen:
@davidchisnall closing due to issues. see last note. i don't want to delete the branch as it was yours. Is it ok to delete?

davidchisnall:
I recently fixed an issue in the VirtIO code that was sleeping with a ticket lock held. Reopening this to see if we’re closer to being able to enable it.

davidchisnall:
It looks as if the code in `vio_enclave_wakeup_event_channel` holds a lock across things that can yield (specifically, waking up another thread.  On closer inspection, that code looks as if it can lose wakeups and needs some careful review.

PR #362, opened 2020-06-02
Override sysinfo syscall handling

jxyang:


davidchisnall:
Looks fine, please can you also remove the musl syscall bypass and bump the sub module?

davidchisnall:
Sorry, I didn’t see that you’d already sent the musl PR. Please bump the sub module so that CO can run with the new libc.

src/enclave/enclave_mem.c
```diff
@@ -17,6 +16,8 @@
 #include "enclave/ticketlock.h"
 #include "enclave/sgxlkl_config.h"
 
+static struct sysinfo;
```
davidchisnall: This looks like a type name without a variable name.  What does this actually do?

src/include/enclave/enclave_mem.h
```diff
@@ -29,7 +29,7 @@ void* enclave_mremap(
 
 int enclave_mmap_flags_supported(int flags, int fd);
 
-int syscall_SYS_sysinfo(struct sysinfo* info);
+long syscall_SYS_sysinfo(struct sysinfo* info);
```
davidchisnall: Please can you pull the overridden syscalls' prototypes into a separate header and move the implementation into the `src/lkl/` directory?  These are LKL-specific, not generic support code.

src/lkl/syscall-overrides.c
```diff
@@ -1,6 +1,9 @@
+#include <sys/syscall.h>
 #include "lkl/syscall-overrides-futex.h"
 #include "lkl/posix-host.h"
+#include "enclave/enclave_mem.h"
 
 void register_lkl_syscall_overrides()
 {
+    lkl_replace_syscall(SYS_sysinfo, syscall_SYS_sysinfo);
```
davidchisnall: I think this is the wrong syscall number.  LKL has different syscall numbers to x86-64 Linux.  This needs to be `__lkl__NR_sysinfo`.  See the changes for the futex override in #162 for how to structure this.

davidchisnall (changes requested):

src/enclave/enclave_mem.c
jxyang: This just declares to the compiler `struct sysinfo` is a type, so it doesn't complain anything like `struct sysinfo*`.

src/lkl/syscall-overrides.c
jxyang: `SYS_sysinfo` has the same value as `__NR_sysinfo`. I couldn't find anywhere we define `__lkl__NR_sysinfo` though.

jxyang:
@davidchisnall Is there a way to bump the submodule before the sgx-lkl-musl change is merged?

src/lkl/syscall-overrides.c
jxyang: In any case, I will update to `__lkl__NR_sysinfo` as the debugger shows it also has the same value.

src/include/enclave/enclave_mem.h
jxyang: The overriding function uses static variables from `enclave_mem.c`. What's the recommended way to share variables between `src/lkl` and `src/enclave`?

src/enclave/enclave_mem.c
davidchisnall: Please remove the `static` here (it doesn't mean anything and I suspect it's not actually valid C).  The entire line is unnecessary though, because it's a forward declaration that comes after the real declaration.  Fields of this struct are used after this line with no other `sysinfo` declaration after this point.  This line should be removed.

Actually, after your most recent change, `sysinfo` isn't used in this file, so the forward declaration isn't needed.

src/include/lkl/syscall-overrides-sysinfo.h
```diff
@@ -0,0 +1,6 @@
+#ifndef _LKL_SYSCALL_OVERRIDES_SYSINFO_H
+#define _LKL_SYSCALL_OVERRIDES_SYSINFO_H
+
+long syscall_sysinfo_override(struct sysinfo* info);
```
davidchisnall, line 9: Please add a doc comment.

src/include/enclave/enclave_mem.h
```diff
@@ -29,7 +29,7 @@ void* enclave_mremap(
 
 int enclave_mmap_flags_supported(int flags, int fd);
 
-int syscall_SYS_sysinfo(struct sysinfo* info);
+void enclave_mem_info(size_t* total, size_t* free);
```
davidchisnall, line 37: Please add a doc comment.

src/lkl/syscall-overrides-sysinfo.c
```diff
@@ -0,0 +1,17 @@
...
+
+long syscall_sysinfo_override(struct sysinfo* info)
+{
+    size_t total, free;
+    enclave_mem_info(&total, &free);
+    info->totalram = total;
+    info->freeram = free;
+    info->totalswap = 0;
+    info->freeswap = 0;
+    info->procs = 1;
```
davidchisnall: Please add a comment that this should report the number of ethreads.

davidchisnall (changes requested):

src/include/lkl/syscall-overrides-sysinfo.h
jxyang: Done

jxyang:
@davidchisnall All comments are addressed. Please review again.

davidchisnall (approved):

davidchisnall:
@jxyang, please rebase this so that it can be merged?

davidchisnall:
I have done the rebase and cleanup, this is now ready to merge once CI is happy.

PR #372, opened 2020-06-03
Fix incorrect one shot timer handling

### LINKED ISSUE #334: FAIL: Bug: next_delay_ns already set for timer with SGXLKL_ETHREADS=1
With SGXLKL_ETHREADS=1 below 4 tests fails in nightly sgx-lkl pipeline:

[[ SGX-LKL ]] FAIL: Bug: next_delay_ns already set for timer
[ SGX-LKL ] init (0: 19 exit=0) [ SGX-LKL ] FAIL: sgxlkl_ethread_init() failed (ethread_id=0 result=19 (OE_ENCLAVE_ABORTING))

**These tests fail with same failure**:
basic-sleep-(debug)-(run-hw)-(1-ethreads)
basic-nonroot_halt_test-(debug)-(run-hw)-(1-ethreads)
java-thread-(debug)-(run-hw)-(1-ethreads) 
dotnet-(debug)-(run-hw)-(1-ethreads)

SeanTAllen:
It should be possible to reset a one shot timer. Why this invariant
existed isn't something that either David Chisnall nor myself were
able to determine.

It is possible that at some point in the past, this invariant made sense,
however, given the current state of sgx-lkl, it doesn't.

Closes #334

### DIFF
--- a/src/lkl/posix-host.c
+++ b/src/lkl/posix-host.c
@@ -571,9 +571,15 @@ static int timer_set_oneshot(void* _timer, unsigned long ns)
 
         if (timer->next_delay_ns)
         {
-            sgxlkl_fail("Bug: next_delay_ns already set for timer\n");
+            if (ns < timer->next_delay_ns)
+            {
+                timer->next_delay_ns = ns;
+            }
         }
-        timer->next_delay_ns = ns;
+        else
+        {
+            timer->next_delay_ns = ns;
+        }
     }
     else
     {


prp (approved):
LGTM. So it seems that the code didn't expect `timer_set_oneshot` to be called multiple times within a callback.

SeanTAllen:
@prp yes, but there is no reason why it wouldn't be now. There might have been something in the past though.

PR #377, opened 2020-06-03
Add -Wall -Werror for enclave code & fix issues

letmaik:
-Wall -Werror was only used for host-side code.


SeanTAllen:
I love this PR!

src/host_interface/virtio_netdev.c
```diff
@@ -227,14 +227,15 @@ static void virtio_net_fd_net_poll_hup(uint8_t netdev_id)
...
-static void virtio_net_fd_net_free(uint8_t netdev_id)
-{
-    struct netdev_fd* nd_fd = get_netdev_fd_instance(netdev_id);
-    close(nd_fd->fd);
-}
+//static void virtio_net_fd_net_free(uint8_t netdev_id)
+//{
+//    struct netdev_fd* nd_fd = get_netdev_fd_instance(netdev_id);
+//    close(nd_fd->fd);
+//}
```
prp, line 238: Perhaps it's better to leave this in and keep the warning? This needs to be fixed.

src/include/enclave/lthread.h
```diff
@@ -278,12 +278,12 @@ extern "C"
      */
     int lthread_setspecific_remote(struct lthread* lt, long key, const void* value);
 
-    static void* lthread_getspecific(long key)
+    static inline void* lthread_getspecific(long key)
```
prp, line 281: Shouldn't the compiler figure this out by itself?

src/main-oe/sgxlkl_run_oe.c
```diff
@@ -1823,7 +1759,7 @@ int main(int argc, char* argv[], char* envp[])
     {
         get_signed_libsgxlkl_path(libsgxlkl, PATH_MAX);
     }
-    sgxlkl_host_verbose_raw("result=%s\n", libsgxlkl ? libsgxlkl : "null");
+    sgxlkl_host_verbose_raw("result=%s\n", libsgxlkl);
```
prp, line 1771: Shouldn't we print "not found" if it couldn't locate the enclave library?

prp (approved):
Good stuff. @davidchisnall should also review.

src/host_interface/virtio_netdev.c
letmaik: I don't want to generally allow unused code, so either we comment it out, remove it, or we temporarily disable the warning using gcc pragmas. Or we actually call the function...

src/include/enclave/lthread.h
letmaik: Apparently not, it spits out a warning and it seems like static inline is the right way to define functions in headers. I'm not an expert on this though.

src/main-oe/sgxlkl_run_oe.c
letmaik: `get_signed_libsgxlkl_path` will already exit and spit out an error.

src/host_interface/virtio_netdev.c
SeanTAllen: commenting out seems like a bad idea. its noise.

delete or call it would be my "vote".

src/include/lkl/syscall-overrides-futex.h
```diff
@@ -1,6 +1,8 @@
 #ifndef _LKL_SYSCALL_OVERRIDES_FUTEX_H
 #define _LKL_SYSCALL_OVERRIDES_FUTEX_H
 
+#include <time.h>
```
davidchisnall: I think this is the wrong include, we need the LKL types, not any system ones. They may be the same, but only by coincidence.

src/lkl/disk_mem.c
```diff
@@ -22,10 +22,8 @@ static struct enclave_disk_config* get_disk_config(int fd)
 
 static int fd_get_capacity(struct lkl_disk disk, unsigned long long* res)
 {
-    off_t off;
-
     struct enclave_disk_config* disk_config;
-    if (disk_config = get_disk_config(disk.fd))
+    if ( (disk_config = get_disk_config(disk.fd)) != NULL)
```
davidchisnall: You should need only one of these changes (double brackets or explicit compare).

src/lkl/posix-host.c
```diff
@@ -535,6 +533,7 @@ static void* timer_callback(void* _timer)
     mutex_unlock(&timer->mtx);
 
     lthread_exit(NULL);
+    return NULL;
```
davidchisnall: This adds dead code. The correct fix for this is to add a no return attribute on the lthread exit prototype.

src/lkl/setup.c
```diff
@@ -1232,6 +1227,7 @@ static int lkl_termination_thread(void* args)
     lthread_exit(NULL);
 
     SGXLKL_VERBOSE("done\n");
```
davidchisnall: There should be a warning that this debug like is unreachable if the exit call has the correct annotation.

src/include/lkl/syscall-overrides-futex.h
letmaik: Ideally yes, but other pieces pull in musl headers as well and if I make a change here then we'll get redefinition errors. Concretely, musl's pthread.h (included by enclave_oe.c) includes bits/alltypes.h which defines timespec. I'll leave it as it is, I'm sure this won't be an issue anymore after relayering.

src/lkl/posix-host.c
letmaik: Done, I had to add noreturn to `_lthread_yield` and `_switch` (assembly code) as well.

src/lkl/setup.c
letmaik: Interestingly this didn't happen. I removed the debug line now.

src/host_interface/virtio_netdev.c
letmaik: I'll leave it commented out for now. I tried to understand if it's safe to call it from `net_dev_remove` but this is called from the enclave and I'm not sure if host threads would still need it. If anyone wants to fix it, please send a follow-up PR :)

src/include/lkl/syscall-overrides-futex.h
davidchisnall: Really?  This file should be included in only two places and neither of them should include musl.

That said, this file is actually dead code, so a simpler fix is to delete it and the `#includes` of it...

src/lkl/posix-host.c
davidchisnall: Thanks.  Can you now remove the `return NULL;' (it should be raising a warning about unreachable code)?

letmaik: I did that already. GitHub is showing you the original change without an option to see the current state in this thread...

src/include/lkl/syscall-overrides-futex.h
letmaik: This requires a change in sgx-lkl-musl (src/internal/syscall.h) which is currently at 393a467 but the futex syscall bypass removal is a few commits later. Very confusing... Should I create a PR against sgx-lkl-musl's oe_port and then bump the submodule to the newest commit?

davidchisnall: Yes please.  I thought we'd bumped the musl version when we merged the change.

letmaik: https://github.com/lsds/sgx-lkl-musl/pull/19. I bumped sgx-lkl-musl to the PR commit. If CI passes here, we can merge the sgx-lkl-musl PR and then bump to the merged commit in oe_port (which may or may not be the same).

letmaik: This is failing left and right... I have a feeling it's due to the sgx-lkl-musl bump. I'll open a PR with just the bump and get that merged first (or not, if it exposes the same issues).

letmaik: I opened https://github.com/lsds/sgx-lkl/pull/410 but I realized that this is already being done in another PR. Probably that's the one you were thinking of, it's not merged yet. I will have to wait for that first.

jxyang: We should undo the [fstat byassing in sgx-lkl-musl](https://github.com/lsds/sgx-lkl-musl/pull/16). We merged the PR for sgx-lkl-musl, but not the corresponding PR to sgx-lkl. So if you bump sgx-lkl-msul, the orphen change will be included.

letmaik: I scheduled another build for #381. Not sure what was/is wrong with CI. If this passes I would go ahead and merge #381.

PR #381, opened 2020-06-04
Handle fstat and newfstatat compatibility in lkl

### LINKED ISSUE #150: Remove stat syscall bypass
Currently, sgx-lkl-musl has a work-around for the different stat layouts between x86-64 and LKL.  We should move that compat layer into `src/lkl` and wrap the system call.  

Relevant musl code: https://github.com/lsds/sgx-lkl-musl/blob/22c91c211aaf4048a4f034084bb7fa202bd6071c/src/internal/syscall.h#L66

jxyang:
Fix #150

### DIFF
--- a/src/include/lkl/syscall-overrides-fstat.h
+++ b/src/include/lkl/syscall-overrides-fstat.h
@@ -1,15 +1,14 @@
 #ifndef _LKL_SYSCALL_OVERRIDES_FSTAT_H
 #define _LKL_SYSCALL_OVERRIDES_FSTAT_H
 
-typedef long (*syscall_handler_t)(long arg1, ...);
+typedef long (*syscall_fstat_handler)(int, void*);
+typedef long (*syscall_newfstatat_handler)(
+    int, const char*, void*, int);
 
-syscall_handler_t orig_fstat;
+// long syscall_fstat_override(int fd, struct stat* stat);
 
-syscall_handler_t orig_newfstatat;
-
-long syscall_fstat_override(int fd, struct stat* stat);
-
-long syscall_newfstatat_override(int dfd, const char *fn,
-      struct stat *statbuf, int flag);
+// long syscall_newfstatat_override(int dfd, const char *fn,
+      // struct stat *statbuf, int flag);
+void syscall_register_fstat_overrides();
 
 #endif
--- a/src/include/lkl/syscall-overrides-fstat.h
+++ b/src/include/lkl/syscall-overrides-fstat.h
@@ -1,0 +1,15 @@
+#ifndef _LKL_SYSCALL_OVERRIDES_FSTAT_H
+#define _LKL_SYSCALL_OVERRIDES_FSTAT_H
+
+typedef long (*syscall_handler_t)(long arg1, ...);
+
+syscall_handler_t orig_fstat;
+
+syscall_handler_t orig_newfstatat;
+
+long syscall_fstat_override(int fd, struct stat* stat);
+
+long syscall_newfstatat_override(int dfd, const char *fn,
+      struct stat *statbuf, int flag);
+
+#endif
--- a/src/include/lkl/syscall-overrides-fstat.h
+++ b/src/include/lkl/syscall-overrides-fstat.h
@@ -5,10 +5,11 @@ typedef long (*syscall_fstat_handler)(int, void*);
 typedef long (*syscall_newfstatat_handler)(
     int, const char*, void*, int);
 
-// long syscall_fstat_override(int fd, struct stat* stat);
-
-// long syscall_newfstatat_override(int dfd, const char *fn,
-      // struct stat *statbuf, int flag);
+/**
+ * Register override functions that maintain compatibility
+ * between LKL's handling of syscall fstat and newfstatat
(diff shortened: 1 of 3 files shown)


src/include/lkl/syscall-overrides-fstat.h
```diff
@@ -0,0 +1,15 @@
+#ifndef _LKL_SYSCALL_OVERRIDES_FSTAT_H
+#define _LKL_SYSCALL_OVERRIDES_FSTAT_H
+
+typedef long (*syscall_handler_t)(long arg1, ...);
+
+syscall_handler_t orig_fstat;
```
davidchisnall: Do these two need to be in the header?

src/include/lkl/syscall-overrides-fstat.h
```diff
@@ -0,0 +1,15 @@
...
+
+syscall_handler_t orig_newfstatat;
+
+long syscall_fstat_override(int fd, struct stat* stat);
```
davidchisnall: Doc comment please.

src/include/lkl/syscall-overrides-fstat.h
```diff
@@ -0,0 +1,15 @@
...
+
+long syscall_newfstatat_override(int dfd, const char *fn,
```
davidchisnall: Doc comment please.

src/lkl/syscall-overrides.c
```diff
@@ -1,6 +1,10 @@
-#include "lkl/syscall-overrides-futex.h"
+#include <sys/syscall.h>
+#include "lkl/syscall-overrides-fstat.h"
 #include "lkl/posix-host.h"
 
 void register_lkl_syscall_overrides()
 {
+    orig_fstat = lkl_replace_syscall(__lkl__NR_fstat, syscall_fstat_override);
```
davidchisnall: If you move this into a `sgxlkl_register_stat_overrides` function in `syscall-overrides-fstat.c` then the two original call functions can be static.

src/lkl/syscall-overrides-fstat.c
```diff
@@ -0,0 +1,56 @@
...
+}
+
+long syscall_newfstatat_override(int dfd, const char *fn, struct stat *stat, int flag) {
+    struct lkl_stat lkl_stat = {0};
+    long ret;
+
+    if (orig_newfstatat == NULL)
+        sgxlkl_fail("Error: newfstatat syscall handler not defined");
+
+    ret = orig_newfstatat(dfd, fn, &lkl_stat, flag);
```
davidchisnall, line 65: This function pointer needs to be cast to the correct type (it is UB to call a C function with a different prototype).

src/lkl/syscall-overrides-fstat.c
```diff
@@ -0,0 +1,56 @@
...
+}
+
+long syscall_fstat_override(int fd, struct stat* stat) {
+    struct lkl_stat lkl_stat = {0};
+    long ret;
+
+    if (orig_fstat == NULL)
+        sgxlkl_fail("Error: fstat syscall handler not defined");
+
+    ret = orig_fstat(fd, &lkl_stat);
```
davidchisnall, line 50: Same comment as below.  This can be fixed more cleanly by providing the correct types for the two function pointers and doing the cast on assignment.

davidchisnall (changes requested):

src/lkl/syscall-overrides.c
jxyang: Good point. Will do.

src/include/lkl/syscall-overrides-fstat.h
jxyang: Removed.

src/include/lkl/syscall-overrides-fstat.h
```diff
@@ -0,0 +1,14 @@
...
+#define _LKL_SYSCALL_OVERRIDES_FSTAT_H
+
+typedef long (*syscall_fstat_handler)(int, void*);
+typedef long (*syscall_newfstatat_handler)(
+    int, const char*, void*, int);
+
+// long syscall_fstat_override(int fd, struct stat* stat);
+
+// long syscall_newfstatat_override(int dfd, const char *fn,
+      // struct stat *statbuf, int flag);
```
davidchisnall: Please make these doc comments, starting /** and correctly indented.

jxyang: Oh, I totally overlooked the `doc` part. Will do.

davidchisnall (approved):

davidchisnall:
I've merged the musl changes, please update the submodule so that CI can run and then we can merge.

davidchisnall:
Can you check the CI logs?  It looks as if there are some crashes from this still.

jxyang:
The pipeline is in a bad state. I can't repro the failures locally. Those are failures probably not related to this PR. Still investigating.

letmaik:
@jxyang I fixed the pipeline issue in #411. Could you merge oe_port? This PR already looks fine for nearly all CI jobs, just want to make sure.

PR #385, opened 2020-06-04
Fix bug in sem_up/sem_down implementation

SeanTAllen:
Prior to this change, it is possible to lose wakeup events.
For example, 3 different things call sem_down resulting in a
sem count of 0.

On waking, only 1 would be awoken. The rest would continue to sleep.

This changes so that all are awoken and they can see which gets
the semaphore and then go back to sleep. Because all are awoken, the
other logic works.

### DIFF
--- a/src/lkl/posix-host.c
+++ b/src/lkl/posix-host.c
@@ -211,7 +211,7 @@ static void sem_up(struct lkl_sem* sem)
     // there may be waiters.  Wake one up.
     if (atomic_fetch_add(&sem->count, 1) == 0)
     {
-        futex_wake(&sem->count, 1);
+        futex_wake(&sem->count, INT_MAX);
     }
 }
 


SeanTAllen:
@davidchisnall I'm planning on following this up with another PR to add more documentation on how the sem_up/sem_down implementation works.

davidchisnall (approved):

SeanTAllen:
@davidchisnall @prp this fixes a bug but from looking at CI results, it doesn't fix all hangs and it kind of looks like from my running CI several times that it makes some issues with non-release, sw builds worse.

It would appear from my trying yesterday that non-release, sw can not pass with whatever gets exposed.

I'd like to consider turning off the non-release, sw tests as part of this PR and merge then and I can continue looking into hangs and with each hang fix hopefully turn tests back on if they become less "guaranteed to fail" which appears to be the case with non-release, sw.

The other option would be to accept that in fixing this bug, it is going to make more CI failure happen for a while as I find and fix others that were hidden before.

Thoughts?

SeanTAllen:
So eventually all the tests passed. I'm going to merge. Looks like I had "some bad luck" yesterday.

I think this is going to expose a number of bugs that were previously hidden.

PR #386, opened 2020-06-04
Add additional output to clone test

SeanTAllen:
In case this ends up hanging on someone, this additional output could
be used to provide a better idea of where things went wrong for someone
trying to get started with diagnosing.


davidchisnall (approved):

SeanTAllen:
Failure is unrelated. It's in untouched code.

PR #388, opened 2020-06-05
LTP: fix test case getcwd04 issue

shaikshavali1:
The original test case create parent and child process.
In child it infinitely rename a testfile and in parent it
checks current CWD(current working directory) with
initial CWD for 5 seconds. If initial CWD and current CWD
remains same for 5 seconds then test case is declared as passed.

This test case is infinitely looping in do_child() function
because “SAFE_FORK” always returns 0. The test case is
modified to use the pthreads instead of multiple processes
to do parallel jobs in the test.


hukoyu (approved):
LGTM

PR #391, opened 2020-06-05
LTP: fix testcase sendfile02 issue

shaikshavali1:
The original test case create a child process to test
Sendfile system call. But, sgx-lkl environment supports single
process. Hence, the test case is modified to use pthread instead
of a child process.


vamsikrishna935 (approved):
@hukoyu Looks good to me. Could you please review.

hukoyu (approved):
LGTM

PR #396, opened 2020-06-08
LTP: fix test cases recv01, recvmsg01 issues

shaikshavali1:
In original test cases(recv01, recvmsg01) the main process create a child process
to wirte/read to/from sockets infinitely. In sgx-lkl supports
single process environment. The test case is modified to use
a child pthread  instead of forking a child process.
One of the sub test case designed to test generation of EFAULT

issue169 is fixed.


hukoyu (approved):
LGTM

PR #397, opened 2020-06-08
LTP: Fix testcases sendfile05 & sendfile06 issue

shaikshavali1:
This is a combined patch for test cases sendfile05
and sendfile06.
In original test case, the main process create a child process
to read from socket. The sgx-lkl supports single process
environment. The test case is modified to use
a child pthread instead of forking a child process.


hukoyu (approved):
LGTM

PR #398, opened 2020-06-08
Use OE's explicit ocall opt-in feature

jhand2:
Switch to the OE configuration that allows OE applications to explicitly opt-in to which system EDL ecalls/ocalls it wishes to use. The previous behavior was that OE compiled in all system EDL by default.

### DIFF
--- a/src/sgxlkl.edl
+++ b/src/sgxlkl.edl
@@ -1,6 +1,24 @@
 // SGX-LKL-OE enclave/host interface
 
 enclave {
+    /*
+     * OE requires applications to import EDL used by the SDK to support some
+     * core functionality.
+     *
+     * - attestation.edl & sgx_attestation.edl are needed for OE attestation
+     *   APIs
+     * - logging.edl is needed to support OE diagnostic console logging
+     * - cpu.edl is not required by SGX-LKL, but OE is linked in such a way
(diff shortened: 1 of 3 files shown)


letmaik (approved):

letmaik:
This depends on #401.

src/sgxlkl.edl
```diff
@@ -1,6 +1,11 @@
 // SGX-LKL-OE enclave/host interface
 
 enclave {
+    from "openenclave/edl/attestation.edl" import *;
```
davidchisnall, line 17: Please can you add a comment explaining what we get from each of these?  It looks as if this is importing more than we need (for example, we have our own ocall for CPUID emulation, which we are planning on removing because it's insecure and replacing with a table of CPUID leaf nodes that are valid and including that in the attestation).

jhand2: cpu.edl and debug.edl are currently always required due to the way oecore is linked. Ultimately cpu.edl should be removed and debug.edl should only be included for debug builds. But someone has to do the work in OE. I'm also working on supporting preprocessor defines in EDL to allow files to be conditionally included.

jhand2: Added a comment

jhand2:
Can this be merged?

PR #401, opened 2020-06-08
Fix pipeline failure due to OE submodules

jxyang:


letmaik:
Just out of curiosity, which folder is OE pulling in as submodule?

letmaik:
I noticed that the OE pipeline step succeeds even if the OE build fails. Should the `install_openenclave.sh` script get a `set -e`?

letmaik (approved):

letmaik:
@jxyang After merging this change the oe_port CI build failed with

CMake Error at host/CMakeLists.txt:465 (message):
  Intel SGX EnclaveCommonAPI library not found, aborting!

Any idea what this is about?

EDIT: Nevermind, this was introduced by myself actually.

Issue #325, opened 2020-05-27
ELF ABI stacks


wintersteiger:
From  #115 via #103: We should support the full ELF stack layout for maximum compatiblity.

@davidchisnall says:

>We should probably mirror the layout mandated by the amd64 ELF psABI here the amd64 ELF psABI here (ideally, we'd have that structure on the stack on userspace process start, because that's what the ABI actually requires). The layout there is (from the lowest allocated stack address upwards):

- 64-bit argc
- `argv` as an array of 64-bit pointers.
- 64-bit 0
- Environment pointers as an array of 64-bit pointers.
- 64-bit 0
- ELF auxiliary argument vector (array of pairs of 64-bit values)
- 64-bit 0.
- Platform-specific things, including all of the things that the pointer arrays point to.

PR #403, opened 2020-06-09
Lthreads disentangled from musl

### LINKED ISSUE #156: Separate userspace TLS from lthreads
As part of the relayering work, we should ensure that nothing in `src/sched` depends on libc.  

- [ ] Extend OE to use %gs to store per-ethread state.
- [ ] Extend OE to provide a per-ethread pointer for library users to consume.
- [ ] Move the lthread scheduler to use that pointer.  Currently, it is stored inside TLS, which produces some interesting circular dependencies. The lthread scheduler is accessed via a [static function in the musl codebase](https://github.com/lsds/sgx-lkl-musl/blob/22c91c211aaf4048a4f034084bb7fa202bd6071c/src/internal/pthread_impl.h#L223), which calls [another musl function](https://github.com/lsds/sgx-lkl-musl/blob/22c91c211aaf4048a4f034084bb7fa202bd6071c/arch/x86_64/pthread_arch.h).
- [ ] Revert musl changes that rely on lthreads
 - [ ] Return TLS region ownership entirely to musl.
 - [ ] Once #155 is completed, revert to using `clone` to create threads

vtikoo:
### Summary
- Remove musl related threading code from lthreads implementation.
- Add auxiliary vector entries on the elf64 stack created during first ecall.

Fixes #156 

Corresponding sgx-lkl-musl PR: https://github.com/lsds/sgx-lkl-musl/pull/18

### DIFF
--- a/.azure-pipelines/other/disabled_tests.txt
+++ b/.azure-pipelines/other/disabled_tests.txt
@@ -1,2 +1,4 @@
 tests/virtio/python_read/Makefile
 tests/languages/java/network/Makefile
+tests/basic/pthread_detach/Makefile
+tests/languages/nodejs/Makefile
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,8 +7,8 @@
 	url = https://github.com/lsds/musl.git
 [submodule "sgx-lkl-musl"]
 	path = sgx-lkl-musl
-	url = https://github.com/lsds/sgx-lkl-musl.git
-	branch = oe_port
+	url = https://github.com/vtikoo/sgx-lkl-musl.git
+	branch = move-lthread-out
 [submodule "openenclave"]
 	path = openenclave
 	url = https://github.com/openenclave/openenclave.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,8 +7,8 @@
 	url = https://github.com/lsds/musl.git
 [submodule "sgx-lkl-musl"]
 	path = sgx-lkl-musl
-	url = https://github.com/vtikoo/sgx-lkl-musl.git
-	branch = move-lthread-out
+	url = https://github.com/lsds/sgx-lkl-musl.git
+	branch = oe_port
 [submodule "openenclave"]
 	path = openenclave
 	url = https://github.com/openenclave/openenclave.git
--- a/src/enclave/enclave_event_channel.c
+++ b/src/enclave/enclave_event_channel.c
@@ -5,6 +5,7 @@
 
 #include "enclave/enclave_util.h"
 #include "enclave/lthread.h"
+#include "enclave/lthread_int.h"
 #include "enclave/sgxlkl_t.h"
 #include "enclave/ticketlock.h"
 #include "enclave/vio_enclave_event_channel.h"
--- a/src/enclave/enclave_init.c
+++ b/src/enclave/enclave_init.c
@@ -138,10 +138,6 @@ int __libc_init_enclave(int argc, char** argv)
     SGXLKL_VERBOSE("calling enclave_mman_init()\n");
     enclave_mman_init(
         sgxlkl_heap_base, sgxlkl_heap_size / PAGESIZE, cfg->mmap_files);
-    // Add read only buffer page towards the end of enclave heap
-    // This is to be defensive against buffer overflows which read
-    // off the  end of heap. See Issue# 742 for more.
-    enclave_mmap(0, 4096, 0, PROT_READ, 1);
 
     libc.user_tls_enabled = sgxlkl_in_sw_debug_mode() ? 1 : cfg->fsgsbase;
 
--- a/src/enclave/enclave_init.c
+++ b/src/enclave/enclave_init.c
@@ -156,10 +156,9 @@ int __libc_init_enclave(int argc, char** argv)
     newmpmcq(&__scheduler_queue, max_lthreads, 0);
     
     init_ethread_tp();
-    
-    char** envp = argv + argc + 1;
-    __init_libc(envp, argv[0]);
 
+    __init_heap_from_libc();
+    
     size_t espins = cfg->espins;
     size_t esleep = cfg->esleep;
     lthread_sched_global_init(espins, esleep);
@@ -173,6 +172,5 @@ int __libc_init_enclave(int argc, char** argv)
     }
 
     lthread_run();
-
     return sgxlkl_enclave_state.exit_status;
 }
--- a/src/enclave/enclave_init.c
+++ b/src/enclave/enclave_init.c
@@ -138,6 +138,10 @@ int __libc_init_enclave(int argc, char** argv)
     SGXLKL_VERBOSE("calling enclave_mman_init()\n");
     enclave_mman_init(
         sgxlkl_heap_base, sgxlkl_heap_size / PAGESIZE, cfg->mmap_files);
+    // Add read only buffer page towards the end of enclave heap
+    // This is to be defensive against buffer overflows which read
+    // off the  end of heap. See Issue# 742 for more.
+    enclave_mmap(0, 4096, 0, PROT_READ, 1);
 
     libc.user_tls_enabled = sgxlkl_in_sw_debug_mode() ? 1 : cfg->fsgsbase;
 
@@ -158,15 +162,6 @@ int __libc_init_enclave(int argc, char** argv)
     newmpmcq(&__scheduler_queue, max_lthreads, 0);
     
     init_ethread_tp();
-
-    /* Temporary workaround!!
-     * Without the following line, overlay and overlay_encrypted
-     * tests cause LKL to panic.
-     * All the function does is a malloc() call from within libc. It seems like
-     * the sequence in which pages are allocated to libc malloc and rest of the
-     * users of enclave_mmap(lthread, LKL etc.) matters for these tests.
-    */
-    __init_heap_from_libc();
 
     size_t espins = cfg->espins;
     size_t esleep = cfg->esleep;
--- a/src/enclave/enclave_init.c
+++ b/src/enclave/enclave_init.c
@@ -10,8 +10,6 @@
 #include "enclave/enclave_util.h"
 #include "enclave/lthread.h"
 #include "enclave/lthread_int.h"
-#include "enclave/sgxlkl_app_config.h"
-#include "enclave/sgxlkl_config.h"
 #include "enclave/wireguard.h"
 #include "enclave/wireguard_util.h"
 #include "shared/env.h"
@@ -110,7 +108,7 @@ static int startmain(void* args)
     /* Set locale for usersapce components using it */
     pthread_t self = __pthread_self();
     self->locale = &libc.global_locale;
-    
+
     init_wireguard();
     find_and_mount_disks();
 
--- a/src/enclave/enclave_init.c
+++ b/src/enclave/enclave_init.c
@@ -9,6 +9,12 @@
 #include "enclave/enclave_oe.h"
 #include "enclave/enclave_util.h"
 #include "enclave/lthread.h"
+<<<<<<< HEAD
+=======
+#include "enclave/lthread_int.h"
+#include "enclave/sgxlkl_app_config.h"
+#include "enclave/sgxlkl_config.h"
+>>>>>>> move lthread init before libc init
 #include "enclave/wireguard.h"
 #include "enclave/wireguard_util.h"
 #include "shared/env.h"
@@ -150,9 +156,10 @@ int __libc_init_enclave(int argc, char** argv)
     max_lthreads = next_power_of_2(max_lthreads);
 
     newmpmcq(&__scheduler_queue, max_lthreads, 0);
-
+    
+    init_ethread_tls();
     __init_libc(envp, argv[0]);
-    __init_tls();
+    
 
     size_t espins = cfg->espins;
     size_t esleep = cfg->esleep;
--- a/src/enclave/enclave_init.c
+++ b/src/enclave/enclave_init.c
@@ -91,6 +91,8 @@ static void init_wireguard()
 
 static int startmain(void* args)
 {
+    __init_libc(sgxlkl_enclave_state.elf64_stack.envp,
+        sgxlkl_enclave_state.elf64_stack.argv[0]);
     __libc_start_init();
     a_barrier();
 
@@ -112,8 +114,6 @@ static int startmain(void* args)
     init_wireguard();
     find_and_mount_disks();
 
-    __init_libc(sgxlkl_enclave_state.elf64_stack.envp,
-        sgxlkl_enclave_state.elf64_stack.argv[0]);
     /* Launch stage 3 dynamic linker, passing in top of stack to overwrite.
      * The dynamic linker will then load the application proper; here goes! */
     __dls3(&sgxlkl_enclave_state.elf64_stack, __builtin_frame_address(0));
--- a/src/enclave/enclave_init.c
+++ b/src/enclave/enclave_init.c
@@ -9,12 +9,9 @@
 #include "enclave/enclave_oe.h"
 #include "enclave/enclave_util.h"
 #include "enclave/lthread.h"
-<<<<<<< HEAD
-=======
 #include "enclave/lthread_int.h"
 #include "enclave/sgxlkl_app_config.h"
 #include "enclave/sgxlkl_config.h"
->>>>>>> move lthread init before libc init
 #include "enclave/wireguard.h"
 #include "enclave/wireguard_util.h"
 #include "shared/env.h"
@@ -110,6 +107,10 @@ static int startmain(void* args)
     lkl_start_init();
     lthread_set_funcname(lthread_self(), "sgx-lkl-init");
 
+    /* Set locale for usersapce components using it */
+    pthread_t self = __pthread_self();
+    self->locale = &libc.global_locale;
+    
     init_wireguard();
     find_and_mount_disks();
 
--- a/src/enclave/enclave_init.c
+++ b/src/enclave/enclave_init.c
@@ -159,6 +159,13 @@ int __libc_init_enclave(int argc, char** argv)
     
     init_ethread_tp();
 
+    /* Temporary workaround!!
+     * Without the following line, overlay and overlay_encrypted
+     * tests cause LKL to panic.
+     * All the function does is a malloc() call from within libc. It seems like
+     * the sequence in which pages are allocated to libc malloc and rest of the
+     * users of enclave_mmap(lthread, LKL etc.) matters for these tests.
+    */
     __init_heap_from_libc();
 
     size_t espins = cfg->espins;
--- a/src/enclave/enclave_init.c
+++ b/src/enclave/enclave_init.c
@@ -157,7 +157,7 @@ int __libc_init_enclave(int argc, char** argv)
 
     newmpmcq(&__scheduler_queue, max_lthreads, 0);
     
-    init_ethread_tls();
+    init_ethread_tp();
     __init_libc(envp, argv[0]);
     
 
--- a/src/enclave/enclave_init.c
+++ b/src/enclave/enclave_init.c
@@ -112,6 +112,8 @@ static int startmain(void* args)
     init_wireguard();
     find_and_mount_disks();
 
+    __init_libc(sgxlkl_enclave_state.elf64_stack.envp,
+        sgxlkl_enclave_state.elf64_stack.argv[0]);
     /* Launch stage 3 dynamic linker, passing in top of stack to overwrite.
      * The dynamic linker will then load the application proper; here goes! */
     __dls3(&sgxlkl_enclave_state.elf64_stack, __builtin_frame_address(0));
@@ -158,7 +160,7 @@ int __libc_init_enclave(int argc, char** argv)
     init_ethread_tp();
 
     __init_heap_from_libc();
-    
+
     size_t espins = cfg->espins;
     size_t esleep = cfg->esleep;
     lthread_sched_global_init(espins, esleep);
--- a/src/enclave/enclave_init.c
+++ b/src/enclave/enclave_init.c
@@ -120,7 +120,6 @@ static int startmain(void* args)
 int __libc_init_enclave(int argc, char** argv)
 {
     struct lthread* lt;
-    char** envp = argv + argc + 1;
     const sgxlkl_enclave_config_t* cfg = sgxlkl_enclave_state.config;
 
     /* Upper heap memory area is allotted to OE and rest is used by SGXLKL */
@@ -157,8 +156,9 @@ int __libc_init_enclave(int argc, char** argv)
     newmpmcq(&__scheduler_queue, max_lthreads, 0);
     
     init_ethread_tp();
-    __init_libc(envp, argv[0]);
     
+    char** envp = argv + argc + 1;
+    __init_libc(envp, argv[0]);
 
     size_t espins = cfg->espins;
     size_t esleep = cfg->esleep;
--- a/src/enclave/enclave_oe.c
+++ b/src/enclave/enclave_oe.c
@@ -225,7 +225,7 @@ void sgxlkl_ethread_init(void)
     void* tls_page;
     __asm__ __volatile__("mov %%gs:0,%0" : "=r"(tls_page));
 
-    struct sched_tcb_base* sched_tcb = (struct sched_tcb_base*)tls_page;
+    struct lthread_tcb_base* sched_tcb = (struct lthread_tcb_base*)tls_page;
     sched_tcb->self = (void*)tls_page;
 
     size_t tls_offset = SCHEDCTX_OFFSET;
@@ -365,7 +365,7 @@ int sgxlkl_enclave_init(const sgxlkl_shared_memory_t* shared_memory)
     void* tls_page;
     __asm__ __volatile__("mov %%gs:0,%0" : "=r"(tls_page));
 
-    struct sched_tcb_base* sched_tcb = (struct sched_tcb_base*)tls_page;
+    struct lthread_tcb_base* sched_tcb = (struct lthread_tcb_base*)tls_page;
     sched_tcb->self = (void*)tls_page;
 
     size_t tls_offset = SCHEDCTX_OFFSET;
--- a/src/enclave/enclave_oe.c
+++ b/src/enclave/enclave_oe.c
@@ -238,7 +238,7 @@ void sgxlkl_ethread_init(void)
     }
 
     /* Initialization completed, now run the scheduler */
-    init_ethread_tls();
+    init_ethread_tp();
     _lthread_sched_init(sgxlkl_enclave_state.config->stacksize);
     lthread_run();
 
--- a/src/enclave/enclave_oe.c
+++ b/src/enclave/enclave_oe.c
@@ -18,9 +18,6 @@
 
 #define AUXV_ENTRIES 13
 
-void cleanup_sgxlkl_enclave_config();
-
-int ethreads_minus_1 = -1;
 char *at_platform = "x86_64";
 sgxlkl_enclave_state_t sgxlkl_enclave_state = {0};
 
@@ -276,14 +273,6 @@ static void _sgxlkl_enclave_show_attribute(const void* sgxlkl_enclave_base)
 }
 #endif
 
-static inline void dec_ethread_counter() {
-    if (a_fetch_add(&ethreads_minus_1, -1)==0) {
-		// last ethread to exit
-		ethreads_minus_1 = 0;
-		cleanup_sgxlkl_enclave_config();
-	}
-}
-
 void sgxlkl_ethread_init(void)
 {
     void* tls_page;
@@ -301,12 +290,10 @@ void sgxlkl_ethread_init(void)
         a_spin();
     }
 
-    a_inc(&ethreads_minus_1);
     /* Initialization completed, now run the scheduler */
     init_ethread_tp();
     _lthread_sched_init(sgxlkl_enclave_state.config->stacksize);
     lthread_run();
-    dec_ethread_counter();
 
     return;
 }
@@ -411,8 +398,6 @@ int sgxlkl_enclave_init(const sgxlkl_shared_memory_t* shared_memory)
     memset(&sgxlkl_enclave_state, 0, sizeof(sgxlkl_enclave_state));
     sgxlkl_enclave_state.libc_state = libc_not_started;
 
-    a_inc(&ethreads_minus_1);
-
 #ifdef DEBUG
     /* Make sure verbosity is off before loading the config (we don't know
      * whether it's enabled yet).*/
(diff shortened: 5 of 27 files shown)


.gitmodules
```diff
@@ -7,5 +7,5 @@
 	url = https://github.com/lsds/musl.git
 [submodule "sgx-lkl-musl"]
 	path = sgx-lkl-musl
-	url = https://github.com/lsds/sgx-lkl-musl.git
-	branch = oe_port
+	url = https://github.com/vtikoo/sgx-lkl-musl.git
+	branch = move-lthread-out
```
davidchisnall: Don't forget to remove these in the non-draft version!

src/enclave/enclave_event_channel.c
```diff
@@ -1,12 +1,13 @@
 #include <atomic.h>
 #include <string.h>
 
-#include "pthread_impl.h"
+// #include "pthread_impl.h"
```
davidchisnall: Please just delete these.  We can always find them in the history.

src/enclave/enclave_init.c
```diff
@@ -240,9 +241,10 @@ int __libc_init_enclave(int argc, char** argv)
     max_lthreads = next_pow2(max_lthreads);
 
     newmpmcq(&__scheduler_queue, max_lthreads, 0);
-
+    
+    init_ethread_tls();
     __init_libc(envp, argv[0]);
```
davidchisnall: Can this be moved after LKL start yet?

src/include/enclave/lthread.h
```diff
@@ -137,7 +137,7 @@ struct lthread
     char funcname[64];            /* optional func name */
     struct lthread* lt_join;      /* lthread we want to join on */
     void** lt_exit_ptr;           /* exit ptr for lthread_join */
-    locale_t locale;              /* locale of current lthread */
+    //locale_t locale;              /* locale of current lthread */
```
davidchisnall: We should also be able to remove `cancelbuf`, `stdio_locks`, `dlerror_buf`, `dlerror_flag`, `err`.  I'm not sure what `dtv` and `dtv_copy` are, we may also be able to remove them.

src/include/enclave/lthread.h
```diff
@@ -185,11 +185,83 @@ struct lthread_sched
...
+struct schedctx {
+	/* Part 1 -- these fields may be external or
+	 * internal (accessed via asm) ABI. Do not change. */
+	struct schedctx *self;
+	uintptr_t *dtv;
+	void *unused1, *unused2;
+	uintptr_t sysinfo;
+	uintptr_t canary, canary2;
+
+	/* Part 2 -- implementation details, non-ABI. */
```
davidchisnall: Note that most of these fields are there only because `schedctx` is the same layout as the TLS segment in the old implementation, so that it can live in the same place.  Most of them can be removed.  In fact, I think `lthread_sched` is the only thing that we actually need here (though the initial self pointer is also useful because x86 doesn't give you any other easy way of accessing things in the gs segment).

src/include/enclave/lthread.h
```diff
@@ -185,11 +185,83 @@ struct lthread_sched
...
+	uintptr_t *dtv_copy;
+    struct lthread_sched sched;
+};
+
+/* Thread Control Block (TCB) for lthreads */
+struct lthread_tcb_base {
+    void *self;
+    char _pad_0[32];
+    // SGX-LKL does not have full stack smashing protection (SSP) support right
+    // now. In particular, we do not generate a random stack guard for every
```
davidchisnall, line 194: Hopefully this comment can be obsolete soon.  Your pthread changes should set one up early and we can generate a canary value for each lthread that isn't created via the primitive call.

src/include/enclave/lthread.h
```diff
@@ -185,11 +185,83 @@ struct lthread_sched
...
+    // (TCB) layout. Among other things, it expects a read-only stack
+    // guard/canary value at an offset 0x28 (40 bytes) from the FS segment
+    // base/start of the TCB (see schedctx struct above).
+    uint64_t stack_guard_dummy; // Equivalent to schedctx->canary (see above).
+                                // canary2 is only used on the x32 arch, so we
+                                // ignore it here.
+    struct schedctx *schedctx;
+};
+
+/* Thread Control Block (TCB) for ethreads/the scheduler (schedctx) */
```
davidchisnall: I guess this comment is out of date.  Is this used at all?

src/include/enclave/lthread.h
```diff
@@ -185,11 +185,83 @@ struct lthread_sched
...
+    uint64_t stack_guard_dummy; // See struct lthread_tcb_base comment
+    struct schedctx *schedctx;
+};
+
 typedef struct lthread* lthread_t;
 #ifdef __cplusplus
 extern "C"
 {
 #endif
+    void init_ethread_tls();
```
davidchisnall: Please add a doc comment here.

src/include/enclave/lthread_int.h
```diff
@@ -71,4 +71,20 @@ static inline uint64_t _lthread_timespec_to_usec(const struct timespec* ts)
     return (ts->tv_sec * 1000000) + ts->tv_nsec / 1000;
 }
 
+// TODO: should this be static?
+static inline struct schedctx *__scheduler_self()
```
davidchisnall, line 78: Please add a doc comment.  It should be `static inline` so that it can be inlined everywhere (it's a single instruction after inlining) and the compiler won't complain about it being in a header.

src/include/enclave/lthread_int.h
```diff
@@ -71,4 +71,20 @@ static inline uint64_t _lthread_timespec_to_usec(const struct timespec* ts)
...
+	struct schedctx *self;
+	__asm__ __volatile__ ("mov %%gs:48,%0" : "=r" (self) );
+	return self;
+}
+
+
+static inline struct lthread_sched*
+lthread_get_sched()
+{
+    struct schedctx *c = __scheduler_self();
```
davidchisnall, line 89: Are there noe any other callers of `__scheduler_self`?  If not, it can be inlined here.

src/sched/lthread.c
```diff
@@ -63,7 +63,33 @@ static void lthread_rundestructors(struct lthread* lt);
...
+}
+
+/* Initialisation of ethread/scheduler TLS */
+void init_ethread_tls()
+{
+	void *mem = __scheduler_self();
+
+	/* Failure to initialize thread pointer is always fatal. */
+	if (__init_ethread_tp(mem) < 0)
+		a_crash();
```
davidchisnall: This is dead code, `__init_ethread_tp` unconditionally returns 0 so this is never reached.

It's probably a good idea to simply inline `__init_ethread_tp` here.

src/sched/lthread.c
```diff
@@ -461,6 +487,50 @@ void _lthread_free(struct lthread* lt)
...
+int __init_utp(void *p, int set_tp)
+{
+	struct lthread_tcb_base *tcb = (struct lthread_tcb_base *)p;
+	tcb->self = p;
+	tcb->schedctx = __scheduler_self();
+	if (libc.user_tls_enabled && set_tp)
+	{
+		if (sgxlkl_enclave->mode == SW_DEBUG_MODE)
+		{
+			int r = __set_thread_area(TP_ADJ(p));
```
davidchisnall: As I recall, this is implemented in musl.  If so, we need to remove it.  I believe that, in software mode, this should do a host syscall to set the %fs register value.

src/sched/lthread.c
```diff
@@ -461,6 +487,50 @@ void _lthread_free(struct lthread* lt)
     lt = 0;
 }
 
+int __init_utp(void *p, int set_tp)
+{
+	struct lthread_tcb_base *tcb = (struct lthread_tcb_base *)p;
+	tcb->self = p;
+	tcb->schedctx = __scheduler_self();
+	if (libc.user_tls_enabled && set_tp)
```
davidchisnall: The first of these checks should be removed.  We want to set up a TLS area with just a stack canary for threads that are not created with `lthread_create_primitive`.

src/sched/lthread.c
```diff
@@ -461,6 +487,50 @@ void _lthread_free(struct lthread* lt)
...
+		}
+		else
+		{
+			__asm__ volatile("wrgsbase %0" ::"r"(p));
+		}
+	}
+	return 0;
+}
+
+void *__copy_utls(struct lthread *lt, unsigned char *mem, size_t sz)
```
davidchisnall: Most of this should go away.  Threads created in musl will handle this, so this is just for kernel and host-interface threads.  The only thing that they need is a new random stack canary value.

src/sched/lthread.c
```diff
@@ -637,47 +707,37 @@ int lthread_create_primitive(
...
-        if (__init_utp(__copy_utls(lt, lt->itls, lt->itlssz), 0))
-        {
-            oe_free(lt);
-            return -1;
-        }
-    }
+    //Rm tls_size from lt 
+    // lt->itlssz = libc.tls_size;
+    lt->itls = tls;
+    // if (libc.tls_size)
```
davidchisnall: You can also remove most of the equivalent changes in `lthread_create`.  That function should only be called for non-userspace threads and so doesn't need to touch any libc state.

davidchisnall (commented):
Thanks for working on this, looks like a great start.

src/sched/lthread.c
```diff
@@ -732,16 +792,16 @@ int lthread_create(
...
-    {
-        for (FILE* f = *__ofl_lock(); f; f = f->next)
-            init_file_lock(f);
-        __ofl_unlock();
-        init_file_lock(__stdin_used);
-        init_file_lock(__stdout_used);
-        init_file_lock(__stderr_used);
-        libc.threaded = 1;
-    }
+    // if (!libc.threaded && libc.threads_minus_1 >= 0)
```
jxyang: We don't need to initialize the locks because `pthread_create` does this? What about the lthreads we created directly without going through libc?

davidchisnall: We will be doing that only for kernel threads, which should not hold any userspace libc locks.

letmaik:
@vtikoo Could you set the PR to draft mode if it's not ready for merging yet?

src/sched/lthread.c
vtikoo: `lthread_create` calls this function with set_tp set to 0. Should we remove the entire block?

src/sched/lthread.c
```diff
@@ -474,23 +476,32 @@ int __init_utp(void *p, int set_tp)
...
 void set_tls_tp(struct lthread* lt)
 {
+    uintptr_t tp_unaligned, tp;
     if (!lt->itls)
         return;
-
-    uintptr_t tp_unaligned =
+    // pthread_create passes thread pointer for lt->itls
+    // no need to calculate offset of tp in tls region
+    if (lt->attr.stack_size == 0)
```
davidchisnall: Sorry, I may have just broken this logic with the deferred stack freeing PR.  It's probably worth adding a flag in the lthread structure to check if this is a cloned task or not.  That said, I'd prefer if we always just set `tp = lt->itls` unconditionally and modified the other code path that set itls to do the adjustment.

src/sched/lthread.c
```diff
@@ -474,23 +476,32 @@ int __init_utp(void *p, int set_tp)
...
-    struct lthread_tcb_base* tp =
-        (struct
-         lthread_tcb_base*)(tp_unaligned - (tp_unaligned & (libc.tls_align - 1)));
-
-    tp->schedctx = __scheduler_self();
+        tp =
+            (struct
+            lthread_tcb_base*)(tp_unaligned - (tp_unaligned & (TLS_ALIGN - 1)));
 
+        tp->schedctx = __scheduler_self();
```
davidchisnall: This should no longer be required.  I don't think we ever read this field, we should always be getting the scheduler via %gs.  If you remove this, does anything break?

src/sched/lthread.c
```diff
@@ -664,6 +675,7 @@ int lthread_create_primitive(
     lt->itlssz = libc.tls_size; // not setting this causes problems in set_tls_tp
```
davidchisnall: I think your changes above make this unnecessary.

src/sched/lthread.c
```diff
@@ -664,6 +675,7 @@ int lthread_create_primitive(
...
     LIST_INIT(&lt->tls);
     lt->attr.state = BIT(LT_ST_READY);
+    lt->attr.stack_size = 0; // lthread doesn't manage userspace thread stacks
```
davidchisnall: The lthread should be allocated with `oe_calloc`, so this should be dead code.

src/sched/lthread.c
```diff
@@ -733,8 +745,8 @@ int lthread_create(
     lt->attr.stack_size = stack_size;
 
     /* mmap tls image */
-    lt->itlssz = libc.tls_size;
-    if (libc.tls_size)
+    lt->itlssz = (sizeof(lthread_tcb_base) + TLS_ALIGN - 1) & -TLS_ALIGN;
+    if (lt->itlssz)
```
davidchisnall: This logic looks confusing now.  We're setting `lt->itlssz` to a constant and then branching on that constant being non-zero.  I think the body of this if statement is reached unconditionally, so we should just remove this line.

src/sched/lthread.c
```diff
@@ -733,8 +745,8 @@ int lthread_create(
...
     {
         if ((lt->itls = (uint8_t*)enclave_mmap(
```
davidchisnall: If we do the adjustment once here (which is a very cold code path), we can avoid the conditional in the `set_tls_tp` function (which is a hot path, hit on every context switch).

src/sched/lthread.c
vtikoo: Good point!

src/include/enclave/lthread.h
vtikoo: This is being used to setup the pointer for schedctx for the ethread in enclave_oe.c. But that can probably be moved to use lthread_tcb_base or some inline assembly. Any thoughts?

src/sched/lthread.c
vtikoo: Removing both copy_utls and __init_utp, as they are effectively serving no purpose currently. The only usage in lthread_create calls __init_utp with set_tp = 0, at which point both of these functions are only doing pointer manipulations on local variables.

src/sched/lthread.c
vtikoo: Adding thread_type to lthread_attr

src/sched/lthread.c
vtikoo: Should be safe to remove, will do so.

src/sched/lthread.c
```diff
@@ -672,10 +605,10 @@ int lthread_create_primitive(
     }
 
     lt->itls = tls;
-    lt->itlssz = libc.tls_size; // not setting this causes problems in set_tls_tp
+    lt->tp = tls; // for cloned threads, the tls argument is a ptr to the TCB
```
davidchisnall: Please expand TCB: in a project like this, Thread Control Block and Trusted Computing Base are both in use and the initialism is confusing!

src/include/enclave/lthread_int.h
vtikoo: Its also used to setup the `schedctx` struct in `enclave_oe.c:init_ethread_tp`.

src/include/enclave/lthread_int.h
vtikoo: Added docs.

src/include/enclave/lthread.h
vtikoo: Do we really need the self pointer? schedctx is located at the end of the ethread gsbase pointed page. And is accessed via a pointer placed at gs:48.

davidchisnall: No, if we're just storing a pointer to this in gs:48 and not the structure in gs, we don't need the self pointer.  The self pointer was there because x86 doesn't have instructions for translating from a segment-relative address to a linear address.  We were storing this structure in the FS segment and the only way of loading a pointer to it was to load the self pointer (%fs:0).  We now have a pointer to it in %gs, so we don't need the self pointer.

src/sched/lthread.c
```diff
@@ -580,13 +588,13 @@ int _lthread_sched_init(size_t stack_size)
...
+    struct lthread_sched* sched = lthread_get_sched();
 
-    c->sched.stack_size = sched_stack_size;
+    sched->stack_size = sched_stack_size;
 
-    c->sched.default_timeout = 3000000u;
+    sched->default_timeout = 3000000u;
 
-    memset(&c->sched.ctx, 0, sizeof(struct cpu_ctx));
+    memset(&sched->ctx, 0, sizeof(struct cpu_ctx));
```
davidchisnall: Not very important, but I wonder if this is actually necessary.  This field should be fully set by `_switch` the first time this scheduler runs an lthread and the zero value should never be read.

src/sched/lthread.c
```diff
@@ -605,14 +613,16 @@ int lthread_create_primitive(
...
+    // for cloned threads, tls argument is a ptr to the Thread Control Block
+    lt->tp = tls;
     LIST_INIT(&lt->tls);
     lt->attr.state = BIT(LT_ST_READY);
     lt->attr.thread_type = USERSPACE_THREAD;
     lt->tid = a_fetch_add(&spawned_lthreads, 1);
 
     static unsigned long long n = 0;
-    snprintf(lt->funcname, 64, "cloned host task %llu", __atomic_fetch_add(&n, 1, __ATOMIC_SEQ_CST));
+    snprintf(lt->funcname, 64, "cloned host task %llu",
```
davidchisnall: This probably should be `oe_snprintf` (my fault, I think).

davidchisnall (commented):
Diffs that mostly delete code make me happy!

src/enclave/enclave_init.c
```diff
@@ -148,6 +148,9 @@ static int startmain(void* args)
         sgxlkl_info("wg0 has public key %s\n", key);
     }
 
+    /* Set locale for usersapce components using it */
+    pthread_t self = __pthread_self();
+    self->locale = &libc.global_locale;
```
davidchisnall, line 112: This setup should be done in the start libc routine in musl.

tests/basic/clone/clone.c
```diff
@@ -125,7 +125,7 @@ int main(int argc, char** argv)
 	fprintf(stderr, "Clone returned %d, ctid: %d ptid: %d\n", clone_ret, ctid, ptid);
 	assert(ctid == clone_ret, "ctid is %d, should be %d\n", ctid, clone_ret);
 	int futex_ret = futex_wait(&ctid, clone_ret);
-	assert(futex_ret == 0, "futex syscall returned %d (%s)\n", strerror(errno));
+	assert(futex_ret == 0, "futex syscall returned %d\n", futex_ret);
```
davidchisnall: The format string was wrong (ooops, my fault, sorry!), but this is in the main thread so strerror should be safe to use.

src/sched/lthread.c
```diff
@@ -566,6 +566,9 @@ int lthread_create_primitive(
     lt->itls = tls;
     // for cloned threads, tls argument is a ptr to the Thread Control Block
     lt->tp = tls;
+    // set self pointer for fs segment
+    struct lthread_tcb_base *tcb = (struct lthread_tcb_base *)lt->tp;
+    tcb->self = lt->tp;
```
davidchisnall: Musl should be doing this for pthreads, is this just needed for the clone test? If so, we should probably fix that test (I think it’s actually fine to remove it and replace it with a pthread test that does the same thing, it’s only needed because pthreads weren’t using clone when I wrote it and we needed to test clone.).

davidchisnall:
Please can you create a draft musl PR for the musl changes so that I can review them as well?  Thanks.

vtikoo:
@davidchisnall the musl PR is here - https://github.com/lsds/sgx-lkl-musl/pull/18

src/sched/lthread.c
vtikoo: `tests/basic/pthread_tls` probably covers the behaviors the clone test covers - userspace thread creation and waiting on futex. It can't cover the deferred stack unmapping test though.

src/enclave/enclave_init.c
vtikoo: Oh okay, makes sense. I'll look for the appropriate place within musl to set this.

.azure-pipelines/other/disabled_tests.txt
```diff
@@ -1,3 +1,4 @@
+tests/basic/pthread_join/Makefile
```
vtikoo: This test is disabled until https://github.com/lsds/sgx-lkl/issues/686 is fixed.

src/enclave/enclave_init.c
```diff
@@ -90,6 +91,8 @@ static void init_wireguard()
 
 static int startmain(void* args)
 {
+    __init_libc(sgxlkl_enclave_state.elf64_stack.envp,
```
vtikoo, line 94: This can't be moved any further than this right now. `lkl_start_init` does some `getenv` calls which requires libc to be initialized.

src/enclave/enclave_event_channel.c
```diff
@@ -1,12 +1,11 @@
 #include <atomic.h>
 #include <string.h>
 
-#include "pthread_impl.h"
-
 #include <lkl/virtio.h>
 
 #include "enclave/enclave_util.h"
 #include "enclave/lthread.h"
+#include "enclave/lthread_int.h"
```
jxyang, line 8: Why do we need the internal lthread header file?

src/enclave/enclave_init.c
jxyang: This is for the `startmain` lthread which is created before libc setup, right? Are you trying to wrap it inside a pthread and patch it up? Where and why do we expect `pthread->locale` to be set?

src/enclave/enclave_init.c
```diff
@@ -150,9 +156,10 @@ int __libc_init_enclave(int argc, char** argv)
     max_lthreads = next_power_of_2(max_lthreads);
 
     newmpmcq(&__scheduler_queue, max_lthreads, 0);
+    
+    init_ethread_tp();
 
-    __init_libc(envp, argv[0]);
-    __init_tls();
+    __init_heap_from_libc();
```
jxyang: Maybe some explanation on why this is needed.

src/enclave/enclave_event_channel.c
jxyang, line 8: Why do we need this internal header file?

src/enclave/enclave_oe.c
```diff
@@ -399,3 +452,12 @@ void sgxlkl_free_enclave_state()
 
     _free_shared_memory();
 }
+
+void cleanup_sgxlkl_enclave_config()
```
jxyang: Seems this function is not called anywhere?

src/sched/lthread.c
```diff
@@ -294,10 +308,12 @@ void _lthread_desched_sleep(struct lthread* lt)
     ticket_unlock(&sleeplock);
 
     SGXLKL_TRACE_THREAD(
```
jxyang: Why do we need two log messages for this simple function?

src/sched/lthread.c
```diff
@@ -337,45 +353,27 @@ void _lthread_yield(struct lthread* lt)
...
-        void (*f)(void*) = lt->cancelbuf->__f;
-        void* x = lt->cancelbuf->__x;
-        lt->cancelbuf = lt->cancelbuf->__next;
-        f(x);
-    }
     if (lthread_self() != NULL)
         lthread_rundestructors(lt);
-    if (lt->itls != 0)
+    
+    // lthread only manages tls region for lkl kernel threads
```
jxyang, line 354: Do we really have kernel lthread using TLS?

src/sched/lthread.c
```diff
@@ -590,51 +590,21 @@ int lthread_create_primitive(
...
-            oe_free(lt);
-            return -1;
-        }
-        if (__init_utp(__copy_utls(lt, lt->itls, lt->itlssz), 0))
-        {
-            oe_free(lt);
-            return -1;
-        }
-    }
+    lt->itls = tls;
```
jxyang: For my understanding, why do we need two pointers to the same TLS?

src/sched/lthread.c
jxyang: The self pointer is already set by musl here: https://github.com/lsds/sgx-lkl-musl/blob/f437aa45469273c414ceba14c884ec7d126f912a/src/thread/pthread_create.c#L265

src/sched/lthread.c
```diff
@@ -720,34 +682,26 @@ int lthread_create(
     lt->attr.stack_size = stack_size;
 
     /* mmap tls image */
-    lt->itlssz = libc.tls_size;
-    if (libc.tls_size)
-    {
-        if ((intptr_t)(lt->itls = (uint8_t*)enclave_mmap(
+    lt->itlssz = (sizeof(struct lthread_tcb_base) + TLS_ALIGN - 1) & -TLS_ALIGN;
```
jxyang, line 673: Can you explain how do we calculate the itlssz here, probably with a comment?

src/sched/lthread.c
```diff
@@ -494,10 +493,23 @@ int _lthread_resume(struct lthread* lt)
...
+    // The first "startmain" thread eventually loads the app's ELF image
+    // and initializes its tls area. As lthread has to properly set the
+    // tls region on context switches, check if the fs has changed and
+    // update the lthread's thread pointer field accordingly.
+    if (sched->current_lthread->tid == 1 &&
+        (sched->current_lthread)->attr.thread_type == LKL_KERNEL_THREAD){
+        void* fs_ptr;
+        __asm__ __volatile__("mov %%fs:0,%0" : "=r"(fs_ptr));
+        if (fs_ptr != sched->current_lthread->tp){
+            sched->current_lthread->tp = fs_ptr;
```
jxyang, line 492: Isn't this initialized during lthread_create (with the init_tp call)?

jxyang:
I will take a look of the musl changes later.

src/enclave/enclave_event_channel.c
vtikoo: Its relying on it for two macros: `BIT` and `CLEARBIT`, and `_lthread_yield_cb()`.

src/enclave/enclave_init.c
vtikoo: Sure. I had added some documentation in `__libc_startmain.c` in musl, where this function is defined. Will add here as well. This should anyway be a temporary fix until memory corruption issues are root caused.

src/sched/lthread.c
vtikoo: Thats a good question, I'm not sure. @davidchisnall do you know if we need `itls` for kernel threads?

src/enclave/enclave_oe.c
vtikoo: Resolved with - https://github.com/vtikoo/sgx-lkl/commit/dd89071da31cdeae5217784c2f4cb5eaecd3bbf3

src/sched/lthread.c
vtikoo: `init_tp` calculates the thread pointer value and sets a self pointer, but it does not set fsbase. FS register is set during context switches in `set_tls_tp`.

This above code is required because for the `startmain` thread, the dynamic loader creates a new tls region before passing execution to it - [here](https://github.com/vtikoo/sgx-lkl-musl/blob/move-lthread-out/ldso/dynlink.c#L1951-L1959).
lthreads needs to save that new pointer.

src/sched/lthread.c
davidchisnall: We need the stack canary value, that should be all.  Eventually, we should just populate a page or two with random numbers on boot and pick a random offset in that region for the itls pointer such that the stack canary lines up somewhere inside it, but for now we need one.

vtikoo: We also need a self pointer in `itls` for `set_tls_tp` to work.

src/sched/lthread.c
vtikoo: This was added to support the clone test, I'll move this logic there. Should we add an assert here to check if the self pointer has been set?

src/sched/lthread.c
vtikoo: Added comment.

davidchisnall (approved):

src/enclave/enclave_init.c
mikbras: Don't worry. My feature.split branch will re-organize this after check in.

src/enclave/enclave_init.c
```diff
@@ -150,9 +156,8 @@ int __libc_init_enclave(int argc, char** argv)
     max_lthreads = next_power_of_2(max_lthreads);
 
     newmpmcq(&__scheduler_queue, max_lthreads, 0);
-
-    __init_libc(envp, argv[0]);
-    __init_tls();
```
mikbras, line 155: Just double checking. Has the call to this function gone away completely? (__init_tls)?

src/enclave/enclave_oe.c
```diff
@@ -48,6 +55,54 @@ static int _strncmp(const char* x, const char* y, size_t n)
     return *px == *py ? 0 : *px < *py ? -1 : +1;
 }
 
+static void
+init_auxv(size_t *auxv, char* buf_ptr, char *pn)
+{
+    // By default auxv[AT_RANDOM] points to a buffer with 16 random bytes.
+    uint64_t *rbuf = (uint64_t*)buf_ptr;
+    buf_ptr += 16;
```
mikbras, line 63: Is this buffer 16-byte aligned? I think the ABI requires it.

src/enclave/enclave_oe.c
```diff
@@ -48,6 +55,54 @@ static int _strncmp(const char* x, const char* y, size_t n)
...
+    auxv[17] = 0;
+    auxv[18] = AT_UID;
+    auxv[19] = 0;
+    auxv[20] = AT_RANDOM;
+    auxv[21] = (size_t)rbuf;
+    auxv[22] = AT_HW_MODE;
+    auxv[23] = !sgxlkl_in_sw_debug_mode();
+    auxv[24] = AT_NULL;
+    auxv[25] = 0;
+}
```
mikbras, line 104: The following seem to be missing: 
- AT_BASE -- base address of image
- AT_SYSINFO_EHDR -- the ELF header address
- AT_PHDR - the program header address
- AT_PHNUM - the number of program headers
 - AT_ENTRY - address of the entry point

These are usually passed by the loader. It may work without them for now, but these may be needed later when we want to just call _start().

You can check in without these I think.

mikbras (changes requested):
Vikas. If you can verify that auxv is 16-byte aligned, I will approve. Also I notice that some of the standard AT_* variables are missing in auxv. This is fine for now but when we back out certain musl libc changes, they will be needed to find the elf header, program headers, and the elf entry point.

mikbras (approved):

vtikoo:
As discussed with @mikbras, #325 covers the effort to get to full ABI compatibility for the Stack on app entry.

jxyang (approved):

src/enclave/enclave_oe.c
davidchisnall: `AT_PHDR` is needed for unwind libraries.  We probably can't throw an exception through `qsort` without that.  That said, I've never actually seen code in the wild that wants to throw exceptions through libc functions outside of test suites: if you're using C++, there are significantly more efficient ways of sorting than calling libc's `qsort` and very few other libc functions take function pointer arguments.

PR #407, opened 2020-06-09
LTP: Fixed lchown03 and lremovexattr01 tests

vamsikrishna935:
Tests were failing with kernel panic in loop filesystem. 
So modified the tests to use root file system.


hukoyu (approved):
LGTM

PR #410, opened 2020-06-09
Bump sgx-lkl-musl submodule

letmaik:
From https://github.com/lsds/sgx-lkl/pull/377#discussion_r437247092.


letmaik:
This will be done in https://github.com/lsds/sgx-lkl/pull/381.

PR #418, opened 2020-06-10
LTP: Fix for sendto01

ashokchandran:
Issue:Test was using child process which is not possible in sgx. Due to product issues some subtests were failing.
Solution: In original test case the main process create a child process.
sgx-lkl supports single process environment.
The test case is modified to use a child
One of the sub test case designed to test generation of EFAULT
and exiting with non-zero exit code. In addition a test related to shutdown a local endpoint is also
disabled until issue 405 is resolved.

### DIFF
--- a/tests/ltp/patches/sendto01.patch
+++ b/tests/ltp/patches/sendto01.patch
@@ -16,7 +16,7 @@ In addition a test related to shutdown a local endpoint is also
 disabled until issue 405.
 url:https://github.com/lsds/sgx-lkl/issues/405
 diff --git a/testcases/kernel/syscalls/sendto/sendto01.c b/testcases/kernel/syscalls/sendto/sendto01.c
-index 6fe0274ee..69cde8361 100644
+index 6fe0274ee..702baeae4 100644
 --- a/testcases/kernel/syscalls/sendto/sendto01.c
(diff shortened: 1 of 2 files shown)


shaikshavali1 (commented):
Changes looks to be fine for me.

ashokchandran:
Please rerun the failed jobs. Appear to be a transient failure due to issue in enclave creation.

ashokchandran:
pipeline failures are in unrelated tests due to know issues in enclave initialization.

tests/ltp/patches/sendto01.patch
```diff
@@ -0,0 +1,294 @@
...
+-	 .flags = 0,
+-	 .to = (struct sockaddr_in *)-1,
+-	 .tolen = sizeof(sin1),
+-	 .retval = -1,
+-	 .experrno = EFAULT,
+-	 .setup = setup1,
+-	 .cleanup = cleanup1,
+-	 .desc = "invalid to buffer"}
+-	,
++//	/* Skip since uClinux does not implement memory protection */
```
hukoyu, line 138: If this code not needed then delete it instead of commenting out.

hukoyu (approved):
LGTM

PR #419, opened 2020-06-10
CI: Fix pkill logic

letmaik:
There were two bugs:
1. A process might be running during `pgrep` but may have exited when calling `pkill`, which in turn would then fail with exit code 1 (no process found).
2. pkill does not block until the process exited. The exit code does not signal whether killing was successful.

This PR fixes both issues by ignoring the exit code of `pkill` and instead waiting for 5 seconds and then checking running processes again with `pgrep`.


hukoyu (approved):
LGTM

Issue #105, opened 2020-04-29
[Build] Should OE be built as a submodule of SGX-LKL 


jhand2:
Currently OE is included from its install location (/opt/openenclave) and the expectation is that it was previously installed on the machine on which SGX-LKL is being built.

OE should be included as a submodule of SGX-LKL and built as part of the SGX-LKL build process similar to other dependencies such as musl and lkl.

Please know that until this is done, SGX-LKL might break with no warning any time OE changes. Breaking changes are announced by OE on a release cycle but depending on the tip of master gives very few stability guarantees for a pre1.0 project.

prp:
I would like to revive the discussion if we should build the SGX-LKL fork of OE as a submodule of SGX-LKL. I think that it would make building more user-friendly and less error-prone while we are using a version of OE that isn't master. If we make this change, we just need to ensure that this will remain compatible with future modifications of the build system.

cc: @paulcallen @davidchisnall @letmaik @wintersteiger @jxyang

letmaik:
Makes sense to me. As long as we add a build option like OE_PREFIX to optionally pick up an already installed OE it should cover all dev use cases.

wintersteiger:
Yeah, I agree; we need `OE_PREFIX` to override the path when necessary, but in general I think it's a good idea, at least until we depend on OE master instead of our own branches.

prp:
Fixed by #423

PR #423, opened 2020-06-10
Make Open Enclave a Git submodule

prp:
This PR makes the Open Enclave dependency a Git submodule. Overall this makes the build process more user-friendly.

The correct OE feature branch for SGX-LKL is built as part of the regular build process. The OE build dependencies must be installed by hand. The PR also updates the documentation and the CI pipeline.

Unresolved issues:
- [x] the instructions for debugging and `sgx-lkl-gdb` still expects OE to be installed under `/opt/openenclave` -- we should decide if we want this to depend on a host-wide installation of OE -- we discussed this offline, and I will open a separate issue on this

Resolves https://github.com/lsds/sgx-lkl/issues/105


README.md
```diff
@@ -82,58 +82,29 @@ sgx-lkl-setup
...
+git clone --branch oe_port git@github.com:lsds/sgx-lkl.git
 cd sgx-lkl
 ```
 
-3. Build SGX-LKL in the source tree:
+3. Install the Open Enclave build dependencies
+```
+cd openenclave
+sudo scripts/ansible/install-ansible.sh
+sudo ansible-playbook scripts/ansible/oe-contributors-setup.yml
```
letmaik, line 104: Be careful here, this installs the driver as well. We may want to offer multiple options depending on the environment.

prp: Ok, this is just copied from what we currently have in the README file. What are the different scripts that we should mention here? I can then update the text as part of this PR.

letmaik: If you run on Azure, then scripts/ansible/oe-contributors-acc-setup-no-driver.yml since new ACC VMs already contain the driver. If you run outside Azure, then likely scripts/ansible/oe-contributors-setup.yml which unfortunately also installs the driver and they don't have a script without the driver.

prp: Done

Makefile
```diff
@@ -11,7 +11,17 @@ OESIGN_CONFIG_PATH = $(SGXLKL_ROOT)/config
...
+all: update-git-submodules $(addprefix $(OE_SDK_ROOT)/lib/openenclave/, $(OE_LIBS)) $(BUILD_DIR)/$(SGXLKL_LIB_TARGET_SIGNED) fsgsbase-kernel-module
+
+# Build and install Open Enclave locally
+$(addprefix $(OE_SDK_ROOT)/lib/openenclave/, $(OE_LIBS)):
+	cd $(OE_SUBMODULE) && git submodule update --recursive --init
+	# Don't build tests.
+	# TODO replace with build option https://github.com/openenclave/openenclave/issues/2894
+	cd $(OE_SUBMODULE) && sed -i '/add_subdirectory(tests)/d' CMakeLists.txt
+	mkdir -p $(OE_SUBMODULE)/build
+	cd $(OE_SUBMODULE)/build && cmake -G "Unix Makefiles" -DCMAKE_INSTALL_PREFIX=$(OE_SDK_ROOT) -DCOMPILE_SYSTEM_EDL=ON ..
```
letmaik: OE should be built in the same mode as SGX-LKL, e.g. `-DCMAKE_BUILD_TYPE=Debug`. In config.mak there's already an CMAKE_BUILD_TYPE variable you can use.

config.mak
```diff
@@ -138,16 +138,16 @@ else
 endif
 
 # OpenEnclave
-export OE_SDK_ROOT := $(shell pkg-config oeenclave-gcc --variable=prefix)
+OE_SUBMODULE := $(SGXLKL_ROOT)/openenclave
+OE_SDK_ROOT := ${BUILD_DIR}/openenclave
 export OE_SDK_INCLUDES := $(OE_SDK_ROOT)/include
 export OE_SDK_LIBS := $(OE_SDK_ROOT)/lib
```
letmaik, line 144: When is `export` needed and when not? Just our of curiosity.

src/Makefile
```diff
@@ -85,7 +85,7 @@ endif
 
 MAIN_OE_OBJS =$(addprefix $(LIB_SGXLKL_BUILD)/main-oe/,$(MAIN_OE_SRCS:.c=.o))
 
-OE_PKGCONFIG_OUTPUT := $(shell pkg-config oehost-gcc --libs)
+OE_PKGCONFIG_OUTPUT := $(shell PKG_CONFIG_PATH=${PKG_CONFIG_PATH}:$(OE_SDK_ROOT)/share/pkgconfig pkg-config oehost-gcc --libs)
```
letmaik:
```suggestion
OE_PKGCONFIG_OUTPUT := $(shell PKG_CONFIG_PATH=$(OE_SDK_ROOT)/share/pkgconfig:${PKG_CONFIG_PATH} pkg-config oehost-gcc --libs)
```

letmaik (approved):
Looks fine, just some minor comments.

Makefile
prp: Done

config.mak
prp: 'export' ensures that a variable is visible to a sub-make. The variable handling is currently a mess, but it's not worth cleaning this up.

src/Makefile
prp: Done

Makefile
```diff
@@ -11,7 +11,17 @@ OESIGN_CONFIG_PATH = $(SGXLKL_ROOT)/config
 .DEFAULT_GOAL:=all
 default: all
 
-all: update-git-submodules $(BUILD_DIR)/$(SGXLKL_LIB_TARGET_SIGNED) fsgsbase-kernel-module
+all: update-git-submodules $(addprefix $(OE_SDK_ROOT)/lib/openenclave/, $(OE_LIBS)) $(BUILD_DIR)/$(SGXLKL_LIB_TARGET_SIGNED) fsgsbase-kernel-module
+
+# Build and install Open Enclave locally
+$(addprefix $(OE_SDK_ROOT)/lib/openenclave/, $(OE_LIBS)):
+	cd $(OE_SUBMODULE) && git submodule update --recursive --init
```
wintersteiger: Shouldn't this rather be in the `update-git-submodules` target so it gets handled in the same way as the other submodules?

config.mak
```diff
@@ -138,16 +138,16 @@ else
...
-
-ifeq (${OE_SDK_ROOT},)
-    $(info Could not find the OpenEnclave SDK installation)
-    $(info Did you source {openenclave_root}/share/openenclave/openenclaverc?)
-    $(error SGX-LKL requires a modified version of the OpenEnclave SDK to run, which can be \
-            found at https://github.com/openenclave/openenclave on the feature/sgx-lkl-support branch)
-endif
+OE_OESIGN_TOOL_PATH := $(OE_SDK_ROOT)/bin
+
+OE_LIBS := enclave/liboecore.a \
```
wintersteiger, line 147: Just on the off-chance that you haven't really tested this: the order of these libraries isn't always obvious. E.g. I'm not 100% sure that `liboeenclave.a` should appear before the mbedtls libs.

.azure-pipelines/scripts/build.sh
```diff
@@ -33,7 +32,13 @@ JunitTestStarted "$test_name"
 
 # Ensure we have a pristine environment
 git submodule foreach --recursive git clean -xdf
```
wintersteiger: If a clean is necessary, should we also add a `git reset --hard HEAD`?

wintersteiger (approved):
I added a few non-blocking comments, but overall LGTM.

config.mak
letmaik: This is only used for makefile targets I think, not for linking.

.azure-pipelines/scripts/build.sh
letmaik: This line should be removed. Cleanup is already done in checkout_submodules.sh.

jhand2 (approved):

Makefile
prp: Fair point, I've moved this. (This is slightly different though, as these are additional submodules within the OE submodule.)

config.mak
prp: Yes, the order is irrelevant here. These are just build targets that trigger the OE build (but without proper dependency tracking, as this has to wait until we migrate to cmake.)

.azure-pipelines/scripts/build.sh
prp: This was already there before, but I've removed it now.

PR #427, opened 2020-06-10
Delete dead code.

davidchisnall:
Since we moved to virtio block devices, we no longer use the LKL
in-memory devices, so remove the associated code.


prp (approved):
LGTM

Issue #422, opened 2020-06-10
Easily allow preparing container-mountable installation


letmaik:
From https://github.com/lsds/sgx-lkl/issues/394#issuecomment-641777474:
> The Debian package contains a stand-alone distribution including bundled libraries that's suitable for mounting into a Docker container. Your local install isn't the same. See the .azure-pipelines/scripts/create_standalone_deb_pkg.sh script to create such a stand-alone package. It may be worth pulling this out of the CI scripts to allow easier local testing from a source build. Not necessarily the building of the deb package, but just the fix-up of the libraries (rpath etc.).

letmaik:
https://github.com/lsds/sgx-lkl/pull/428 got merged, what's missing now is developer docs. This should probably be part of #273.

PR #428, opened 2020-06-10
Add script to make installation self-contained

letmaik:
Fixes https://github.com/lsds/sgx-lkl/issues/422.

This PR moves the part that makes an SGX-LKL installation stand-alone out of the Debian packaging script into a separate script. This allows developers that build SGX-LKL from source to run Docker containers by mounting SGX-LKL into the container at `/opt/sgx-lkl`.

As part of this PR I also removed the build mode option when building the Docker wrapper images with `sgx-lkl-docker build-cc`. The build mode was previously used to construct the right installation prefix path for the ENTRYPOINT. By using a single expected SGX-LKL prefix and launching SGX-LKL with the bundled loader library it doesn't matter anymore what the hard-coded path to the loader is within `sgx-lkl-run-oe`. This change means that the Docker image can be used more flexibly without having to mess with overriding the ENTRYPOINT when launching the container. It is enough to bind-mount a different installation into the container at the same path (`/opt/sgx-lkl`).

### DIFF
--- a/tools/make_self_contained.sh
+++ b/tools/make_self_contained.sh
@@ -1,0 +1,140 @@
+#!/bin/bash
+
+set -e
+
+# This script transforms an SGX-LKL installation such that all shared library dependencies
+# are bundled in the installation prefix and SGX-LKL can be run stand-alone
+# without having other packages installed. The main purpose is to be able to
+# volume mount an SGX-LKL installation into a Docker container without requiring
+# the container to have any runtime dependencies installed.
+#
+# The folder structure of the self-contained installation is as follows:
+#
+# /opt/sgx-lkl
+#   bin/
+#     sgx-lkl-run-oe
+#     ...
+#   lib/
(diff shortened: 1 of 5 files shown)


tools/make_self_contained.sh
```diff
@@ -0,0 +1,140 @@
...
+#   tools/
+#
+# In general, this reflects a regular SGX-LKL installation tree plus the lib/external
+# folder containing all bundled libraries.
+#
+# Implementation notes:
+# 
+# Automatic dependency discovery relies on dynamically linked libraries, however, some
+# libraries like the Azure DCAP Client are dlopen'd at runtime.
+# Because of that, there is extra logic that deals with bundling them manually.
```
prp, line 35: Not as part of this PR, but at some point, we should also document the folder structure etc. in a file under `docs/`.

prp (approved):
LGTM, glad that the build mode is gone.

PR #429, opened 2020-06-10
LTP: Fixed link04 test case

vamsikrishna935:
Tests were failing as mmap is not supported. So modified the tests without mmap.


shaikshavali1 (commented):
Changes looks fine

hukoyu (approved):
LGTM

PR #440, opened 2020-06-11
Remove fsgsbase module from deb pkg

### LINKED ISSUE #437: Remove fsgsbase kernel module from Debian package
(from #435)

The sgx-lkl (clc) Debian package currently contains the binary fsgsbase kernel module because `make install` puts it in the install tree. When a user then runs `sgx-lkl-setup` then this script actually tries to modprobe the module, which will fail due to kernel version mismatches.

The suggested way to install the driver is through the DKMS Debian package. Therefore the kernel module should not be part of the Debian package and sgx-lkl-setup should not try to modprobe it if it doesn't exist.

@prp

letmaik:
Fixes #437.

### DIFF
--- a/tools/sgx-lkl-setup
+++ b/tools/sgx-lkl-setup
@@ -25,9 +25,11 @@ sudo iptables -t nat -A POSTROUTING -s 10.0.1.0/24 ! -d 10.0.1.0/24 -j MASQUERAD
 
 echo "Done"
 
-echo "Setting up Linux kernel module to support userspace FSGSBase..."
-sudo insmod ${SGXLKL_ROOT}/tools/kmod-set-fsgsbase/mod_set_cr4_fsgsbase.ko val=1
-sudo rmmod mod_set_cr4_fsgsbase
-dmesg | tail -n 10
-echo "Done"
+if [[ -f ${SGXLKL_ROOT}/tools/kmod-set-fsgsbase ]]; then
+    echo "Setting up Linux kernel module to support userspace FSGSBase..."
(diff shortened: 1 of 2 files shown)


tools/sgx-lkl-setup
```diff
@@ -25,9 +25,11 @@ sudo iptables -t nat -A POSTROUTING -s 10.0.1.0/24 ! -d 10.0.1.0/24 -j MASQUERAD
 
 echo "Done"
 
-echo "Setting up Linux kernel module to support userspace FSGSBase..."
-sudo insmod ${SGXLKL_ROOT}/tools/kmod-set-fsgsbase/mod_set_cr4_fsgsbase.ko val=1
-sudo rmmod mod_set_cr4_fsgsbase
-dmesg | tail -n 10
-echo "Done"
+if [[ -f ${SGXLKL_ROOT}/tools/kmod-set-fsgsbase ]]; then
```
prp, line 28: Wouldn't it be helpful to also output a message if the module has not been loaded, so the the user knows that SGX-LKL will either require a suitable installed kernel or emulate userspace FSGSBASE with a performance overhead?

prp (approved):
LGTM, left one comment

tools/sgx-lkl-setup
letmaik: Well, then you need to do the check to see if the module was loaded, e.g. by the DKMS package. I don't want to complicate this more. SGX-LKL already outputs a message if FSGSBASE is emulated.

prp: Ok, I was just thinking about outputting a message that no module was loaded.

letmaik: It's already documented on how to install and enable the fsgsbase dkms package when using the binary packages, that should be enough. For building from source it will be loaded.

PR #441, opened 2020-06-11
Install oe gdb plugin and bundle in debian package

### LINKED ISSUE #424: Integrate sgx-lkl-gdb with SGX-LKL source tree and Debian packages
Currently the `sgx-lkl-gdb` tool relies on an Open Enclave install under `/opt/openenclave` on the host. We will no longer assume that such an installation exists.

Proposed changes:

1. `sgx-lkl-gdb` should be included in the Debian packages
2. The behaviour of sgx-lkl-gdb should be changed as follows:
     - if `sgx-lkl-gdb` is invoked from the SGX-LKL source tree, it should use the OE version installed as part of the build tree for gdb support.
     - `sgx-lkl-gdb` in the Debian package should be self-contained, i.e., it should provide all dependencies as part of the package (e.g., the OE gdb `ptrace` library and the gdb Python plugin).

cc: @letmaik

letmaik:
Fixes #424.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -146,7 +146,7 @@ fsgsbase-kernel-module:
 	make -C ${TOOLS}/kmod-set-fsgsbase
 
 install:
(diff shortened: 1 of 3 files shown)


prp (approved):
LGTM

Issue #405, opened 2020-06-09
LTP: socket shutdown fails


ashokchandran:
Test:sendto01
purpose of the test: Verify that sendto() returns the proper errno for various failure cases
Issue: Test was using a parent child setup to test. patch was applied to make it thread-based. 
But still one of the sub test related to local endpoint shutdown was failing. 
The test does a shutdown and expect EPIPE . It was getting error ENOTCONN instead.
```
{.domain = PF_INET,
	 .type = SOCK_STREAM,
	 .proto = 0,
	 .buf = buf,
	 .buflen = sizeof(buf),
	 .flags = 0,
	 .to = &sin1,
	 .tolen = sizeof(sin1),
	 .retval = -1,
	 .experrno = EPIPE,
	 .setup = setup2,
	 .cleanup = cleanup1,
	 .desc = "local endpoint shutdown"}



static void setup1(void)
{
	s = SAFE_SOCKET(cleanup, tdat[testno].domain, tdat[testno].type,
			tdat[testno].proto);
	SAFE_CONNECT(cleanup, s, (const struct sockaddr *)&sin1, sizeof(sin1));
}
static void setup2(void)
{
	setup1();
	if (shutdown(s, 1) < 0)
		tst_brkm(TBROK | TERRNO, cleanup, "socket setup failed connect "
			 "test %d", testno);
}
```
logs after adding threads : Subtest for shutdown local endpoint failing.
```
[[  SGX-LKL ]] libc_start_main_stage2(): Calling app main: /ltp/testcases/kernel/syscalls/sendto/sendto01
sendto01    0  TINFO  :  Thread created
sendto01    1  TPASS  :  bad file descriptor successful
sendto01    2  TPASS  :  invalid socket successful
sendto01    3  TPASS  :  not connected TCP successful
sendto01    4  TPASS  :  invalid to buffer length successful
sendto01    5  TPASS  :  UDP message too big successful
sendto01    6  TBROK  :  sendto01.c:410: socket setup failed connect test 5: errno=ENOTCONN(107): Socket not connected
sendto01    7  TBROK  :  sendto01.c:410: Remaining cases broken
[[  SGX-LKL ]] lkl_terminate(): terminating LKL (exit_status=2)
```

ashokchandran:
additional logs
```
sendto01    5  TPASS  :  UDP message too big successful
[...]

PR #444, opened 2020-06-11
LTP: Fix for send01 and settimeofday01

ashokchandran:
send01
---------
Issue: In the original test case, the main process creates a child process.
sgx-lkl supports a single process environment.

One of the subtest cases designed to test the generation of EFAULT
by accessing invalid address values. Currently, sgx behavior is
to call enclave abort, if the address is not within the enclave address
Solution:
The test case is modified to use a child
pthread instead of forking a child process.Github issue169 (https://github.com/lsds/sgx-lkl/issues/169)
is raised to fix this behavior. The subtest cases which test
EFAULT error behavior is commented/disabled until GitHub
issue169 is fixed.
Also, a test related to shutdown a local endpoint is also
disabled until issue 405.
url:https://github.com/lsds/sgx-lkl/issues/405

settimeofday01:
----------------
Issue: One of the subtest cases designed to test the generation of EFAULT
by accessing invalid address values. Currently, sgx behavior is
to call enclave abort, if the address is not within the enclave address

Solution:Github issue169 (https://github.com/lsds/sgx-lkl/issues/169)
is raised to fix this behavior. The subtest cases which test
EFAULT error behavior is commented/disabled until GitHub
issue169 is fixed.

### DIFF
--- a/tests/ltp/patches/send01.patch
+++ b/tests/ltp/patches/send01.patch
@@ -16,17 +16,19 @@ Also, a test related to shutdown a local endpoint is also
 disabled until issue 405.
 url:https://github.com/lsds/sgx-lkl/issues/405
 diff --git a/testcases/kernel/syscalls/send/send01.c b/testcases/kernel/syscalls/send/send01.c
-index 2e0ae2177..b0e10028e 100644
+index 2e0ae2177..9d7caa4a3 100644
 --- a/testcases/kernel/syscalls/send/send01.c
 +++ b/testcases/kernel/syscalls/send/send01.c
-@@ -43,6 +43,7 @@
+@@ -43,7 +43,8 @@
  
  #include "test.h"
  #include "safe_macros.h"
+-
 +#include "pthread.h"
- 
++#include "tst_safe_pthread.h"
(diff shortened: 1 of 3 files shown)


tests/ltp/patches/send01.patch
```diff
@@ -0,0 +1,222 @@
...
+-		do_child();
+-#endif
+-		break;
+-	case -1:
+-		tst_brkm(TBROK | TERRNO, cleanup, "server fork failed");
+-	default:
+-		close(sfd);
+-		return pid;
++	 //start a child thread to create a directory
++	if(pthread_create(&tid, NULL, do_child_thread, NULL)== -1)
```
shaikshavali1: please use "SAFE_PTHREAD_CREATE".

shaikshavali1 (changes requested):
@ashokchandran , Please address the review comment.

tests/ltp/patches/send01.patch
ashokchandran: done

hukoyu (approved):
LGTM

PR #446, opened 2020-06-11
Add LGTM config

letmaik:
https://lgtm.com/help/lgtm/lgtm.yml-configuration-file

### DIFF
--- a/.azure-pipelines/scripts/build.sh
+++ b/.azure-pipelines/scripts/build.sh
@@ -9,7 +9,7 @@ if [ -z $SGXLKL_PREFIX ]; then
     exit 1
 fi
 
-SUDO=${SUDO:-sudo}
+[[ "$NOSUDO" == 1 ]] && SUDO="" || SUDO="sudo"
 
 . $SGXLKL_ROOT/.azure-pipelines/scripts/junit_utils.sh
 
--- a/.azure-pipelines/scripts/build.sh
+++ b/.azure-pipelines/scripts/build.sh
@@ -9,8 +9,6 @@ if [ -z $SGXLKL_PREFIX ]; then
     exit 1
 fi
 
-[[ "$NOSUDO" == 1 ]] && SUDO="" || SUDO="sudo"
-
 . $SGXLKL_ROOT/.azure-pipelines/scripts/junit_utils.sh
 
 # Initialize the variables.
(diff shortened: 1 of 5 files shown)


prp:
@letmaik Perhaps we should only analyse our code?

letmaik:
Yeah I'm trying to figure out if that's possible. It looks like the enclave code is excluded unfortunately. And it seems the source folder paths are somehow taken from the built artifacts and are mixed together partially. I'll see if this can be improved somehow, otherwise it may not be that useful after all...

letmaik:
Looks like LGTM runs inside a Docker container as non-root without sudo privileges. That's quite limited... And they only allow to install apt packages (declaratively) from the official channels, which wouldn't be enough for us (Intel libraries etc.). Since this is only building, not testing, I wonder if we can avoid such dependencies by using the SGX1 build option and ansible scripts of OE.

letmaik:
This is blocked now by https://github.com/openenclave/openenclave/issues/3127.
@jhand2 @jxyang 

EDIT: I worked around the OE issue with a nasty hack:
```sh
# Work around https://github.com/openenclave/openenclave/issues/3127
# by linking against a dummy library instead (libm) and disable relevant code paths.
sed -i 's/NAMES sgx_enclave_common/NAMES sgx_enclave_common m/g' openenclave/host/CMakeLists.txt
echo "target_compile_options(oehost PRIVATE -DOEHOSTMR)" >> openenclave/host/CMakeLists.txt
```

letmaik:
After resolving the OE issue with a hack, the next blocking issue is caused by LGTM itself:
https://discuss.lgtm.com/t/error-loading-shared-library-ld-linux-x86-64-so-2-no-such-file-or-directory-needed-by-opt-dist-tools-lib64trace-so/2944

davidchisnall:
What is the rationale for picking LGTM and not, for example, Coverity (also free for open source projects and with a long track record of finding security issues)?

aibaars:
> After resolving the OE issue with a hack, the next blocking issue is caused by LGTM itself:
> https://discuss.lgtm.com/t/error-loading-shared-library-ld-linux-x86-64-so-2-no-such-file-or-directory-needed-by-opt-dist-tools-lib64trace-so/2944

LGTM determines which files to analyse by tracing compiler calls that happen during the build. The tracer apparently does not work on binaries that were linked with `musl`. We'll look into this, but it may take a while for this to be addressed.

If I am not mistaken, the tracer is only active in the `configure` and `index` phases. To make things work now, you should make that no `musl`-linked binaries are run during the `configure` and `index` phases.

Looking at the log, the crash happens when `configure` tries to run `./a.out` which was generated by `musl-gcc` . Perhaps you could run that `configure` script  in the `after_prepare` phase which should run without the tracer. Alternatively, you could perhaps use plain `gcc` instead of `musl-gcc` .

letmaik:
@aibaars Thanks for looking into it. We won't change our build system just for LGTM. Building with musl is not something we can change here as it's a strict requirement. I'll close this PR until LGTM works for this scenario.

PR #447, opened 2020-06-11
LTP: fix test case getsockopt02 issue

shaikshavali1:
In original test case, child process is created. Child process
creates a client socket and connect to server socket owned by
parent process. The parent process will create a listening socket
and accepts connection from client socket owned by child process.
This test case performs two parallel jobs using parent and child
process. sgxlkl environment supports single process. This test case
is modified to use a pthread in place of child process.


hukoyu (approved):
LGTM

PR #448, opened 2020-06-11
Test turning off what I suspect is the source of a lot of hangs in CI

SeanTAllen:


jxyang:
I was looking at the same code today, and my conclusion is that's unlikely the cause of the hang.

SeanTAllen:
@jxyang I have a reproduceable case to show that it matches up with many of the CI hangs. I have recreated that hang locally umount sometimes hangs. I haven't figured out why.

It's possible that it is spooky action at a distance, but given the conditions that I have, I don't think so right now.

I can see run_syscall for umount2 being called and it never exits, however, the bug is very timing related and I'm unable to recreate in a debugger. I can go over it next week with you if you are interested.

PR #454, opened 2020-06-12
Rename Debian package from clc to sgx-lkl

### LINKED ISSUE #421: Debian packages: rename clc to sgx-lkl
Currently the Debian packages are named "clc" for historic reasons, but I think it makes more sense to name them after the project, becoming `sgx-lkl-debug` etc.

@davidchisnall @paulcallen @jxyang @agowdamsft

letmaik:
Fixes #421.


hukoyu (approved):
LGTM

Issue #456, opened 2020-06-12
Unsetting FSGSBASE crashes VMs


letmaik:
@mikbras @jhand2 @jxyang @davidchisnall @prp 

The following can be observed in CI, both SGX and non-SGX VMs, and was reproduced locally by @wintersteiger and me.

EDIT: The old kernel without the issue was 1022, the new one is 1028.

```
$ make -C tools/kmod-set-fsgsbase set-cr4-fsgsbase make: Entering directory '/data/user/git/sgx-lkl-oe/tools/kmod-set-fsgsbase'
make -C /lib/modules/5.3.0-1028-azure/build M=/data/user/git/sgx-lkl-oe/tools/kmod-set-fsgsbase modules
make[1]: Entering directory '/usr/src/linux-headers-5.3.0-1028-azure'
  Building modules, stage 2.
  MODPOST 1 modules
make[1]: Leaving directory '/usr/src/linux-headers-5.3.0-1028-azure'
sudo insmod ./mod_set_cr4_fsgsbase.ko val=1
sudo rmmod mod_set_cr4_fsgsbase
dmesg | tail -n 10
[34571.535696] mod_set_cr4_fsgsbase: module verification failed: signature and/or required key missing - tainting kernel
[34571.536951] SGX-LKL: Setting CR4.FSGSBASE to 1...
[34571.539370] SGX-LKL: Successfully set CR4.FSGSBASE to 1 on CPU #3, CR4 before: 0x003706e0, after: 0x003706e0.
[34571.543108] SGX-LKL: Successfully set CR4.FSGSBASE to 1 on CPU #0, CR4 before: 0x003706f0, after: 0x003706f0.
[34571.543129] SGX-LKL: Successfully set CR4.FSGSBASE to 1 on CPU #2, CR4 before: 0x003706e0, after: 0x003706e0.
[34571.543167] SGX-LKL: Successfully set CR4.FSGSBASE to 1 on CPU #1, CR4 before: 0x003706e0, after: 0x003706e0.
[34571.543168] SGX-LKL: Successfully set CR4.FSGSBASE to 1 on CPU #5, CR4 before: 0x003706e0, after: 0x003706e0.
[34571.543169] SGX-LKL: Successfully set CR4.FSGSBASE to 1 on CPU #6, CR4 before: 0x003706e0, after: 0x003706e0.
[...]

PR #457, opened 2020-06-12
Remove fsgsbase_unset test as new kernel has it built in

letmaik:
Fixes https://github.com/lsds/sgx-lkl/issues/456


wintersteiger (approved):
LGTM!

PR #462, opened 2020-06-15
LTP: Fix for renameat201,renameat202

ashokchandran:
Issue: Subtest related to RENAME_EXCHANGE flag in renameat2 system call were failing as file system was not supporting the flag. With this flag, an existing file at the new name will not be deleted; instead, it will be renamed to old name. 
Solution: Using root file system to test the API


vamsikrishna935 (approved):
@hukoyu LGTM

ashokchandran:
pipeline failure is in unrelated test languages-java-thread-(nonrelease)-(run-sw)-(8-ethreads) in enclave init sgxlkl_enclave_init

hukoyu (approved):
LGTM

PR #467, opened 2020-06-15
Add instructions for running cc test locally

letmaik:
Also fixes a minor issue when re-running `tools/make_self_contained.sh` after a re-install of SGX-LKL:

### DIFF
--- a/tests/containers/cc/Makefile
+++ b/tests/containers/cc/Makefile
@@ -36,22 +36,10 @@ run: run-hw run-sw
 
 ifeq (${SGXLKL_PREFIX},)
 run-hw:
-	@echo SGXLKL_PREFIX missing, skipping test.
-	@echo For local testing, run the following first:
-	@echo "> cd $$SGXLKL_ROOT"
-	@echo "> export SGXLKL_PREFIX=$$SGXLKL_ROOT/build/install"
-	@echo "> make install PREFIX=\$$SGXLKL_PREFIX"
-	@echo "> tools/make_self_contained.sh"
-	@echo "Then re-run the test with SGXLKL_PREFIX still set."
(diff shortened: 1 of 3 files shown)


tests/containers/cc/Makefile
```diff
@@ -35,10 +36,22 @@ run: run-hw run-sw
 
 ifeq (${SGXLKL_PREFIX},)
 run-hw:
-	@echo SGXLKL_PREFIX missing, skipping test
+	@echo SGXLKL_PREFIX missing, skipping test.
+	@echo For local testing, run the following first:
+	@echo "> cd $$SGXLKL_ROOT"
+	@echo "> export SGXLKL_PREFIX=$$SGXLKL_ROOT/build/install"
+	@echo "> make install PREFIX=\$$SGXLKL_PREFIX"
```
SeanTAllen: So if I understand this correctly, you can't test without doing an install? Is there a reason it's different than the others where you can test without doing install?

letmaik: That's correct. It is different because all other tests run SGX-LKL on the host, whereas this test runs SGX-LKL inside a Docker container where the SGX-LKL distribution/installation is mounted into it. Mounting the build folder is not possible.

SeanTAllen: I think that difference is worth the inclusion of a README that notes why it's different. A less knowledgeable person, such as myself might think that was unintentional.

Otherwise, this looks good.

letmaik: I moved it into a README. Could you have another look?

SeanTAllen (approved):
This looks great to me @letmaik. I'd say it is ready to merge.

PR #471, opened 2020-06-15
Allow overriding of OE_SDK_ROOT

wintersteiger:
This makes it easy to build with non-default OE_SDK_ROOT.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,7 @@ default: all
 
 all: update-git-submodules $(addprefix $(OE_SDK_ROOT)/lib/openenclave/, $(OE_LIBS)) $(BUILD_DIR)/$(SGXLKL_LIB_TARGET_SIGNED) fsgsbase-kernel-module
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -13,6 +13,7 @@ default: all
 
 all: update-git-submodules $(addprefix $(OE_SDK_ROOT)/lib/openenclave/, $(OE_LIBS)) $(BUILD_DIR)/$(SGXLKL_LIB_TARGET_SIGNED) fsgsbase-kernel-module
 
+ifeq ($(OE_SDK_ROOT),$(OE_SDK_ROOT_DEFAULT))
```
prp, line 17:
```suggestion
# Check if the user didn't override the default in-tree OE install location with another OE host install
ifeq ($(OE_SDK_ROOT),$(OE_SDK_ROOT_DEFAULT))
```

prp (approved):
LGTM, left one comment.

letmaik (approved):

PR #476, opened 2020-06-16
Extend sgx-lkl-cfg to add optional fields with defaults

letmaik:
Sample output:
```
$ tools/sgx-lkl-cfg add-optional --app-cfg tests/containers/cc/app-config.json
Adding optional fields in tests/containers/cc/app-config.json

Added $.cwd: "/"
Added $.exit_status: "full"
Added $.disk_config[0].overlay: false
Added $.disk_config[0].create: false
```

The field notation is JSONPath.

I also added a `--with-optional` flag to the `sgx-lkl-cfg create` command for convenience.

### DIFF
--- a/tools/schemas/app-config.schema.json
+++ b/tools/schemas/app-config.schema.json
@@ -54,9 +54,7 @@
                 "additionalProperties": false,
                 "dependencies": {
                     "roothash": ["roothash_offset"],
-                    "roothash_offset": ["roothash"],
-                    "create": ["size"],
-                    "size": ["create"]
+                    "roothash_offset": ["roothash"]
                 },
                 "properties": {
                     "disk": {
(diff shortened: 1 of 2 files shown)


prp:
I'm not sure that "add optional" is the right terminology here. Isn't this making the enclave_config "complete" by adding all fields (and setting them to defaults)? The tool would output this as a non-canonical version of the enclave_config, and a user would then edit some of the values, right?

letmaik:
Yes, I wasn't sure about the right word. Please suggest an alternative! :)

prp:
`--complete` and `make-complete`?

letmaik:
I'm fine with that, I'll wait a bit and see if anyone objects and whether I still like it in the afternoon...

letmaik:
I changed it to "complete", ready for review. I also fixed a minor schema issue.

prp (approved):
LGTM

PR #477, opened 2020-06-16
LTP: fix testcase pipe04 issue

shaikshavali1:
In original test case, parent process create a pipe
and fork 2 child processes. These two child processes
continuously write into pipe and parent process read
from this pipe for 100 times and kills both the child
process.
sgx-lkl supports a single process environment. Hence,
child processes are changed to child thread.


vamsikrishna935 (approved):
@hukoyu LGTM

hukoyu (approved):
LGTM

PR #481, opened 2020-06-17
Run Docker container as non-root user

### LINKED ISSUE #479: Run SGX-LKL inside Docker container as non-root
The image created by `sgx-lkl-docker build` should allow SGX-LKL to be run as a non-root process. This may already be the case but the existing test should be modified to check that it works.

letmaik:
This PR changes `sgx-lkl-docker build` so that the created Docker image runs the SGX-LKL process as non-root user.

In principle this works, but I'm having issues with the TAP device. When running as root user then everything is fine, but as non-root user I'm getting:
```
[   SGX-LKL  ] FAIL: Tap device sgxlkl_tap0 unavailable, ioctl("/dev/net/tun"), TUNSETIFF) failed: Operation not permitted
```

Things I've tried:
- `--cap-add=NET_ADMIN` when launching the container, same problem. This should only be needed for creating new TAP devices but for some reason I also need that when running as root.
- Remove the user part in `sudo ip tuntap add dev sgxlkl_tap0 mode tap user $(whoami)` in `sgx-lkl-setup`, same problem.

@prp @Francis-Liu Any idea?

Closes #479.


Francis-Liu:
Sorry, @letmaik, I have just seen your question. Can you try adding back the `--network=host` parameter? Did you run `sgx-lkl-setup` from the host?

~~With that parameter, your newly created container will be attached to the `host` network namespace. On `host` netns, you have already created the "sgxlkl_tap0" interface. Without `--network=host`, you create a container by attaching it to the default 'bridge' network, on which you only have created a `eth0` interface for the container.~~

Also, I want to understand if this error is because your docker instance does not have enough privilege. What happen if you add the `--privileged` option back?

prp:
@letmaik:

- When creating the `sgxlkl_tap0` device outside of the container on the host, are you making sure that the  username passed in is the same as the one used inside the container?

- What is the owner/permissions of `/dev/net/tun` inside the container?

- As @Francis-Liu said, does giving the container all permissions with `--privileged` make a difference?

letmaik:
I think I generally understand the permission situation now.

1. If the TAP device was created on the host, then you can bind to it if either:
a) you have `NET_ADMIN` capability, or
b) the process trying to bind it runs as the same user for which the TAP device was registered for.

2. If the TAP device does not exist yet, then you can create it if:
a) you have `NET_ADMIN` capability.

Side note: `ioctl(fd, TUNSETIFF, &ifr)` either binds or creates, depending on available permissions and whether the device exists already.

In general we don't want to expose unneeded capabilities to the container, and `NET_ADMIN` was used in the past for [privilege escalation](https://vigilance.fr/vulnerability/Linux-kernel-privilege-escalation-via-TUNSETIFF-13426).

I was able to use the existing TAP device with the following flags: `--device /dev/net/tun --network host -u $(UID)` where `UID=$(id -u)` is the id of the user for which the TAP device was created. This is progress.

In the context of Kubernetes, does that mean there has to be a host user with well-known user id which is then used for launching the container? I wasn't able to create a TAP device that can be used by all users. Leaving out the `user $(whoami)` part in `sudo ip tuntap add dev sgxlkl_tap0 mode tap user $(whoami)` didn't make a difference.

I changed the PR to run the container with the same user as on the host and that seems to work.

Francis-Liu:
> In the context of Kubernetes, does that mean there has to be a host user with well-known user id which is then used for launching the container? I wasn't able to create a TAP device that can be used by all users. Leaving out the user $(whoami) part in sudo ip tuntap add dev sgxlkl_tap0 mode tap user $(whoami) didn't make a difference.

@letmaik In Kubernetes, each Pod has its own network namespace. So the TAP device should be created within each Pod's netns, not sharing host netns as this PR did. `sudo ip tuntap add dev sgxlkl_tap0` will be called by an init container per each Pod, using a username as we defined. And we need to create a unique TAP interfaces for each confidential container. TAP interface on different POD can use the same interface name, as they are isolated by netns. Since all containers in a single Pod belong to one customer, we  can use the same username when creating those TAP interfaces.

letmaik:
@Francis-Liu It won't be enough to have the same user name, it's the user id that must match. Do you think that's doable?

Francis-Liu:
Thanks @letmaik for pointing this out. So far I have not yet removed `--previleged` from my SGX-LKL container spec. So it does not seem using a different user id to call ioctl with `/dev/net/tun` has caused any error. I will keep this in mind once I remove the `--previleged=true` security context from my deployment spec. (I am speaking container spec, security context, and deployment spec in the Kubernetes context)

Also, I do not understand how the user id within docker container would affect the permission to access an interface on host netns. Since user id(s) are separated by namespaces, namely the host and the docker instance maintains two different sets of user id(s).

letmaik:
> Also, I do not understand how the user id within docker container would affect the permission to access an interface on host netns. Since user id(s) are separated by namespaces, namely the host and the docker instance maintains two different sets of user id(s).

I think the issue is that you directly talk to `/dev/net/tun` on the host and this device will see the user id from the container, and if that doesn't match the one for which the TAP device was registered then you get permission errors.

tools/sgx-lkl-docker
```diff
@@ -44,22 +44,46 @@ function build() {
...
-FROM scratch
 # ca-certificates.crt is read by libcurl which is used by the Azure DCAP Client.
-COPY --from=0 /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
-COPY . /
 
-# Note: WORKDIR also creates the directory.
-WORKDIR /tmp
-ENV TMPDIR=/tmp
+RUN groupadd -g 999 sgxlkl && \
+    useradd -g 999 -u 999 sgxlkl
```
prp, line 50: Is that the convention what uid/gid is used in containers?

tools/sgx-lkl-docker
```diff
@@ -44,22 +44,46 @@ function build() {
...
+
+RUN mkdir /prep && \
+    cd /prep && \
+    mkdir etc && \
+    echo "sgxlkl:x:999:999::/tmp:/usr/sbin/nologin" > etc/passwd && \
+    echo "sgxlkl:x:999:" > etc/group && \
+    mkdir -p home/sgxlkl && \
+    chown sgxlkl:sgxlkl home/sgxlkl && \
```
prp, line 58: I wonder how robust that is. Isn't it better to invoke `useradd` with the correct parameters?

I actually don't understand why you do this as part of a multi-stage build and not directly in the target container.

tools/sgx-lkl-docker
```diff
@@ -44,22 +44,46 @@ function build() {
 FROM ubuntu:18.04
 RUN apt-get update && apt-get install -y ca-certificates
```
prp, line 45: Are you using a multi-stage build to get a subset of these certificates? Is a regular Ubuntu image the best place to get these certificates?

tools/sgx-lkl-docker
```diff
@@ -44,22 +44,46 @@ function build() {
...
 # See https://github.com/microsoft/Azure-DCAP-Client.
-ENV AZDCAP_CACHE=${TMPDIR}
+ENV AZDCAP_CACHE=\${TMPDIR}
+
+COPY --from=0 /prep /
+
+USER sgxlkl
+WORKDIR \${HOME}
 
 ENTRYPOINT ["/opt/sgx-lkl/lib/external/loader", "/opt/sgx-lkl/bin/sgx-lkl-run-oe",\
```
prp: What is "/opt/sgx-lkl/lib/external/loader"?

tests/containers/cc/Makefile
```diff
@@ -11,14 +11,18 @@ CC=python-helloworld-cc
...
+CC_DOCKER_ARGS_COMMON=\
+	--rm \
+	--user $(shell id -u) \
+	--device /dev/net/tun --network host \
 	-v $(SGXLKL_PREFIX):/opt/sgx-lkl \
 	-e SGXLKL_VERBOSE=1 -e SGXLKL_KERNEL_VERBOSE=1 -e SGXLKL_TAP=sgxlkl_tap0 \
 	-e AZDCAP_DEBUG_LOG_LEVEL=INFO
 
+CC_DOCKER_ARGS_HW=\
+	--device /dev/sgx
```
prp, line 24: After creating a CC, how will a user learn about these parameters for docker run? Is this just documented in README.md? In addition, do we have `sgx-lkl-docker` or `sgx-lkl-cfg` output these after creating the CC?

prp (commented):
Thanks, left some comments/questions.

tools/sgx-lkl-docker
letmaik: Which part do you think is not robust here? The contents of `/etc/passwd` and `/etc/group` are standard and Docker itself parses these files if you use the name form of `-u user`.
The reason I'm creating the files/folders manually is that I want them to be minimal and independent of the first stage base image (Ubuntu).

tools/sgx-lkl-docker
letmaik: I'm using a multi-stage build because the second stage starts from scratch with empty base image and hence can't invoke any commands. The certificates are not a subset, it's all of them. I don't see a reason why the Ubuntu image is not a good source for getting the certificates, since that's what most containers would generally rely on.

tools/sgx-lkl-docker
letmaik: It's a symlink to the dynamic loader (`/opt/sgx-lkl/lib/external/`ld-linux-x86-64.so.2`). If you don't use the loader to run the executable then it would use the absolute loader path embedded in the executable and that would prevent us from mountain into a fixed location at `/opt/sgx-lkl`.

tests/containers/cc/Makefile
letmaik: This is documented in `docs/Docker.md` and as mentioned in that PR there will be follow-up PRs to integrate this into the overall documentation.

tools/sgx-lkl-docker
letmaik: I've seen it elsewhere but there's not really a convention. It's likely that we change the uid later to match whatever the k8s plugin expects when bringing in the TAP device.

tools/sgx-lkl-docker
prp: Perhaps instead of `loader`, it should have a name that describes more what it is. How about `ld-linux-x86-64`?

tools/sgx-lkl-docker
prp: I'm worried when files like this are edited directly. For example, what about if there is a clash with uid 999?

letmaik: Normal UIDs start at 1000. And in a standard Ubuntu base image there is only uid 0 anyway.

tools/sgx-lkl-docker
letmaik: This is not part of this PR, feel free to open an issue if you think it should be changed.

prp: It would be trivial for you to change it here :-)

prp (approved):

tools/sgx-lkl-docker
letmaik: You won.

prp: It's not about winning -- it's all about making the project better ;-)

prp (approved):
LGTM

PR #482, opened 2020-06-17
LTP: Fix for setpgid01

ashokchandran:
Issue: Test sets pgid for a process and retrieves it and checks. But it has a caveat that when pgid passed is 0 then it sets pgid to process id of calling process. Curret ltp code was not taking care of that.
Solution: Fix the test suite to check if pgid passed to setpgid is 0 then while retrieving pgid compare returned value with process id of calling process.


hukoyu (approved):
LGTM

PR #483, opened 2020-06-17
LTP: fix testcase futex_cmp_requeue01 issue

shaikshavali1:
The original test case creates waiter child
processes based on the values passed and wait
on the futex. The parent process creates waiter
child processes, wait until all child processes
go to suspend state and perform FUTEX_CMP_REQUEUE
& FUTEX_WAIT operations to wake up the waiters.

This test case is failed because sgx-lkl supports
single process environment.
Below modifications are performed to the test case.
    1. Used pthreads instead of child processes
    2. There no mechanism to know the specific
       thread going to suspend state. Hence,
       added a sleep of 3 seconds to go all threads
       in waiting state.
    3. Subtest cases which are using 100 and 1000
       test cases are removed because test case is
       randomly failing with these number of threads.


hukoyu (approved):
LGTM

PR #484, opened 2020-06-17
LTP: fix issue futex_cmp_requeue02 issue

shaikshavali1:
This test case is passing in latest sgx-lkl
build. Hence, enabling this test case.


hukoyu (approved):
LGTM

PR #485, opened 2020-06-18
LTP: fix testcase pipe09 issue

shaikshavali1:
Original test case performs below operation.
1. Open a  pipe
2. Fork a child which writes to the pipe
3. Fork another child which writes a different character to the pipe
4. Have the parent read from the pipe
5. It should get the characters from both children.
Sgx-lkl environment supports single process environment.
The test case is modified to work in single process
without forking child processes .


vamsikrishna935 (approved):
LGTM @hukoyu

hukoyu (approved):
LGTM

PR #487, opened 2020-06-18
LTP: fix test case futex_futex_wait03 issue

shaikshavali1:
This test case invokes “tst_process_state_wait2” in thread.
Main issue is, this function indefinitely wait for process state
to be changed to “S” by monitoring the proc file
(/proc/[pid]/state). In sgx-lkl process state is not updating
properly in proc filesystem. Hence, the call to
“tst_process_state_wait2” is removed and added a sleep.


vamsikrishna935 (approved):
@hukoyu LGTM

hukoyu (approved):
LGTM

PR #488, opened 2020-06-18
Add sample that attacks memory

letmaik:
@prp For testing I used the standard params.conf file with `Debug=1`. I was slightly surprised that I couldn't read the memory, I was expecting that I need `Debug=0`. How does gdb do it? Is there a special protocol to "enter" the enclave for debugging purposes?


prp:
@letmaik yes, there are special instructions to read/write enclave memory: EDBGRD/EDBGWR

letmaik:
I see. I guess the sample should eventually be updated to use release builds/mode only, once we have those.

samples/basic/attack/read_secret.c
```diff
@@ -0,0 +1,33 @@
...
+    {
+        fprintf(
+            stderr,
+            "Could not read first line of file %s: %s\n",
+            HW_FILE,
+            strerror(errno));
+        exit(1);
+    }
+    printf("Ready to be attacked...\n");
+    sleep(30);
```
prp: Do you want to pause until the user presses enter instead of a timeout? This may be more user friendly.

prp (approved):
LGTM, left one comment.

samples/basic/attack/read_secret.c
letmaik: Good idea, I always forget that SGX-LKL has a working stdin/console.

PR #489, opened 2020-06-18
LTP: Fix for setrlimit04

ashokchandran:
Issue: Test was failing while trying to set rlimit to a value and then run a process. As SGX has only one process this method cannot be used.
Solution: Set the rlimit using setrlimit. Then retrieve it using getrlimit and verify that it is properly set.


shaikshavali1 (approved):
@ashokchandran , LGTM

hukoyu (approved):
LGTM

PR #490, opened 2020-06-18
LTP: fix issue in test case futex_wake02

shaikshavali1:
This test case invokes a function “wait_for_threads()”
from main process. This function indefinitely wait for
the thread to change its state to “S” by monitoring
the proc file (/proc/[pid]/tasks) entry. In sgx-lkl
thread state is not properly getting updated in proc
filesystem. Hence, the call to “wait_for_threads” is
removed and added a sleep. Also updated the TC to work
in single process environment.


hukoyu (approved):
LGTM

PR #497, opened 2020-06-19
LTP: Fix for munmap03 testcase failure

rsrinivas01:
LTP: Fix for issue 10 , munmap03 test case failure..3 tests are run within munmap03. The return value of munmap is 0 but errorno is set correctly. So modified the check for validation to include TERRNO also.


rsrinivas01 (commented):
@hukoyu , @shaikshavali1 ...Please review the changes.

hukoyu (approved):
LGTM

Issue #491, opened 2020-06-19
Host network completely unreachable


hukoyu:
Trying to run tests/basic/network
It fails for both run-hw and run-sw
with error "Could not resolve host"

```
[[  SGX-LKL ]] lkl_mount_disks(): Set working directory: /
[[  SGX-LKL ]] libc_start_main_stage2(): Calling app main: /usr/bin/curl
* Could not resolve host: www.microsoft.com
* Closing connection 0
curl: (6) Could not resolve host: www.microsoft.com
[[  SGX-LKL ]] lkl_terminate(): terminating LKL (exit_status=6)
[[  SGX-LKL ]] lkl_termination_thread(): termination thread unblocked
[[  SGX-LKL ]] lkl_termination_thread(): calling lkl_sys_chdir(/)
[[  SGX-LKL ]] lkl_termination_thread(): calling lkl_umount_timeout(/)
[    5.124483] EXT4-fs (vda): re-mounted. Opts: (null)
[[  SGX-LKL ]] lkl_termination_thread(): calling lkl_virtio_netdev_remove()
Makefile:31: recipe for target 'run-sw' failed
make: *** [run-sw] Error 141
```

hukoyu:
cc @prp @davidchisnall @jxyang @letmaik

hukoyu:
I added www.microsoft.com to /etc/hosts file with ip
Now it is failing with error:

```
run-hw:
[[  SGX-LKL ]] lkl_mount_disks(): Set working directory: /
[[  SGX-LKL ]] libc_start_main_stage2(): Calling app main: /usr/bin/curl
*   Trying 88.221.41.6:443...
* TCP_NODELAY set
* connect to 88.221.41.6 port 443 failed: Host is unreachable
* Failed to connect to www.microsoft.com port 443: Host is unreachable
* Closing connection 0
curl: (7) Failed to connect to www.microsoft.com port 443: Host is unreachable
[[  SGX-LKL ]] lkl_terminate(): terminating LKL (exit_status=7)
[[  SGX-LKL ]] lkl_termination_thread(): termination thread unblocked
[[  SGX-LKL ]] lkl_termination_thread(): calling lkl_sys_chdir(/)
[[  SGX-LKL ]] lkl_termination_thread(): calling lkl_umount_timeout(/)
[    3.292266] EXT4-fs (vda): re-mounted. Opts: (null)
[...]

Issue #375, opened 2020-06-03
virtio_process_one() ignores error during queueing


letmaik:
https://github.com/lsds/sgx-lkl/blob/a674a8f6c6479fbfcb8578aded364b188c6b933e/src/host_interface/virtio.c#L233-L256

`enqueue()` returns a negative number if the request has not been queued and needs to be retried later.

Also, `virtio_process_one()` is supposed to return a status which actually gets checked in `virtio_process_queue`. That is probably meant to be the status from `enqueue()`.

prp:
Addressed by #501

PR #501, opened 2020-06-20
Fix broken networking due to virtio queue bug

prp:
Addresses https://github.com/lsds/sgx-lkl/issues/491 and https://github.com/lsds/sgx-lkl/issues/375.

### DIFF
--- a/src/host_interface/virtio.c
+++ b/src/host_interface/virtio.c
@@ -252,9 +252,8 @@ static int virtio_process_one(struct virtio_dev* dev, int qidx)
         desc = get_next_desc(q, desc, &idx);
     } while (desc && req->buf_count < VIRTIO_REQ_MAX_BUFS);
 
-    dev->ops->enqueue(dev, qidx, req);
-    // TODO https://github.com/lsds/sgx-lkl/issues/375
-    return 0;
+    // Return result of enqueue operation
+    return dev->ops->enqueue(dev, qidx, req);
 }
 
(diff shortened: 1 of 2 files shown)


hukoyu (approved):
Pipeline passed. Also locally verified that this PR fixes issue https://github.com/lsds/sgx-lkl/issues/491

src/host_interface/virtio.c
```diff
@@ -252,9 +252,8 @@ static int virtio_process_one(struct virtio_dev* dev, int qidx)
         desc = get_next_desc(q, desc, &idx);
     } while (desc && req->buf_count < VIRTIO_REQ_MAX_BUFS);
 
-    dev->ops->enqueue(dev, qidx, req);
```
letmaik, line 255: What's the return value when return is missing? Some random number?

prp: Yes, in that case, the return value is undefined (unless the function is `main()`, which returns 0).

davidchisnall: In theory, it's undefined (and for a while clang would emit `ud2` at the end of any function that returned a value and didn't have a return statement, but people complained).  In practice, it will be whatever was left in the register.  The compiler will likely have tail-called the enqueue function, so the value that it stored in the return register would be seen by the caller of this function, which happened to be what we wanted.

prp: Right, but interestingly, it didn't happen each time, which is what led to the bug.

PR #503, opened 2020-06-22
LTP: Fix for getdents01

ashokchandran:
Issue : The test was not using 64 bit API as it needs -l option. So it was using 32 bit API and string was not matching.
Solution: Set the flag to use 64 bit API


hukoyu (approved):
LGTM

PR #504, opened 2020-06-22
Distclean LKL to pick up configuration changes

prp:
This change ensures that `make clean/distclean` also performs a `make distclean` on LKL. Otherwise LKL configuration changes (by overriding `defconfig`) are not picked up.


davidchisnall (approved):
Thank you!

PR #506, opened 2020-06-22
Remove unnecessary attestation report. Fixes #470.

### LINKED ISSUE #470: Support for non-FLC SGX1 machines broken
**Summary:** Currently the support for SGX1 machines that do not support FLC is broken.

**Problem:** We compile OE with `HAS_QUOTE_PROVIDER=ON`, which is only supported on machines with FLC support/DCAP attestation. SGX-LKL then tries to obtain an attestation report as a test (even when running without attestation), which consequently fails.

**Solution:** Instead, we should check if the platform supports FLC and change how we build OE depending on that. The user should be notified that this means that attestation or a `RELEASE` build is unavailable on their platform.

wintersteiger:


prp (approved):
LGTM

Issue #297, opened 2020-05-22
[Tests] lkl_access_ok() should return failure on test invalid access, not fatal exit.


vamsikrishna935:
LTP tests expects EFAULT error returned when trying to access an invalid address. 
sgx-lkl doesn't return EFAULT but ABORTs.

Sample LTP tests:
https://github.com/lsds/sgx-lkl/pull/550
listxattr01,02 

```
2020-05-22T08:36:47.954418Z [(H)ERROR] tid(0x7f682a12e700) | :OE_ENCLAVE_ABORTING [/home/prashant/openenclave/host/calls.c:oe_call_enclave_function_by_table_id:91]
[   SGX-LKL  ] ethread (1: 19) [   SGX-LKL  ] FAIL: sgxlkl_ethread_init() failed (id=1 result=19 (OE_ENCLAVE_ABORTING))
```

in umount02 when it test umount with null, it causes enclave to abort. Normally application is expected to get EFAULT. But the back trace is showing that in lkl it calls oe_abort when address is invalid.
```
#0 lkl_access_ok (addr=0, size=1) at lkl/setup.c:1499
#1 0x00007fff0010aaff in __lkl_access_ok (size=, addr=) at ./arch/lkl/include/asm/uaccess.h:14
#2 strncpy_from_user (count=, src=, dst=) at ./include/asm-generic/uaccess.h:233
#3 getname_flags (filename=0x0, flags=0, empty=0x0) at fs/namei.c:149
#4 0x00007fff0010ad2d in getname (filename=) at fs/namei.c:209
#5 0x00007fff0010da1d in user_path_mountpoint_at (dfd=0, name=, flags=1049890816, path=0x7fff3e941020)
at fs/namei.c:2745
#6 0x00007fff0012003d in ksys_umount (name=0x7fff3e941000 " \020\224>\377\177", flags=0) at fs/namespace.c:1680
#7 0x00007fff00120479 in __do_sys_umount (flags=, name=) at fs/namespace.c:1706
#8 __se_sys_umount (name=, flags=) at fs/namespace.c:1704
#9 0x00007fff00088076 in run_syscall (params=, no=) at arch/lkl/kernel/syscalls.c:44
#10 lkl_syscall (no=140734243278880, params=0x0) at arch/lkl/kernel/syscalls.c:183
#11 0x00007fff0054901a in __filter_syscall2 (n=39, a1=0, a2=0) at ./src/internal/syscall.h:161
[...]

PR #507, opened 2020-06-23
LTP: fix test case sendmsg01 issue

shaikshavali1:
In original test case the main process create a child process
to write/read to/from sockets infinitely. In sgx-lkl supports single
process environment. The test case is modified to use a child

some of the sub test case designed to test generation of EFAULT

Github issue297 (https://github.com/lsds/sgx-lkl/issues/297)
issue297 is fixed.

This test case is using "system" syscall causing
"bad count while changing owner". This is not fully
supported in sgxlkl. Removed the code calling "system"
call.


ashokchandran:
Reviewed . LGTM. Approved

hukoyu (approved):
LGTM

PR #509, opened 2020-06-23
Defer unmapping the thread stack.

davidchisnall:
When a thread is created with clone and detached, the last thing that
libc does before the thread exits is unmap the stack.  This works on
most Linux architectures because the last bit of assembly code can put
the thread exit code in a register and perform an exit system call.  It
does not work with LKL, because LKL does not maintain separate kernel
and userspace threads and because LKL passes arguments in memory (which,
in turn, means that the caller typically allocates them on the stack).

To work around this, we check on `munmap` calls whether the region being
unmapped contains the stack pointer.  If it does, then we defer
unmapping the stack until lthreads cleans up the thread (at which point,
nothing is using it and the unmap operation runs on a different thread's
stack).

### DIFF
--- a/src/enclave/enclave_mem.c
+++ b/src/enclave/enclave_mem.c
@@ -12,6 +12,9 @@
 
 #include "shared/bitops.h"
 
+#include <enclave/lthread.h>
+
+#include "enclave/lthread_int.h"
 #include "enclave/enclave_mem.h"
 #include "enclave/enclave_util.h"
 #include "enclave/ticketlock.h"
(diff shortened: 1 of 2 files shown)


src/enclave/enclave_mem.c
```diff
@@ -164,6 +167,22 @@ void* syscall_SYS_mremap(
...
 int syscall_SYS_munmap(void* addr, size_t length)
 {
+    // During thread teardown, libc unmaps the stack of the current thread
+    // before doing an exit system call.  This works on a conventional system
+    // because it's possible to do an exit system call without a stack.  With
+    // LKL, the kernel and userspace share a stack and so any system call needs
+    // a stack.  We work around this by deferring any attempt to unmap the
+    // current stack.
+    register void *rsp __asm__ ("rsp");
+    oe_host_printf("Unmapping: %p + %ld\n", addr, length);
```
SeanTAllen: this is a compilation error for me as oe_host_printf is an implicit declaration.

davidchisnall: Ooops, that was a debugging line I accidentally left in.

prp (approved):
LGTM

PR #512, opened 2020-06-23
LTP: fix test case execve02 issue

shaikshavali1:
This test case is failed because, copying of resource
file “execve_child” to the current directory is failed.
Below modifications are performed to fix error.
1. Complete bin file path is mentioned in TEST_APP.
2. Removed the resource file configuration.
3. Modified to work as a single process.


hukoyu (approved):
LGTM

PR #513, opened 2020-06-23
Avoid Enclave MAC address Collision

Francis-Liu:
Existing code fixes the MAC address to be used in LKL. This generate MAC address conflicts when running multiple Enclaves on the same host, attaching their TAP interfaces to a `bridge` device, and letting them communicate with each other.

### DIFF
--- a/src/host_interface/virtio_netdev.c
+++ b/src/host_interface/virtio_netdev.c
@@ -29,6 +29,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <sys/mman.h>
+#include <sys/random.h>
 
 /* We always have 2 queues on a netdev: one for tx, one for rx. */
 #define RX_QUEUE_IDX 0
@@ -517,7 +518,21 @@ int netdev_init(sgxlkl_config_t* config)
 {
     void* netdev_vq_mem = NULL;
     struct virtio_net_dev* net_dev = NULL;
-    char mac[6] = {0xCA, 0xFE, 0x00, 0x00, 0x00, 0x01};
+    char mac[6];
+    // Generate a completely random MAC address
+    size_t b = 0;
+    while (b < sizeof(mac)) {
+        ssize_t ret = getrandom(&mac[b], sizeof(mac) - b, GRND_RANDOM);
+        if (ret < 0) {
+            sgxlkl_host_fail("%s: get random MAC address failed: %s\n", __func__, strerror(errno));
+            return -1;
+        }
+        b += ret;
+    }
+    // Set the locally administered bit:
+    mac[0] |= 2;
+    // Clear the multicast bit (give a unicast MAC address)
+    mac[0] &= 0xfe;
 
     size_t host_netdev_size = next_pow2(sizeof(struct virtio_net_dev));
     size_t netdev_vq_size = NUM_QUEUES * sizeof(struct virtq);


src/host_interface/virtio_netdev.c
```diff
@@ -518,6 +519,10 @@ int netdev_init(sgxlkl_config_t* config)
     void* netdev_vq_mem = NULL;
     struct virtio_net_dev* net_dev = NULL;
     char mac[6] = {0xCA, 0xFE, 0x00, 0x00, 0x00, 0x01};
+    srand(time(NULL));
+    for (int i = 3; i < 6; i++) {
+        mac[i] = rand() % 256;
+    }
```
prp: MAC addresses encode manufacturer identifiers. Does this follow the best practice on generating "random" MAC addresses, or are we clashing with an assigned manufacturer id? If this is safe, a comment explaining why would be useful.

prp (commented):
Thanks, I left a comment/question.

src/host_interface/virtio_netdev.c
Francis-Liu: @prp The first 3 octets CAFE00 doesn't match any OUI (Organizationally Unique Identifier). The only caveat is the last 2 bits in the 1st octet has to be 0b10, for b1=1 means locally administered versus globally, and b0=0 means unicast versus multicast. 0xCA follows this rule. If we keep Enclave to be running in NAT network, Enclave MAC address won't be visible outside host.

prp: Sounds good. Do you want to add a comment that explains the above? Thanks.

davidchisnall: `rand()` is a deterministic random number generator and `srand(time(NULL))` is the example that most crypto courses give you as how not to seed it.  An AKS cluster spinning up a bunch of enclaves at the same time and connecting them to a bridge has a very high chance of collisions.  

```suggestion
    // Generate a completely random MAC address
    for (size_t b = 0 ; b < sizeof(mac) ; b += getrandom(&mac[b], sizeof(mac)-b, GRND_RANDOM) {}
    // Set the locally administered bit:
    mac[0] |= 2;
    // Clear the multicast bit (give a unicast MAC address)
    mac[0] &= 0xfe;
```

Francis-Liu: @davidchisnall `getrandom()` needs access to the `/dev/random` device, which isn't available in a docker container. I could use Enclave IP address as the seed. As I mentioned, we only need to guarantee uniqueness within a Pod's NAT network instead of an AKS cluster network. Our init-container, which runs in each Pod to up network for app containers, would guarantee the uniqueness of Enclave IP address. And, if we want the MAC to be deterministic -- randomization isn't our focus which is collision avoiding, we can use the last 3 numbers in Enclave IP address as the last 3 octets of MAC address. For example, 10.0.1.1 --> CA-FE-00-00-01-01. What do you think?

Another point is I actually prefer using 0xCAFE00 for Enclave MAC OUI. When a developer sees this type of MAC address during debugging, he can easily tell where does it come from.

letmaik: @Francis-Liu Are you sure `/dev/random` is not available in a Docker container? That would break a lot of software, wouldn't it?

davidchisnall: From the `getrandom(2)` man page:

> Unlike  /dev/random  and  /dev/urandom, getrandom() does not involve the use of pathnames or file descriptors.  Thus, getrandom() can be useful in cases where chroot(2) makes /dev pathnames invisible, and where an application (e.g., a daemon during start-up) closes a file descriptor for one of these files that was opened by a library.

That said, I agree with @letmaik, if `/dev/random` is not available inside Docker containers, that's a big security hole for most microservices.

src/host_interface/virtio_netdev.c
```diff
@@ -29,6 +29,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <sys/mman.h>
+#include <time.h>
```
davidchisnall:
```suggestion
```

This is not needed if you're not doing `srand(time(NULL))` (which you shouldn't be doing).

src/host_interface/virtio_netdev.c
Francis-Liu: @letmaik @davidchisnall Thank you, I ran a test and verified `getrandom` actually works in a docker container. Moreover, both `dev/random` and `/dev/urandom` still exist in a non-privileged docker container -- I am happy to know that my assumption was wrong. With that said, I did not verify the situation as described by the `getrandom(2)` man page, that is when these device files were invisible. Next, I am going to verify this in Kubernetes.

I am curious how things work if you call `getrandom` within SGX-LKL. What is behind `/dev/random` device file within SGX-LKL? Is that the same device as on the host?

prp: No, it's entirely implemented inside the enclave by LKL, with the entropy pool initialised using `rdrand`: https://github.com/lsds/sgx-lkl/blob/c5c6da350e61f4fea8bafc14bb07131a765d92ad/src/lkl/setup.c#L1109

Francis-Liu: Thank you @prp. Does this line of code guarantees that calling `getrandom` within LKL would return hardware-generated random number, and its source of randomization is the hardware on the host? In other words, LKL `getrandom` is a real random number generator.

prp: I believe that the implicit source of randomness is going to be the `rdrand` instruction executed inside the enclave. This should be properly security-audited though.

davidchisnall (approved):

PR #516, opened 2020-06-24
LTP: fix test case setgid02 issue

shaikshavali1:
Test is failed to open /etc/passwd file with nobody user.
As test is not intended to test opening /etc/passwd.
Hence, moved the related code to setup function.
Git hub issue 224 “https://github.com/lsds/sgx-lkl/issues/224”
is raised.


ashokchandran:
LTGM. Approved.

hukoyu (approved):
LGTM

PR #505, opened 2020-06-22
Update OE submodule

jhand2:
OE submodule branch was rebased against latest master on June 22, 2020

PR #520, opened 2020-06-24
CI: Run tests with unbuffered output streams to avoid losing output

letmaik:
The issue in https://github.com/lsds/sgx-lkl/pull/505 could have been diagnosed instantly if the error message output by OE would not have been swallowed before crashing. Running the tests with unbuffered output streams avoids this.

Note that OE is sending errors to stdout unfortunately, otherwise it would not have been an issue since stderr is unbuffered by default.


prp (approved):
LGTM

PR #521, opened 2020-06-24
make_self_contained.sh: fail on pipe failures

letmaik:
Previously, the script would happily succeed even if some of the libraries weren't found. This is because the output of `lddtree` was sent into a pipe which swallowed the error code.


prp (approved):
LGTM

PR #522, opened 2020-06-24
Add sgx-lkl-gdb test

### LINKED ISSUE #442: Add unit test for sgx-lkl-gdb
This can be a crash test or something more clever.

letmaik:
This can be considered a crash test for sgx-lkl-gdb. It's also useful for diagnosing systemic issues like in #505 as it will dump a stack trace which can be looked at in CI logs.

Closes #442.


tests/basic/gdb/Makefile
```diff
@@ -0,0 +1,41 @@
...
+
+run: run-hw run-sw
+
+run-gdb: run-hw-gdb
+
+run-hw: ${SGXLKL_ROOTFS}
+	  $(SGXLKL_ENV) $(SGXLKL_GDB) -batch -ex "handle SIGILL nostop noprint" -ex run -ex bt --args $(SGXLKL_STARTER) $(SGXLKL_HW_PARAMS) $(SGXLKL_ROOTFS) $(PROG)
+
+run-sw: ${SGXLKL_ROOTFS}
+	  $(SGXLKL_ENV) $(SGXLKL_GDB) -batch -ex "handle SIGILL nostop noprint" -ex run -ex bt --args $(SGXLKL_STARTER) $(SGXLKL_SW_PARAMS) $(SGXLKL_ROOTFS) $(PROG)
```
prp: Can you document what this test is supposed to do? How do you distinguish success vs failure?

prp (approved):
LGTM, but this test needs some documentation.

tests/basic/gdb/Makefile
letmaik: I did some more tests and it turns out due to a [GDB bug](https://sourceware.org/bugzilla/show_bug.cgi?id=13000) fixed in 8.3 (available in Ubuntu >= 19.10) gdb will always exit with 0 in batch mode. This means this test will never fail.

I guess to make the test generally more useful (and which would work around the issue above) I could run gdb commands that simulate a normal developer, like setting a breakpoint in an in-enclave function, get a backtrace, dump a value, etc. and compare that to expected output. We'd have to find something that is as stable as possible and doesn't depend on compiler optimizations etc. otherwise the test would break eventually.

letmaik: I modified the test to set a breakpoint and check that it gets hit including the right function argument value. Can you check whether this is self-explanatory or it needs a comment?

prp (approved):
LGTM, thanks.

tests/basic/gdb/Makefile
prp: I think that's clear.

letmaik:
There seems to be an issue with the non-release build (`make DEBUG=false`) of SGX-LKL where sgx-lkl-gdb doesn't hit the breakpoint, even though the application code itself is always built with debug symbols and without optimizations. Is this expected? Is the gdb support in SGX-LKL disabled for all except debug builds?

EDIT: Yes, the code is guarded by #ifdefs in `sgxlkl_run_oe.c`.

letmaik:
I tried uncommenting the `#ifdef`d code in sgxlkl_run_oe.c:
```
//#ifdef DEBUG
/* Need to extract the enclave base address from OE for GDB */
typedef struct _oe_enclave_partial_t
{
    uint64_t magic;
    char* path;
    uint64_t addr;
} _oe_enclave_partial;

void __attribute__((noinline))
__gdb_hook_starter_ready(void* base_addr, int mode, char* libsgxlkl_path)
{
    __asm__ volatile("nop" : : "m"(base_addr), "m"(mode), "m"(libsgxlkl_path));
}
//#endif
```
and 
```
//#ifdef DEBUG
    /* Need base address for GDB to work */
    _oe_enclave_partial* oe_enclave_content = (_oe_enclave_partial*)oe_enclave;
    void* base_addr = (void*)oe_enclave_content->addr;
    __gdb_hook_starter_ready(base_addr, encl.mode, libsgxlkl);
//#endif
```

However gdb complains with:
```
Python Exception <class 'gdb.error'> No symbol "base_addr" in current context.: 

Thread 1 "sgx-lkl-run-oe" hit Breakpoint -3, 0x000000004000d900 in __gdb_hook_starter_ready ()
#0  0x000000004000d900 in __gdb_hook_starter_ready ()
#1  0x000000004000c482 in main ()
```

letmaik:
I tried adding `__attribute__((optimize("O0")))` to the `__gdb_hook_starter_ready` function but same error. I guess the issue is just that the debug symbols are missing.

prp:
Yes, only with `DEBUG=true` or `DEBUG=opt`, the debug symbols and gdb hook is included in `libsgxlkl.so`.

letmaik:
Hm, since this is code on the host side and it's quite unexpected that it doesn't work should we move the `__gdb_hook_starter_ready` function into a separate C file which we always compile with `-O0 -g`?

letmaik:
Looks like there is a lot more attached to that (see `sgx-lkl-gdb.py`) and I'm coming to the realization that it's not worth the effort trying to get this running in the builds without debug symbols. I wonder whether we should embed debug symbols for the non-release build as well.

prp:
> Looks like there is a lot more attached to that (see `sgx-lkl-gdb.py`) and I'm coming to the realization that it's not worth the effort trying to get this running in the builds without debug symbols. I wonder whether we should embed debug symbols for the non-release build as well.

Why do you want to have debug symbols in SGX-LKL for the NON-RELEASE build? Even without debug symbols in SGX-LKL, it should be possible to debug enclave applications with debug symbols (as long as we have the gdb starter hook inside of the enclave).

letmaik:
It's not just the starter hook. There's all kinds of dynamic things going on in sgx-lkl-gdb.py, like dealing with the lthreads. I'm not saying it's impossible but it would require quite a bit of investigation.

prp:
> It's not just the starter hook. There's all kinds of dynamic things going on in sgx-lkl-gdb.py, like dealing with the lthreads. I'm not saying it's impossible but it would require quite a bit of investigation.

It's the starter hook and the symbols exposed by the dynamic loader:
https://github.com/lsds/sgx-lkl-musl/blob/24624fb2cc09599b78d7aa223775c58fed99c44b/ldso/dynlink.c#L668

letmaik:
After building the starter hook with debug symbols the next error I got was that the symbol for `SW_DEBUG_MODE` wasn't found. It's another piece that's fixable, but I'm sure the next one would be `__scheduler_queue` etc.

letmaik:
I'm disabling this test for non-debug modes for now so that I can merge it. I'll open an issue.

PR #527, opened 2020-06-25
LTP: fix test case vhangup02 issue

shaikshavali1:
The “setsid” API returning the error because current
process is the group leader. Sgx-lkl supports single
process environment and current process is the group
leader. Hence, code related to fork is removed and
Error from “setsid” is ignored.


ashokchandran:
Reviewed. LGTM. Approved

hukoyu (approved):
LGTM

PR #539, opened 2020-06-25
LTP: fix test case alarm02 issue

shaikshavali1:
Alarm can return non zero if previous alarms are existing.
Here test suite itself has alarm running to check for timeouts.
In sgxlkl environment, the test suite is already modified
to run test framework and test in a single process
environment. Hence, in the test case “alarm” system call
returning non zero value and test case failed.
The alarm is reset in testcase before verification.


hukoyu (approved):
LGTM

PR #541, opened 2020-06-25
LTP: Fix for ptrace03

ashokchandran:
Issue: Test was failing as it requires child process for ptrace calls. 
Solution: Removed the wait that was causing test to hang and also disabled one test for PTRACE_TRACEME that requires child process.

### DIFF
--- a/tests/ltp/patches/ptrace03.patch
+++ b/tests/ltp/patches/ptrace03.patch
@@ -1,8 +1,7 @@
 Removed the wait calls as it is in single process context.
 PTRACE_TRACEME needs child process. Hence removing it from tests for sgxlkl
-
 diff --git a/testcases/kernel/syscalls/ptrace/ptrace03.c b/testcases/kernel/syscalls/ptrace/ptrace03.c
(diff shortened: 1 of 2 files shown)


tests/ltp/patches/ptrace03.patch
```diff
@@ -0,0 +1,131 @@
...
++		tst_resm(TINFO,"Running test %d",i);
++		/* since Linux 2.6.26, it's allowed to trace init,
++		   so just skip this test case */
++		if (i == 0 && tst_kvercmp(2, 6, 25) > 0) {
++			tst_resm(TINFO,
++				 "skipping this test as this kernel allows to trace init");
++			continue;
++		}
++
++
```
shaikshavali1: Remove additional line

shaikshavali1 (commented):
Please remove additional lines in code

tests/ltp/patches/ptrace03.patch
ashokchandran: done

hukoyu (approved):
LGTM

PR #543, opened 2020-06-25
Make failed exception handling more verbose

prp:
This PR generates a descriptive warning when an exception cannot be handled inside of the enclave.

This should help debug some of the exception failures when running LTP tests.

### DIFF
--- a/src/enclave/enclave_signal.c
+++ b/src/enclave/enclave_signal.c
@@ -17,10 +17,11 @@
 /* Mapping between OE and hardware exception */
 struct oe_hw_exception_map
 {
-    uint32_t oe_code; /* OE exception code */
-    int trapnr;       /* Hardware trap no  */
-    int signo;        /* Signal for trap   */
-    bool supported;   /* Enabled in SGX-LKL*/
+    uint32_t oe_code;  /* OE exception code */
+    int trapnr;        /* Hardware trap no  */
+    int signo;         /* Signal for trap   */
+    bool supported;    /* Enabled in SGX-LKL */
+    char* description; /* Description string */
 };
 
 /* Encapsulating the exception information to a datastructure.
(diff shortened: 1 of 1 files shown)


tests/ltp/ltp-batch2/ltp_disabled_tests.txt
```diff
@@ -800,9 +800,9 @@
 #/ltp/testcases/kernel/syscalls/sendfile/sendfile08
 #/ltp/testcases/kernel/syscalls/sendfile/sendfile09
 #/ltp/testcases/kernel/syscalls/sendmmsg/sendmmsg01
-#/ltp/testcases/kernel/syscalls/sendmsg/sendmsg01
+/ltp/testcases/kernel/syscalls/sendmsg/sendmsg01
```
hukoyu: Why are you disabling these test cases? This time these tests failed randomly, next time different tests can fail. I think we shouldn't disable them.

tests/ltp/ltp-batch1/ltp_disabled_tests.txt
```diff
@@ -261,10 +261,10 @@
 #/ltp/testcases/kernel/syscalls/fstatat/fstatat01
 #/ltp/testcases/kernel/syscalls/fstatfs/fstatfs01
 #/ltp/testcases/kernel/syscalls/fstatfs/fstatfs02
-#/ltp/testcases/kernel/syscalls/fsync/fsync01
-#/ltp/testcases/kernel/syscalls/fsync/fsync02
-#/ltp/testcases/kernel/syscalls/fsync/fsync03
-#/ltp/testcases/kernel/syscalls/fsync/fsync04
+/ltp/testcases/kernel/syscalls/fsync/fsync01
```
hukoyu: Why are we disabling fsync tests? We are loosing total syscall test coverage here. As we discussed we cannot use loop device file system and we use root file system to enable these tests. What solution do you recommend to run these  tests?

hukoyu (changes requested):
I don't think we should disable these tests. We are loosing test coverae here. Also due to some issues tests fail randomly. This time it was sendmsg and sendto but it doesn't mean they will fail next time. Shall we disable every randomly failed test due to issues we haven't figued out root cause? Also if you disable these tests, how shall we see the additional info you ar eadding in this PR?
cc @paulcallen @Pengpeng-Microsoft @mikbras

prp:
@hukoyu As I explained, these tests are fundamentally broken:
- `fsync` relies on block devices being created inside the enclave, which is not possible currently.
- `sendto`/`sendmsg` want to create a client/server setup inside the enclave. This requires multiple processes, which is also not possible.

These tests therefore do not test anything meaningful, and they should have never been merged in the first place.

paulcallen:
We need to stop disabling tests that blow holes in our testing. If they are flakey then we need to fix the tests. Disabling at this point is not acceptable unless we get consensus across the team that we have a contingency plan to re-enable it.

paulcallen (changes requested):
We need to stop disabling tests that blow holes in our testing. If they are flakey then we need to fix the tests. Disabling at this point is not acceptable unless we get consensus across the team that we have a contingency plan to re-enable it.

prp:
@paulcallen the LTP tests that I disabled were not flakey but incorrect. We need better QA and not add incorrectly patched LTP tests to our CI pipeline.

davidchisnall (approved):
Looks good to me, but it might be better to pull the test suite changes into a different PR. I agree that we shouldn’t be trying to run LTP tests that test features that will not work with SGX. I am a bit surprised that here we’re passing previously, that implies a need for better QA before enabling an LTP test.

prp:
> Looks good to me, but it might be better to pull the test suite changes into a different PR. I agree that we shouldn’t be trying to run LTP tests that test features that will not work with SGX. I am a bit surprised that here we’re passing previously, that implies a need for better QA before enabling an LTP test.

@davidchisnall I've moved the disabling of incorrect LTP tests to a separate PR https://github.com/lsds/sgx-lkl/pull/549.

tests/ltp/ltp-batch2/ltp_disabled_tests.txt
prp: I've provided an explanation, and moved the changes to the CI pipeline to a separate PR.

tests/ltp/ltp-batch1/ltp_disabled_tests.txt
prp: I've provided an explanation, and moved the changes to the CI pipeline to a separate PR.

hukoyu (approved):
DIsabling test moved to another PR. This PR looks good.

paulcallen (approved):

PR #547, opened 2020-06-26
fix: wireguard client peer configuration without mandatory endpoint

fnerdman:
Wireguard peers do not necessarily need an endpoint, as long as the other side (the enclave) provides an endpoint. I therefore removed the endpoint checks where necessary. This enables users to remotely attest and control enclaves behind NAT.


prp:
Thanks @lead4good. Would it be possible to provide this PR for the OE version of SGX-LKL in branch `oe_port`? We can then review it.

fnerdman:
See #563

PR #548, opened 2020-06-26
CI: Don't fail publish test results task on test failures

letmaik:
The Run Tests task already fails in this case (as the test runner script exits with 1) and this just creates more noise, in particular it adds one more entry per failed test suite to the failure list on GitHub:

grafik


prp (approved):
LGTM

PR #549, opened 2020-06-26
Disable incorrect sendto, sendmsg and fsync LTP tests

prp:
This PR disables LTP tests (sendto, sendmsg and fsync) that are patched incorrectly and therefore fail arbitrarily and/or don't increase test coverage. More specifically:

- the `fsync` test relies on block devices being created inside the enclave, which is not possible currently;
- the `sendto`/`sendmsg` tests need to create a client/server setup inside the enclave. This requires multiple processes, which is also not supported.

The current patches for these tests do not overcome the above problems correctly.

### DIFF
--- a/tests/ltp/ltp-batch1/ltp_disabled_tests.txt
+++ b/tests/ltp/ltp-batch1/ltp_disabled_tests.txt
@@ -261,10 +261,10 @@
 #/ltp/testcases/kernel/syscalls/fstatat/fstatat01
 #/ltp/testcases/kernel/syscalls/fstatfs/fstatfs01
 #/ltp/testcases/kernel/syscalls/fstatfs/fstatfs02
-#/ltp/testcases/kernel/syscalls/fsync/fsync01
-#/ltp/testcases/kernel/syscalls/fsync/fsync02
-#/ltp/testcases/kernel/syscalls/fsync/fsync03
-#/ltp/testcases/kernel/syscalls/fsync/fsync04
+/ltp/testcases/kernel/syscalls/fsync/fsync01
+/ltp/testcases/kernel/syscalls/fsync/fsync02
+/ltp/testcases/kernel/syscalls/fsync/fsync03
+/ltp/testcases/kernel/syscalls/fsync/fsync04
 #/ltp/testcases/kernel/syscalls/ftruncate/ftruncate01
 #/ltp/testcases/kernel/syscalls/ftruncate/ftruncate03
 /ltp/testcases/kernel/syscalls/ftruncate/ftruncate04
--- a/tests/ltp/ltp-batch2/ltp_disabled_tests.txt
+++ b/tests/ltp/ltp-batch2/ltp_disabled_tests.txt
@@ -264,7 +264,7 @@
 /ltp/testcases/kernel/syscalls/fsync/fsync01
 /ltp/testcases/kernel/syscalls/fsync/fsync02
 /ltp/testcases/kernel/syscalls/fsync/fsync03
-#/ltp/testcases/kernel/syscalls/fsync/fsync04
+/ltp/testcases/kernel/syscalls/fsync/fsync04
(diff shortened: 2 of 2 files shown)


davidchisnall:
I don't see where [https://github.com/linux-test-project/ltp/blob/master/testcases/kernel/syscalls/fsync/fsync01.c](this test), for example, is creating a block device.  

Agreed on the other two.  They do appear to use vfork if fork isn't available, so I'm hopeful that we can eventually enable them, but not any time soon.

prp:
> I don't see where [https://github.com/linux-test-project/ltp/blob/master/testcases/kernel/syscalls/fsync/fsync01.c](this test), for example, is creating a block device.

@davidchisnall for the `fsync` tests, the problem is with the patches that try to make the test work. For example, the patches try to mount the root block device (?), which is already mounted: https://github.com/lsds/sgx-lkl/blob/740b469f514a32519f3e7307290bd8e26609d592/tests/ltp/patches/ltp_fsync_fsync01.patch#L12 Some of the other patches then also write directly to the block device, which is definitely a bad idea. Since (I presume) the root file system is shared across LTP test invocations, any corruption may have unexpected effects on the LTP CI pipeline.

davidchisnall:
I see.  That patch is clearly nonsense and doesn't look as if it was tested.

davidchisnall (approved):

letmaik:
Let's add some comments in the disabled_tests.txt on why they are disabled and that the existing patch is incorrect.

prp:
> Let's add some comments in the disabled_tests.txt on why they are disabled and that the existing patch is incorrect.

Agreed, although this is probably a separate PR as we should survey all disabled LTP tests and review the associated patches. The current disabled_tests.txt file also doesn’t support comments.

davidchisnall:
What are all of the lines that start with `#`, if they are not comments?

davidchisnall:
@letmaik, I've opened a bug to track the ones that we should reenable with better patches.

prp:
> What are all of the lines that start with `#`, if they are not comments?

They are re-enabled tests, I think. The whole file is just processed with grep -vf.

SeanTAllen:
@prp is correct. a `#` means the test is active.

prp:
It can also be any other character or change to the line that prevents that line from matching.

SeanTAllen:
It was definitely confusing to me the first time I encountered it.

tests/ltp/ltp-batch1/ltp_disabled_tests.txt
```diff
@@ -261,10 +261,10 @@
 #/ltp/testcases/kernel/syscalls/fstatat/fstatat01
 #/ltp/testcases/kernel/syscalls/fstatfs/fstatfs01
 #/ltp/testcases/kernel/syscalls/fstatfs/fstatfs02
-#/ltp/testcases/kernel/syscalls/fsync/fsync01
-#/ltp/testcases/kernel/syscalls/fsync/fsync02
-#/ltp/testcases/kernel/syscalls/fsync/fsync03
-#/ltp/testcases/kernel/syscalls/fsync/fsync04
+/ltp/testcases/kernel/syscalls/fsync/fsync01
+/ltp/testcases/kernel/syscalls/fsync/fsync02
```
hukoyu, line 265: @prp @davidchisnall @SeanTAllen @letmaik @paulcallen @Pengpeng-Microsoft 
fsync02 and fsync03 is not even patched. They are original LTP tests. Why should we disable them? Do you mean they are not supported by sgx-lkl?

https://github.com/linux-test-project/ltp/blob/master/testcases/kernel/syscalls/fsync/fsync02.c
https://github.com/linux-test-project/ltp/blob/master/testcases/kernel/syscalls/fsync/fsync03.c

prp: I suggest that we look at all of the tests together and first decide how they test fsync and if this makes sense for SGX-LKL.

PR #550, opened 2020-06-26
LTP: fixed listxattr01,02 test cases

vamsikrishna935:
Patch Description: Tests were failing as no xattr support in loop device
So modofied the tests to use root filesystem with xattr support

### DIFF
--- a/tests/ltp/patches/ltp_listxattr_listxattr01_fix.patch
+++ b/tests/ltp/patches/ltp_listxattr_listxattr01_fix.patch
@@ -1,0 +1,49 @@
++ Patch Description: Tests were failing as no xattr support in loop device
++ So modofied the tests to use root filesystem with xattr support
+diff --git a/testcases/kernel/syscalls/listxattr/listxattr01.c b/testcases/kernel/syscalls/listxattr/listxattr01.c
+index 62198b2a7..ca3e70232 100644
+--- a/testcases/kernel/syscalls/listxattr/listxattr01.c
++++ b/testcases/kernel/syscalls/listxattr/listxattr01.c
+@@ -31,7 +31,9 @@
+ #define VALUE  "test"
+ #define VALUE_SIZE     (sizeof(VALUE) - 1)
+ #define KEY_SIZE    (sizeof(SECURITY_KEY1) - 1)
+-#define TESTFILE    "testfile"
++#define TESTFILE    "mntpoint/testfile"
++#define MNTPOINT    "mntpoint"
++#define DIR_MODE    (S_IRWXU | S_IRWXG | S_IROTH | S_IXOTH)
+
+ static int has_attribute(const char *list, int llen, const char *attr)
+ {
+@@ -64,17 +66,29 @@ static void verify_listxattr(void)
+ }
+
+ static void setup(void)
+-{
++{
++       rmdir(MNTPOINT);
++       SAFE_MKDIR(MNTPOINT, DIR_MODE);
++       SAFE_MOUNT("/dev/vda", MNTPOINT, "ext4", 0, NULL);
++
+        SAFE_TOUCH(TESTFILE, 0644, NULL);
+
+        SAFE_SETXATTR(TESTFILE, SECURITY_KEY1, VALUE, VALUE_SIZE, XATTR_CREATE);
+ }
+
++static void cleanup(void)
++{
++        remove(TESTFILE);
++        SAFE_UMOUNT(MNTPOINT);
++        SAFE_RMDIR(MNTPOINT);
++}
++
+ static struct tst_test test = {
+        .needs_tmpdir = 1,
+        .needs_root = 1,
+        .test_all = verify_listxattr,
+        .setup = setup,
++       .cleanup = cleanup,
+ };
+
+ #else
+
--- a/tests/ltp/patches/ltp_listxattr_listxattr02_fix.patch
+++ b/tests/ltp/patches/ltp_listxattr_listxattr02_fix.patch
@@ -31,7 +31,7 @@ index 98a0985b3..fd5c50ca0 100644
  {
 +       rmdir(MNTPOINT);
 +       SAFE_MKDIR(MNTPOINT, 0644);
-+       SAFE_MOUNT("/dev/vda", MNTPOINT, "ext4", 0, "user_xattr");
++       SAFE_MOUNT("/dev/vda", MNTPOINT, "ext4", 0, NULL);
 +
         SAFE_TOUCH(TESTFILE, 0644, NULL);
 
--- a/tests/ltp/patches/ltp_listxattr_listxattr02_fix.patch
+++ b/tests/ltp/patches/ltp_listxattr_listxattr02_fix.patch
@@ -1,0 +1,60 @@
++ Patch Description: Tests were failing as no xattr support in loop device
++ So modofied the tests to use root filesystem with xattr support
++ Issue 297: Sgx enclave is getting aborted while test is trying to access invalid address
++ https://github.com/lsds/sgx-lkl/issues/297
+diff --git a/testcases/kernel/syscalls/listxattr/listxattr02.c b/testcases/kernel/syscalls/listxattr/listxattr02.c
+index 98a0985b3..fd5c50ca0 100644
+--- a/testcases/kernel/syscalls/listxattr/listxattr02.c
++++ b/testcases/kernel/syscalls/listxattr/listxattr02.c
+@@ -36,7 +36,8 @@
+ #define SECURITY_KEY   "security.ltptest"
+ #define VALUE  "test"
+ #define VALUE_SIZE     (sizeof(VALUE) - 1)
+-#define TESTFILE    "testfile"
++#define TESTFILE    "mntpoint/testfile"
++#define MNTPOINT    "mntpoint"
+
+ char longpathname[PATH_MAX + 2];
+
+@@ -47,7 +48,7 @@ static struct test_case {
+ } tc[] = {
+        {TESTFILE, 1, ERANGE},
+        {"", 20, ENOENT},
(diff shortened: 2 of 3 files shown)


tests/ltp/patches/ltp_listxattr_listxattr01_fix.patch
```diff
@@ -0,0 +1,49 @@
...
+ {
+@@ -64,17 +66,29 @@ static void verify_listxattr(void)
+ }
+
+ static void setup(void)
+-{
++{
++       rmdir(MNTPOINT);
++       SAFE_MKDIR(MNTPOINT, DIR_MODE);
++       SAFE_MOUNT("/dev/vda", MNTPOINT, "ext4", 0, NULL);
```
prp, line 26: Can you explain what this line does? The root block device is already mounted.

tests/ltp/patches/ltp_listxattr_listxattr01_fix.patch
```diff
@@ -0,0 +1,49 @@
++ Patch Description: Tests were failing as no xattr support in loop device
++ So modofied the tests to use root filesystem with xattr support
```
prp, line 2: Are sure that the root file system has xattr support?

tests/ltp/patches/ltp_listxattr_listxattr02_fix.patch
```diff
@@ -0,0 +1,60 @@
...
++#define MNTPOINT    "mntpoint"
+
+ char longpathname[PATH_MAX + 2];
+
+@@ -47,7 +48,7 @@ static struct test_case {
+ } tc[] = {
+        {TESTFILE, 1, ERANGE},
+        {"", 20, ENOENT},
+-       {(char *)-1, 20, EFAULT},
++//     {(char *)-1, 20, EFAULT}, // TODO: Enable once git issue 297 is fixed
```
prp, line 24: Can you explain this change? I also don't understand issue #297

tests/ltp/patches/ltp_listxattr_listxattr01_fix.patch
```diff
@@ -0,0 +1,49 @@
...
++
+        SAFE_TOUCH(TESTFILE, 0644, NULL);
+
+        SAFE_SETXATTR(TESTFILE, SECURITY_KEY1, VALUE, VALUE_SIZE, XATTR_CREATE);
+ }
+
++static void cleanup(void)
++{
++        remove(TESTFILE);
++        SAFE_UMOUNT(MNTPOINT);
```
prp, line 36: What is being unmounted here?

prp (changes requested):
I left a bunch of questions/comments.

tests/ltp/patches/ltp_listxattr_listxattr02_fix.patch
hukoyu: LTP tests expects EFAULT error returned when trying to access an invalid address, it doesn't return EFAULT but aborts.

tests/ltp/patches/ltp_listxattr_listxattr01_fix.patch
hukoyu: Folder ./mountpoint is created
/dev/vda is mounted to ./mountpoint folder
test runs and then
 It is unmounted
mountpoint folder is removed.

tests/ltp/patches/ltp_listxattr_listxattr01_fix.patch
hukoyu: @prp actually this is a great opportunity to discuss about that and find the best solution for enablng this kind of tests. This approach is used for many test cases. @shaikshavali1 please you also get involved in this discussion.

@prp since we cannot use loop back device with sgx-lkl due to  memory limitations of sgx-lkl, here this mounting approach is used. We are mounting /dev/vda and running tests and then unmounting. This approah is used for 50+ enabled LTP tests. We want to understand if there is an issue with this approach and what is a better approach and we wll appriciate your expertiese and input here.

Here is Shaik's explanation about why we are using this approach:

Default size of LKL memory is set to 32M. This is kept low due to limited size of EPC (Enclave page cache) and to avoid page cache being swapped out.
 
In tests, where loop device is used ltp framework was causing a problem, as it tries to create a 256M in memory file system.
ltp framework perform below activity.
1. Create a .img of size 256M
2. Create a loop device and mount.
 
Also, it needs to run mkfs utility using unsupported system() system call. So, the work around on Peter's suggestion was to mount the root file system and test.
 
So, in some tests, where it mounts test fs, was replaced with root fs. eg: For attribute related tests the file system needs to be mounted with user_xattr option. But this was not set for "/". So, we had to mount root file system with the option and test, that's why we cannot use existing root filesystem and needed to mount /dev/vda with required options.

tests/ltp/patches/ltp_listxattr_listxattr01_fix.patch
hukoyu: root file system doesn't have it. That's why we are mounting /dev/vda with this option to be able to run the test.

tests/ltp/patches/ltp_listxattr_listxattr01_fix.patch
shaikshavali1: @prp, The root fs is not mounted with extended attributes option. In this test case, a test file is created on current directory and try to set extended attribute on test file. While setting the attribute below error is observed. 
"safe_macros.c:877: CONF: listxattr01.c:70: no xattr support in fs or mounted without user_xattr option"

To resolve this issue "/dev/vda" is mounted explicitly to a newly created directory ./mntpoint. After this change the setxattr is able to set the extended attribute of the file.

prp: `/dev/vda` is the block device with the root file system in the enclave. It is already mounted and being used. I do not understand how you can mount it again.

My suggestion was not to mount the root file system again (which is not possible), but instead to change the test to create test files directly on the root file system (if the test still works when it doesn't get to use a separate/fresh file system).

shaikshavali1: @prp, I will modify the test as per your suggestion and check the results.

shaikshavali1: @prp, I modified the test case to create a testfile  in "/" and "/ltp". In both the cases, the test case is passing.

prp: Does that mean that the root file system has xattr support? I thought we were mounting it without that option?

tests/ltp/patches/ltp_listxattr_listxattr02_fix.patch
prp: Ok, thanks. Can you clean up issue #297, so we can properly triage it? I can provide feedback what detail we require (e.g. a gdb trace) to fix this.

hukoyu: @prp cleaned https://github.com/lsds/sgx-lkl/issues/297

prp:
@vamsikrishna935 can you rebase this, so we get a cleaner CI run? Thanks.

vamsikrishna935:
> 
> 
> @vamsikrishna935 can you rebase this, so we get a cleaner CI run? Thanks.

@prp Done

tests/ltp/patches/ltp_listxattr_listxattr01_fix.patch
prp: The root file system shouldn't be unmounted here.

hukoyu: @prp  here we are unmounting /dev/vda. Do you mean this shoulnd't be unmounted?

prp: Indeed. It doesn't make sense for a userspace process to try to unmount the root file system while running.

hukoyu: @prp what about mounting? You said userspace process shouldn't unmount, but it is mounting at the beginning of the test. In the test to leave the image clean, we are unmounting it and deleting the mounterd folder. Actually after PR https://github.com/lsds/sgx-lkl/pull/562 merged, we will have fresh image file for each LTP test and then we can remove the unmount code here.

prp: See my comment above -- it makes no sense to mount `/dev/vda`. This is the root partition, and it has already been mounted by SGX-LKL during start-up.

shaikshavali1: > Indeed. It doesn't make sense for a userspace process to try to unmount the root file system while running.

@prp, in setup function of the test case "/dev/vda" is mounted to "./mnt_point". This code is counter part of un-mounting "./mnt_point". we will remove this un-mount code. 
As I mentioned in above discussion, the test case is passed by creating file in "/" and "/ltp".  As per initial investigation the root file system is not mounted with "xattr" option, I am further checking on why setxattr is passed on this file.

prp: As I said several times now, mounting `/dev/vda` makes no sense.

If the root file system doesn't support `xattr`, this test cannot be run, and it is indeed surprising why it would pass.

hukoyu:
@vamsikrishna935 can you remove all mounting/unmounting , mount folder create/delete related code from this PR and verify if it passes with a file in already mounted root files system (not /dev/vda) as @prp recommended.

hukoyu:
Dupe of https://github.com/lsds/sgx-lkl/pull/606

PR #553, opened 2020-06-26
Add sgx-lkl-cfg test

letmaik:
See the README for details on this test.

Also, note that I only included a reference file to check against for the enclave config, not the host config, since only the former is important (for attestation).

cc @wintersteiger


wintersteiger:
Awesome, thanks!

prp (approved):
LGTM

PR #561, opened 2020-06-28
Add pre-commit hook & fix shell script lint errors

letmaik:
This PR is copying some of the pre-commit scripts from OE. Currently I only enabled linting of shell scripts using `shellcheck` and fixed corresponding issues. There's also code formatting using `clang-format` but since this is much more disruptive I will enable that (and re-format code) once we agreed on the right strategic time for it.

cc @wintersteiger

### DIFF
--- a/src/sched/lthread.c
+++ b/src/sched/lthread.c
@@ -27,8 +27,6 @@
  *
  */
 
-// clang-format off
-
 #define WANT_REAL_ARCH_SYSCALLS
 
 #include <inttypes.h>
--- a/src/sched/lthread.c
+++ b/src/sched/lthread.c
@@ -27,6 +27,8 @@
  *
  */
 
+// clang-format off
+
 #define WANT_REAL_ARCH_SYSCALLS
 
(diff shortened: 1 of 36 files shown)


src/sched/lthread.c
```diff
@@ -27,6 +27,8 @@
  *
  */
 
+// clang-format off
```
prp: I don’t see the point of keeping different formatting here, as we are not tracking any upstream.

prp (approved):
Thanks, LGTM. I left one comment.

prp:
Oops, sgx-lkl-disk needs fixing.

letmaik:
> 
> 
> Oops, sgx-lkl-disk needs fixing.

Yeah, I'll mark this as draft until CI is passing.

src/sched/lthread.c
letmaik: OK makes sense, the license header mislead me.

wintersteiger:
This is great, many thanks @letmaik!

PR #562, opened 2020-06-29
Use fresh copy of image for each LTP test

hukoyu:
Currently we are using same image file for each LTP test.
If any test corrupts the image file (like mounting /dev/vda) it will affect following tests.
With this change we are creating a fresh copy of master/original image for each LTP test.

cc @prp @letmaik

### DIFF
--- a/tests/ltp/Makefile
+++ b/tests/ltp/Makefile
@@ -52,23 +52,23 @@ run-sw: $(ROOT_FS)
 	@${LTP_TEST_SCRIPT} run-sw
 
 run-hw-single: $(ROOT_FS)
-	@rm -fr $(ROOT_FS_FRESH_COPY)
-	@cp -f $(ROOT_FS) $(ROOT_FS_FRESH_COPY)
+	@rm -f $(ROOT_FS_FRESH_COPY)
+	@cp $(ROOT_FS) $(ROOT_FS_FRESH_COPY)
 	${SGXLKL_STARTER} --hw-debug $(ROOT_FS_FRESH_COPY) $(test)
 
 run-sw-single: $(ROOT_FS)
-	@rm -fr $(ROOT_FS_FRESH_COPY)
-	@cp -f $(ROOT_FS) $(ROOT_FS_FRESH_COPY)
+	@rm -f $(ROOT_FS_FRESH_COPY)
+	@cp $(ROOT_FS) $(ROOT_FS_FRESH_COPY)
 	${SGXLKL_STARTER} --sw-debug $(ROOT_FS_FRESH_COPY) $(test)
 
 run-hw-single-gdb: $(ROOT_FS)
-	@rm -fr $(ROOT_FS_FRESH_COPY)
-	@cp -f $(ROOT_FS) $(ROOT_FS_FRESH_COPY)
+	@rm -f $(ROOT_FS_FRESH_COPY)
+	@cp $(ROOT_FS) $(ROOT_FS_FRESH_COPY)
 	${SGXLKL_GDB} --args ${SGXLKL_STARTER} --hw-debug $(ROOT_FS_FRESH_COPY) $(test)
 
 run-sw-single-gdb: $(ROOT_FS)
-	@rm -fr $(ROOT_FS_FRESH_COPY)
-	@cp -f $(ROOT_FS) $(ROOT_FS_FRESH_COPY)
+	@rm -f $(ROOT_FS_FRESH_COPY)
+	@cp $(ROOT_FS) $(ROOT_FS_FRESH_COPY)
 	${SGXLKL_GDB} --args ${SGXLKL_STARTER} --sw-debug $(ROOT_FS_FRESH_COPY) $(test)
 
 clean:
--- a/tests/ltp/Makefile
+++ b/tests/ltp/Makefile
@@ -52,18 +52,22 @@ run-sw: $(ROOT_FS)
 	@${LTP_TEST_SCRIPT} run-sw
 
 run-hw-single: $(ROOT_FS)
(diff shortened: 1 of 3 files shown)


letmaik (approved):

prp (approved):
LGTM, thanks.

letmaik (changes requested):
CI is broken

letmaik:
@hukoyu There is some script error which has to be fixed, but more worrying, this script error doesn't make the job fail. Could you check that the LTP scripts use `set -e`?

hukoyu:
@letmaik can you be more specific? WHat is broken. I see pipeline is good. Built and ran the tests and test passed. 
I only added one line to copy image file. How come it broke pipeline? Also which script errors you ar mentioning?

hukoyu:
@letmaik There was one test failure and I reran the failed job and it passed. Are you talking about this? (You will see the failure in first Attempt)

letmaik:
Every single LTP test has this log:
```
[76/275] Running LTP test '/ltp/testcases/kernel/syscalls/fchdir/fchdir03' (Timeout = 60 seconds)...
cp: cannot stat 'sgxlkl-miniroot-fs.img.master': No such file or directory
SGXLKL_CMDLINE=mem=512m SGXLKL_VERBOSE=1 SGXLKL_KERNEL_VERBOSE=1 SGXLKL_TRACE_SIGNAL=1 timeout 60 /opt/sgx-lkl-debug/bin/sgx-lkl-run-oe --hw-debug sgxlkl-miniroot-fs.img /ltp/testcases/kernel/syscalls/fchdir/fchdir03 > "report/kernel-syscalls-fchdir-fchdir03-(debug)-(run-hw)-(8-ethreads).stdout.txt" 2>&1
/opt/sgx-lkl-debug/bin/sgx-lkl-run-oe --hw-debug sgxlkl-miniroot-fs.img /ltp/testcases/kernel/syscalls/fchdir/fchdir03: RETURNED EXIT CODE: 0
'kernel-syscalls-fchdir-fchdir03-(debug)-(run-hw)-(8-ethreads)' passed.
76, kernel-syscalls-fchdir-fchdir03-(debug)-(run-hw)-(8-ethreads), report/kernel-syscalls-fchdir-fchdir03-(debug)-(run-hw)-(8-ethreads).stdout.txt, Pass
INFO: Junit data recorded to 'report/TEST-kernel-syscalls-fchdir-fchdir03-(debug)-(run-hw)-(8-ethreads)-junit.xml'
```

Is the `cp: cannot stat 'sgxlkl-miniroot-fs.img.master': No such file or directory` line expected?

hukoyu:
@letmaik  this is definitely not expected. I only updated Makefile. I also have to update batch.mk which is used in ltp batch 1 and batch 2 folders. Tests were passing since batch.mk was still generating image with old expected name. Fixed this. Also added exit code check for image file copy.

hukoyu:
Copying over 1.5G image file was taking 8-20 seconds. If delete first and copy it is sub second. Submitted change for that.

hukoyu:
@letmaik  fixed the issues. Now it is working as expected. Can you review again? Thanks

tests/ltp/Makefile
```diff
@@ -51,19 +52,28 @@ run-sw: $(ROOT_FS)
 	@${LTP_TEST_SCRIPT} run-sw
 
 run-hw-single: $(ROOT_FS)
-	${SGXLKL_STARTER} --hw-debug $(ROOT_FS) $(test)
+	@rm -fr $(ROOT_FS_FRESH_COPY)
```
prp: Why specify `-r`? This is not a directory.

tests/ltp/Makefile
```diff
@@ -51,19 +52,28 @@ run-sw: $(ROOT_FS)
...
+	@cp -f $(ROOT_FS) $(ROOT_FS_FRESH_COPY)
```
prp: Since you've just removed the destination file, I don't understand why `-f` would be necessary.

prp (commented):
Thanks. I left some comments on the used flags.

tests/ltp/run_ltp_test.sh
```diff
@@ -71,6 +71,16 @@ for file in "${ltp_tests[@]}"; do
     # Start the test timer.
     JunitTestStarted "$test_name"
 
+    # Master copy of image is sgxlkl-miniroot-fs.img.master
+    # Before running each test copy a fresh copy of image
+    rm -fr sgxlkl-miniroot-fs.img
+    cp -f sgxlkl-miniroot-fs.img.master sgxlkl-miniroot-fs.img
+    cp_exit_code=$?
+    if [[ $cp_exit_code -ne 0 ]]; then
```
letmaik, line 79: This is OK for now but I think all bash scripts should be changed to run in `set -e` mode to avoid missing any errors. This should be a separate PR.

letmaik (approved):
Looks fine, but please fix the issues @prp mentioned.

SeanTAllen:
How is deleting the target file and then copying over a new one faster than just copying over a new one?

I don't see how anything would be done differently in either scenario except that when just copying, all the time that it takes would be perceived as being part of "copy time".

prp:
> How is deleting the target file and then copying over a new one faster than just copying over a new one?

I think that the copy then mostly reuses the data blocks that were just freed, instead of having to allocate new blocks and then unlinking the original file.

tests/ltp/Makefile
hukoyu: Removed.

tests/ltp/Makefile
hukoyu: Removed.

PR #565, opened 2020-06-29
Separate build-time scripts from user tools

letmaik:
I left the fsgsbase kernel module in tools/ as it will be removed anyway.


prp (approved):
LGTM. Is `sgx-lkl.ld` used for anything?

letmaik:
It's used in src/Makefile, but not sure if it has any effect.

letmaik:
Moving the lkl_*.c tools got a bit too fiddly. I'll leave them there for now until they get removed eventually.

PR #566, opened 2020-06-29
CI: add missing package python3-setuptools

letmaik:
This prevented installing sdist packages. It failed only in the publish stage because the VM environment is different (hosted agent vs scaleset pool).

Fixes https://dev.azure.com/sgx-lkl/sgx-lkl/_build/results?buildId=1007&view=results.


prp (approved):
LGTM

PR #567, opened 2020-06-29
Add sgx-lkl-cfg shell wrapper script

letmaik:
I will soon introduce Python script auto-formatting and for that I need all Python scripts to have a .py extension. The sgx-lkl-cfg script didn't have that.

Note that I didn't change the content of the Python script. I don't know why GitHub doesn't detect it as a rename.


prp:
Hmm, are we only adding this complexity, just to have some git auto-formatting support? The auto-formatting should adapt to the current codebase and not the other way around. Will we have to wrap all Python tools in the future?

letmaik:
Most tooling relies on file extensions, otherwise you have to specify those files manually. Shell scripts are an exception and the scripts from OE (which I merged earlier) analyse file content to determine whether something is a shell script or not.
The tooling of SGX-LKL is a bit funny anyway. It's not portable and it's not written in a single language. It would be better if there was a proper "SDK" which is separately packaged and which is written in a single language and portable, e.g. in Python, and published to a package manager. 

Long story short, I see your concern, and I will abandon this PR and include that file manually. It's not worth the trouble as things will change anyway in the future.

Issue #215, opened 2020-05-06
Improve logging infrastructure


letmaik:
Before I go into issues and how to improve things, I first want to document the status quo:

## Host-side

Option | Access | Notes
--|--|--
`SGXLKL_TRACE_HOST_SYSCALL=1` &<br> `"trace_host_syscall": true` | `struct sgxlkl_config_elem[]` | **dead code, not used**
`SGXLKL_VERBOSE=1` &<br> `"verbose": true` | `struct sgxlkl_config_elem[]` &<br> `sgxlkl_host_verbose(..)` &<br> `sgxlkl_host_verbose_raw(..)` | 
N/A | `sgxlkl_host_info(..)` |  always enabled
N/A | `sgxlkl_host_warn(..)` | always enabled
N/A | `sgxlkl_host_err(..)` | always enabled
N/A | `sgxlkl_host_fail(..)` | always enabled

## Enclave-side

Option | Access | Notes
--|--|--
`SGXLKL_TRACE_THREAD=1` | `getenv_bool("SGXLKL_TRACE_THREAD", 0)` & <br> `extern int sgxlkl_trace_thread` & <br> `SGXLKL_TRACE_THREAD(..)` | `int sgxlkl_trace_thread` defined in `setup.c`
`SGXLKL_TRACE_SIGNAL=1` | `getenv_bool("SGXLKL_TRACE_SIGNAL", 0)` &<br> `extern int sgxlkl_trace_signal` &<br> `SGXLKL_TRACE_SIGNAL(..)` | `int sgxlkl_trace_thread` defined in `setup.c`
`SGXLKL_TRACE_MMAP=1` | `getenv_bool("SGXLKL_TRACE_MMAP", 0)` &<br> `extern int sgxlkl_trace_mmap` &<br> `SGXLKL_TRACE_MMAP(..)` | `int sgxlkl_trace_thread` defined in `setup.c`
[...]

PR #570, opened 2020-06-29
Remove libc dependencies in enclave initialisation + add instrumentation

### LINKED ISSUE #564: [Tests] Rare segfault immediately after enclave_init
The following has happened in CI for kernel-syscalls-times-times01-(nonrelease)-(run-sw)-(8-ethreads):
```
[   SGX-LKL  ] SGX-LKL (OE) 0.2.0-dev (3694a68) LKL 5.3.0 [NON-RELEASE build (-O3)] [SOFTWARE DEBUG]
[   SGX-LKL  ] Warning: SGXLKL_TRACE_SIGNAL ignored in non-debug mode.
[   SGX-LKL  ] nproc=8 ETHREADS=8 CMDLINE="mem=512m"
[   SGX-LKL  ] HW TLS support: conf->fsgsbase=0
[   SGX-LKL  ] Registering disk 0 (path='sgxlkl-miniroot-fs.img', mnt='/', [RW    ])
[   SGX-LKL  ] get_signed_libsgxlkl_path... result=/opt/sgx-lkl-nonrelease/bin/../lib/libsgxlkl.so.signed
[   SGX-LKL  ] oe_create_enclave... result=0 (OE_OK)
[[  SGX-LKL ]] WARN: Unhandled exception SIGSEGV (page fault) received (code=5 addr=0x160 opcode=0x8948 lkl_is_running()=0 ret=0)
[[  SGX-LKL ]] WARN: Unhandled exception SIGSEGV (page fault) received (code=5 addr=0x160 opcode=0x8948 lkl_is_running()=0 ret=0)
```

The above problem may also explain other issues immediately after enclave initialisation. I have not observed this for the DEBUG build (yet?).

prp:
This PR attempts to fix #564. It removes references to libc functions (`snprintf`, `memset`, ...) early in the enclave initialisation sequence, before libc has been initialised. Without libc `malloc` being initialised, such calls may fail. The PR also removes a dependency on shared/env.c. 

The PR doesn't appear to fully solve the problem reported in #564, which now happens rather rarely in CI pipeline runs. Therefore, this PR also adds more instrumentation that should help debug #564 from CI logs.


src/enclave/enclave_oe.c
```diff
@@ -24,30 +20,56 @@ int __sgx_init_enclave()
     return __libc_init_enclave(sgxlkl_enclave->argc, sgxlkl_enclave->argv);
 }
 
-void sgxlkl_enclave_show_attribute(const void* sgxlkl_enclave_base)
+#ifdef DEBUG
+static void _size_uint64_to_str(uint64_t size, char* buf, uint64_t len)
+{
+    int i = 0;
+    double bytes = size;
+    const char* units[] = {"B", "KB", "MB", "GB", "TB", "PB"};
```
davidchisnall:
```suggestion
    const char* units[] = {"B", "KiB", "MiB", "GiB", "TiB", "PiB", "EiB"};
```
2^64 B is 16 EiB, so we need EiB to support all possible values in callers of this.

davidchisnall (approved):

src/enclave/enclave_oe.c
prp: Done.

prp:
Hmm, there still seems to be (another?) bug causing segfaults here: https://dev.azure.com/sgx-lkl/sgx-lkl/_build/results?buildId=1082&view=ms.vss-test-web.build-test-results-tab&runId=17572&resultId=100041&paneView=attachments

I've added more verbose output to reproduce this in the CI pipeline.

davidchisnall:
Looks as if the extra tracing fixed the segfaults?

prp:
> Looks as if the extra tracing fixed the segfaults?

I only saw the segfault after running the CI pipeline several times, so I'll keep triggering it and see what I get. The new tracing will help pinpoint where in the initialisation sequence it segfaults.

prp:
Gotcha:
```
[   SGX-LKL  ] SGX-LKL (OE) 0.2.0-dev (3722355) LKL 5.3.0 [DEBUG build (-O0)] [HARDWARE DEBUG]
[   SGX-LKL  ] nproc=4 ETHREADS=8 CMDLINE="mem=512m"
[   SGX-LKL  ] HW TLS support: conf->fsgsbase=1
[   SGX-LKL  ] Registering disk 0 (path='sgxlkl-miniroot-fs.img', mnt='/', [RW    ])
[   SGX-LKL  ] get_signed_libsgxlkl_path... result=/opt/sgx-lkl-debug/bin/../lib/libsgxlkl.so.signed
[   SGX-LKL  ] oe_create_enclave... result=0 (OE_OK)
[[  SGX-LKL ]] sgxlkl_enclave_init(): enter
[[  SGX-LKL ]] _sgxlkl_enclave_show_attribute(): enclave base=0x7fcc00000000 size=2.0 GiB
[[  SGX-LKL ]] _sgxlkl_enclave_show_attribute(): enclave heap base=0x7fcc00b55000 size=1024.0 MiB end=0x7fcc40b55000
[[  SGX-LKL ]] sgxlkl_enclave_init(): calling _dlstart_c()
[[  SGX-LKL ]] _register_enclave_signal_handlers(): Registering OE exception handler...
[[  SGX-LKL ]] __libc_init_enclave(): enclave_mman_init() finished
[[  SGX-LKL ]] __libc_init_enclave(): init_clock_res() finished
[[  SGX-LKL ]] __libc_init_enclave(): newmpmcq() finished
[[  SGX-LKL ]] __libc_init_enclave(): __init_tls() finished
[[  SGX-LKL ]] __libc_init_enclave(): _lthread_sched_init() finished
[[  SGX-LKL ]] WARN: [[  SGX-LKL ]] lkl_start_init(): kernel command line: 'mem=512m console=hvc0'
[[  SGX-LKL ]] lkl_start_init(): lkl_start_kernel() called
Unhandled exception SIGSEGV (page fault) received (code=5 addr=0x7fcc005c7834 opcode=0x8948 lkl_is_running()=0 ret=0)
[[  SGX-LKL ]] WARN: Unhandled exception SIGSEGV (page fault) received (code=5 addr=0x7fcc005c7834 opcode=0x8948 lkl_is_running()=0 ret=0)
[[  SGX-LKL ]] WARN: Unhandled exception SIGSEGV (page fault) received (code=5 addr=0x7fcc005c7834 opcode=0x8948 lkl_is_running()=0 ret=0)
[[  SGX-LKL ]] WARN: Unhandled exception SIGSEGV (page fault) received (code=5 addr=0x7fcc005c7834 opcode=0x8948 lkl_is_running()=0 ret=0)
```

davidchisnall:
Mem=512m? I wonder if we’re sometimes seeing sufficient fragmentation that we can’t do that allocation? The host malloc function doesn’t check for failure in the enclave mmap, so may end up returning -1, which would lead to some exciting segfaults here.

prp:
> Mem=512m? I wonder if we’re sometimes seeing sufficient fragmentation that we can’t do that allocation? The host malloc function doesn’t check for failure in the enclave mmap, so may end up returning -1, which would lead to some exciting segfaults here.

Hmm, that wouldn't explain this segfault, as it happens before we have called the bootmem initialisation that allocates the LKL  kernel memory.

That said, I agree that we shouldn't be running the LTP tests with a different amount of LKL memory. We should raise a separate PR to change this.

prp:
@wintersteiger could you review this? I would like to merge this PR, as it gives more detailed output when segfaults happen during the CI pipeline run.

wintersteiger:
Sure, no problem.

src/enclave/enclave_event_channel.c
```diff
@@ -145,18 +150,36 @@ void initialize_enclave_event_channel(
...
     evt_chn_lock = (struct ticketlock**)oe_calloc(
         evt_channel_num, sizeof(struct ticketlock*));
+    if (!evt_chn_lock)
+    {
+        sgxlkl_fail("Could not allocate memory for evt_chn_lock\n");
+    }
 
     vio_tasks =
         (struct lthread**)oe_calloc(evt_channel_num, sizeof(struct lthread*));
+    if (!vio_tasks)
```
davidchisnall: This pattern is repeated a lot.  Is it worth adding a helper along the lines of:
```c
void *oe_alloc_or_die(size_t size, size_t count, const char *fail_msg)
{
  void *ptr = oe_calloc(size, count);
  if (ptr == NULL)
  {
    sgxlkl_fail(fail_msg);
  }
  return ptr;
}
```

prp: Good idea. Where shall we put this?

src/enclave/enclave_event_channel.c
```diff
@@ -145,18 +150,36 @@ void initialize_enclave_event_channel(
 
     evt_chn_lock = (struct ticketlock**)oe_calloc(
         evt_channel_num, sizeof(struct ticketlock*));
+    if (!evt_chn_lock)
+    {
+        sgxlkl_fail("Could not allocate memory for evt_chn_lock\n");
```
wintersteiger: I wonder if we should add a checked-malloc macro. At least in src/enclave we always abort anyways and it would make the code more compact.

src/enclave/enclave_event_channel.c
```diff
@@ -145,18 +150,36 @@ void initialize_enclave_event_channel(
...
     for (int i = 0; i < evt_channel_num; i++)
     {
         evt_chn_lock[i] =
             (struct ticketlock*)oe_calloc(1, sizeof(struct ticketlock));
+        if (!evt_chn_lock[i])
+        {
+            sgxlkl_fail("Could not allocate memory for evt_chn_lock[%i]\n", i);
+        }
+
         memset(evt_chn_lock[i], 0, sizeof(struct ticketlock));
```
wintersteiger: This is already `calloc`'d, I don't think the `memset` is necessary.

src/enclave/enclave_event_channel.c
prp: Yeah, David suggested the same. Where would we put it, as we don't want to edit the OE headers?

src/enclave/enclave_init.c
```diff
@@ -231,16 +227,21 @@ int __libc_init_enclave(int argc, char** argv)
     libc.user_tls_enabled =
         sgxlkl_enclave->mode == SW_DEBUG_MODE ? 1 : sgxlkl_enclave->fsgsbase;
 
+    SGXLKL_VERBOSE("calling init_sysconf()\n");
     init_sysconf(
         sgxlkl_enclave->sysconf_nproc_conf, sgxlkl_enclave->sysconf_nproc_onln);
+
+    SGXLKL_VERBOSE("calling init_clock_res()\n");
```
wintersteiger: This type of output is super verbose, maybe we want a new flag or a trace for this. Maybe someone could look into extracting stack traces on error?

prp: I think that the whole tracing needs to be rethought (we have #215 for this), but I don't want to do this as part of this PR. If we catch some of the initialisation segfaults, I can do another PR that removes some of these messages again.

Having stack traces would be wonderful, but getting these to work with symbols would require some work because of the custom way how we load them.

src/include/openenclave/internal/utils.h
```diff
@@ -0,0 +1,233 @@
+// Copyright (c) Open Enclave SDK contributors.
```
wintersteiger: If we can avoid internal OE headers, we should. I can't easily see why we need it here. If it's just for the round-up to the next power of 2, I think we should re-implement that ourselves.

src/sched/lthread.c
```diff
@@ -1079,3 +1079,23 @@ void lthread_set_expired(struct lthread* lt)
...
 }
+
+#ifdef DEBUG
+void lthread_dump_all_threads(void)
+{
+    struct lthread_queue* lt_queue = __active_lthreads;
+
+    SGXLKL_VERBOSE("Dumping all lthreads:\n");
+
+    for (int i = 1; lt_queue; i++)
```
wintersteiger, line 1090: The `1` here is probably correct, but I'm missing the context to say so for sure.

wintersteiger (approved):
I think this is fine, I left some optional comments.

src/enclave/enclave_event_channel.c
wintersteiger: +1, I just said the same in my review. I was thinking along the lines of `#define CHECKED_CALLOC(X,Y) if ((X = oe_calloc(1, (Y)) == NULL) oe_abort();`

src/enclave/enclave_init.c
wintersteiger: Yeah, you're right, that should be part of the general tracing changes. 

I think for the kernel and below, `oe_print_backtrace()` should work, which may help with some of our CI debugging. But yeah, correctly loading symbols from application code is likely a lot of effort.

src/enclave/enclave_event_channel.c
wintersteiger: Maybe `enclave_util.h`?

src/enclave/enclave_init.c
prp: I tried it a while back, and it didn't work even for libsgxlkl.so symbols.

wintersteiger: I see! I might have a few spare minutes over the weekend, maybe I'll take another look at it.

src/enclave/enclave_event_channel.c
prp: Done.

src/enclave/enclave_event_channel.c
prp: Done.

src/include/openenclave/internal/utils.h
prp: Done.

src/sched/lthread.c
prp: We use this to prefix the printed lines, and this can start from 1 when counting the threads.

src/enclave/enclave_event_channel.c
prp: Done.

PR #571, opened 2020-06-29
Adding env variable mappings for some ADO stored secrets to be consumed by tests

hukoyu:
We need some secrets like MAA attestation credentials for some scenario tests.
These secrets are defined on ADO and stored encrypted. With this PR we are mapping these secrets to some environment variables on the running VM so that can be consumed by scenario tests. These credentials will be needed for several tests that will be added with soon coming PRs.

### DIFF
--- a/.azure-pipelines/template.yml
+++ b/.azure-pipelines/template.yml
@@ -148,6 +148,11 @@ stages:
                 run_mode: "run-${{ run_mode }}"
                 SGXLKL_PREFIX: ${{ parameters.install_prefix_mapping[build_mode] }}
                 SGXLKL_ETHREADS: ${{ ethreads }}
+                MAA_CLIENT_ID: $(MaaClientId)
+                MAA_CLIENT_SECRET: $(MaaClientSecret)
+                MAA_APP_ID: $(MaaAppId)
+                DB_USERNAME: $(DbUserName)
+                DB_PASSWORD: $(DbPassword)
 
             - task: PublishTestResults@2
               displayName: "Publish Test Results *.xml"
@@ -218,4 +223,4 @@ stages:
               --account-name clcpackages -d ${{ parameters.apt_repo_blob_container }} \
               --destination-path 1fa5fb889b8efa6ea07354c3b54903f7 \
               -s $(Build.SourcesDirectory)/apt
-            
+


davidchisnall:
I think I must be misreading this PR, so please can you add some more comments?  It looks as if these secrets will be disclosed to anyone who submits a PR to this repo, at which point they're not secret.

hukoyu:
> I think I must be misreading this PR, so please can you add some more comments? It looks as if these secrets will be disclosed to anyone who submits a PR to this repo, at which point they're not secret.

@davidchisnall @letmaik someone will define these secrets regardless of where they are stored ADO or Azure Keyvault and so on.
In MAA attestation scenario we need client id, client secret and app id. I defined them on ADO and stored encrypted. On ADO once it is defined as secret and saved, its value cannot be seen. You can override or delete and redefine but its value is not revealed. 

In this PR we are mapping these secrets to environment variables. As part of pipeline, ADO will decrypt these secrets and export and mapped to environment variables. And in the tests we will use these environment variables to use these secrets to access to MAA. Similarly there is SQL database username and password that will be used in another test scenario.

letmaik:
@hukoyu What prevents an external user from submitting a PR that sends the environment variable content to a remote server?

hukoyu:
> @hukoyu What prevents an external user from submitting a PR that sends the environment variable content to a remote server?

You are right, someone can do that. This credentials are test only credentials. The MAA service is also test only endpoint.
They can just call endpoint using credentials and get response, what else can they do?

@letmaik @davidchisnall What is your recommendation? This solution is better than having credentials directly in github. 
We can have AKV and certificate solution but this will not solve the problem since this is an open source project and someone can submit a PR to see the value of credentials and certificates will be deployed to agents by pipeline. 

I am open to suggestions.   
cc @mikbras @paulcallen @Pengpeng-Microsoft

letmaik:
> what else can they do?

I'm sure you've read the [Important security considerations](https://docs.microsoft.com/en-us/azure/devops/pipelines/repos/github?view=azure-devops&tabs=yaml#validate-contributions-from-forks) page linked from here when you ticked the checkbox: 
grafik

Ticking the checkbox also makes "Service connection credentials" available. This includes the connection to Azure Blob Storage where the Debian packages are uploaded to (see the "Publish" stage in `.azure-pipelines/template.yml`).

A solution to this problem is to have a separate test suite and pipeline containing all tests that require access to secrets and which is [not triggered automatically](https://docs.microsoft.com/en-us/azure/devops/pipelines/repos/github?view=azure-devops&tabs=yaml#run-pull-request-validation-only-when-authorized-by-your-team) for non-contributors but requires a `/azp run ..` comment by a contributor (after that person has reviewed the code). The nightly and master builds would always have access to the secrets.

I noticed you added secrets directly to the `sgx-lkl` pipeline in DevOps. Since the variables will be used in multiple pipelines this would be a pain to manage. Instead, add them as variable group in [Library](https://dev.azure.com/sgx-lkl/sgx-lkl/_library?itemType=VariableGroups).

hukoyu:
> > what else can they do?
> 
> I'm sure you've read the [Important security considerations](https://docs.microsoft.com/en-us/azure/devops/pipelines/repos/github?view=azure-devops&tabs=yaml#validate-contributions-from-forks) page linked from here when you ticked the checkbox:
> grafik
> 
> Ticking the checkbox also makes "Service connection credentials" available. This includes the connection to Azure Blob Storage where the Debian packages are uploaded to (see the "Publish" stage in `.azure-pipelines/template.yml`).
> 
> A solution to this problem is to have a separate test suite and pipeline containing all tests that require access to secrets and which is [not triggered automatically](https://docs.microsoft.com/en-us/azure/devops/pipelines/repos/github?view=azure-devops&tabs=yaml#run-pull-request-validation-only-when-authorized-by-your-team) for non-contributors but requires a `/azp run ..` comment by a contributor (after that person has reviewed the code). The nightly and master builds would always have access to the secrets.
> 
> I noticed you added secrets directly to the `sgx-lkl` pipeline in DevOps. Since the variables will be used in multiple pipelines this would be a pain to manage. Instead, add them as variable group in [Library](https://dev.azure.com/sgx-lkl/sgx-lkl/_library?itemType=VariableGroups).

@letmaik  thanks for the info. I unchecked "Make secrets available to builds of forks" for all three pipelines for now until we decide how to handle secrets in tests. I will look into solution you recommended and will discuss with the team. Thanks.

cc @mikbras @paulcallen @Pengpeng-Microsoft @jxyang @davidchisnall @prp

hukoyu:
@letmaik do you think we can merge this  PR? I need your approval.
These env variables will have secrets only when ran in nightly build on latest code.

.azure-pipelines/template.yml
```diff
@@ -218,4 +223,4 @@ stages:
               --account-name clcpackages -d ${{ parameters.apt_repo_blob_container }} \
               --destination-path 1fa5fb889b8efa6ea07354c3b54903f7 \
               -s $(Build.SourcesDirectory)/apt
-
```
letmaik, line 221: Can you revert this change?

letmaik:
As I said earlier:
> I noticed you added secrets directly to the sgx-lkl pipeline in DevOps. Since the variables will be used in multiple pipelines this would be a pain to manage. Instead, add them as variable group in Library.

hukoyu:
Dupe of https://github.com/lsds/sgx-lkl/pull/635

PR #572, opened 2020-06-30
Add sgx-lkl-docker test

letmaik:


tests/tools/sgx-lkl-docker/Makefile
```diff
@@ -0,0 +1,53 @@
...
+PROG=${APP_ROOT}/python-helloworld.py
+
+DISK_IMAGE=rootfs.img
+DATA_IMAGE=data.img
+IMAGE_SIZE=100M
+
+# The re-packaged Docker image.
+WRAP_DOCKER_IMAGE=test-sgx-lkl-docker
+WRAP_DOCKER_IMAGE_STAMP=$(WRAP_DOCKER_IMAGE).stamp
+# Note that an arbitrary/non-existing non-root user is used for launching the container.
```
prp: Why is this?

tests/tools/sgx-lkl-docker/README.md
```diff
@@ -0,0 +1,11 @@
+This test does not run SGX-LKL on the host but inside a Docker container.
+For this, a self-contained SGX-LKL installation is mounted into the container.
+To run this test, make sure you installed SGX-LKL and made it self-contained:
+
+```sh
+export SGXLKL_PREFIX=/opt/sgx-lkl
+sudo make install PREFIX=$SGXLKL_PREFIX
+sudo -E scripts/make_self_contained.sh
```
prp, line 8: Out of curiosity, why isn't this done automatically when SGX-LKL is installed?

prp (approved):
LGTM, left two questions.

tests/tools/sgx-lkl-docker/README.md
letmaik: Because this is just one particular distribution/packaging mechanism optimized for mount-into-Docker usage. I can imagine that people may want to use SGX-LKL in different ways than that as well. I agree it should be integrated into the makefiles but maybe as separate install target, like `install-contained`.

tests/tools/sgx-lkl-docker/Makefile
letmaik: Just as a proof that it doesn't just work with the pre-defined non-root user inside the image. It probably makes sense to remove the pre-defined user (999) altogether. I created that user before I added 777 permissions to the copied disk images. Once I remove the pre-defined user then I can just `USER 1000` and remove the comment above. What do you think?

letmaik: I did the change, works fine.

tests/tools/sgx-lkl-docker/README.md
prp: Is there harm in having this as the default?

letmaik: Apart from being a little unusual I don't think so. I'd prefer to do this in another PR though as it involves a few changes in the build system.

PR #575, opened 2020-07-01
Move ignored / unsupported syscalls into the kernel.

davidchisnall:
Userspace can issue these via the normal mechanism, their
implementations are replaced with no-op versions.

### DIFF
--- a/src/include/lkl/lkl_util.h
+++ b/src/include/lkl/lkl_util.h
@@ -3,6 +3,10 @@
 
 #include <openenclave/enclave.h>
 
+// Eventually we will want to turn system call tracing on and off independently
+// of the debug configuration, but for now they are the same.
+#define SGXLKL_ENABLE_SYSCALL_TRACING DEBUG
+
 /**
  * The mechanism used to implement a specified system call.
  */
@@ -45,7 +49,7 @@ typedef enum
  * This function returns `res`, so that it can be tail called on the return
  * path.
  */
-#if DEBUG
+#if SGXLKL_ENABLE_SYSCALL_TRACING
 long __sgxlkl_log_syscall(
     sgxlkl_syscall_kind type,
     long n,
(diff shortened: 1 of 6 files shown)


prp (approved):
LGTM. My only comment is that I see that we've lost the logging (e.g. `__sgxlkl_log_syscall(SGXLKL_UNSUPPORTED_SYSCALL, n, -ENOSYS, 0)`). Especially for unsupported and ignored syscalls, logging will be very important to users.

src/include/lkl/lkl_util.h
```diff
@@ -3,13 +3,66 @@
...
+/**
+ * Logs a trace message for the specified syscall.  The type argument
+ * identifies how the system call is implemented, `n` gives the number of
+ * parameters, `res` the result value. There are `params_len` variadic
+ * parameters (0 to 6) that give the arguments for the system call.
+ *
+ * This function returns `res`, so that it can be tail called on the return
+ * path.
+ */
+#if DEBUG
```
prp: We're having a discussing if this kind of tracing should only be possible for the DEBUG build. I believe that the emerging opinion is that it should also be possible for the other build types.  @letmaik @wintersteiger can you confirm?

src/lkl/syscall-overrides.c
```diff
@@ -1,10 +1,254 @@
...
+#include "lkl/posix-host.h"
 #include "lkl/syscall-overrides-fstat.h"
 #include "lkl/syscall-overrides-sysinfo.h"
-#include "lkl/posix-host.h"
+
+/**
+ * Macros for generating functions for implementing ignored system calls and
+ * for their log variants.
+ */
+#if DEBUG
```
prp: Same comment as above.

prp (approved):
LGTM, thanks. I left one comment about only supporting syscall logging with DEBUG builds.

src/include/lkl/lkl_util.h
davidchisnall: I've moved it under a separate flag, which is currently defined to be the same as `DEBUG` but can easily be changed later.

wintersteiger: Yes, confirmed, new flag is fine too.

PR #577, opened 2020-07-01
cryptsetup replacement branch

mikbras:
This PR replaces **cryptsetup** with **vicsetup** (**v**erity-**i**ntegrity-**c**rypt **setup**). **Vicsetup** implements the following:
- dm-crypt setup
- dm-integrity setup
- dm-verity setup

**Vicsetup** provides **libcryptsetup.h** as a compatibility interface (to limit changes to **src/lkl/setup.c**).

Future considerations:
1. Like **cryptsetup**, **vicsetup** formats integrity disks by invoking the **dm-integrity** kernel module (through dm-ioctl's). This formatting consumes at least 64MB of kernel memory. Kernels with less memory result in out-of-memory panics in the kernel. We might consider implementing formatting in user space to avoid this problem.

2. **Vicsetup** uses **libdevmapper**. I prototyped a replacement for this library, which would take an additional week of effort.


src/vicsetup/libvicsetup/argon2/.travis.yml
```diff
@@ -0,0 +1,25 @@
+language: c
```
davidchisnall: We are using Azure DevOps for CI, do we need this?

src/vicsetup/libvicsetup/argon2/Argon2.sln
```diff
@@ -0,0 +1,158 @@
+﻿
```
davidchisnall: We don't build anything with Visual Studio, we probably don't want visual studio solution files committed.

src/vicsetup/libvicsetup/argon2/.travis.yml
letmaik: See also https://github.com/lsds/sgx-lkl/issues/579. argon2 should be a submodule.

src/lkl/setup.c
```diff
@@ -402,28 +384,42 @@ static void* lkl_activate_crypto_disk_thread(struct lkl_crypt_device* lkl_cd)
...
-    if ((munmap_ret = lkl_sys_munmap(
-             (unsigned long)lkl_cd->disk_config->key,
-             lkl_cd->disk_config->key_len)))
-    {
-        sgxlkl_fail(
-            "Unable to unmap memory for disk encryption key: %s\n",
-            lkl_strerror((int)munmap_ret));
-    }
     lkl_cd->disk_config->key = NULL;
-    lkl_cd->disk_config->key_len = 0;
```
douglasmaciver, line 421: Why is key_len no longer set to 0 when its corresponding key is nulled?

src/vicsetup/libvicsetup/argon2/.travis.yml
mikbras: @davidchisnall: I just copied argon2 from the argon2 distro. Maik is right, I should have made argon2 a submodule.

src/vicsetup/libvicsetup/argon2/Argon2.sln
mikbras: Right. This is part of the argon2 distro. I will create a submodule for it.

src/lkl/setup.c
mikbras: Thank you! I re-added the line. My guess is that this code is an attempt to scrub the key since it is no longer needed. I assumed the size is not a secret, but I added it back just in case something uses it later. :)

tests/containers/encrypted/verify-crypt/Makefile
```diff
@@ -0,0 +1,102 @@
...
+
+FORMAT_OPTS += --type luks2
+FORMAT_OPTS += --key-file=-
+FORMAT_OPTS += --iter-time 8
+FORMAT_OPTS += --pbkdf-memory 1024
+FORMAT_OPTS += --pbkdf-parallel 2
+FORMAT_OPTS += --integrity "hmac-sha256"
+
+$(DISK_IMAGE): $(TARGET)
+	$(eval TMPNAME=$(shell basename $(shell tempfile)))
```
letmaik: Please only use the `sgx-lkl-disk` tool for creating disk images. Is it missing an option that you need for this test?

tests/containers/encrypted/verify-crypt/Makefile
```diff
@@ -0,0 +1,102 @@
+SGXLKL_ROOT=../../../..
+
+# Use this compiler to build the program
+CC=${SGXLKL_ROOT}/build/host-musl/bin/musl-gcc
```
letmaik: Tests should be able to run from an installed SGX-LKL, which wouldn't have musl-gcc available. Can you convert that to use Docker support of sgx-lkl-disk like in the other tests?

tests/containers/encrypted/create-disk/Makefile
```diff
@@ -37,3 +40,108 @@ run-hw: $(CC_IMAGE) ${CC_IMAGE_DATA}
 
 run-sw: $(CC_IMAGE) ${CC_IMAGE_DATA}
 	${SGXLKL_ENV_APP_CONFIG} ${SGXLKL_STARTER} --app-config=${CC_APP_CONFIG} --sw-debug $(CC_IMAGE)
+
+world:
```
letmaik: Do you want to remove this? Looks like a left-over.

src/vicsetup/rules.mak
```diff
@@ -0,0 +1,28 @@
+OBJECTS = $(SOURCES:.c=.o)
+
+ifdef PROGRAM
+$(PROGRAM): dirs $(OBJECTS)
+	gcc -o $(PROGRAM) $(CFLAGS) $(OBJECTS) $(LDFLAGS)
```
letmaik: Don't hard-code the compiler, use CC.

src/vicsetup/libvicsetup/result.c
```diff
@@ -0,0 +1,127 @@
...
+        case VIC_BAD_FLAGS:
+            return "VIC_BAD_FLAGS";
+        case VIC_BAD_PBKDF_TYPE:
+            return "VIC_BAD_PBKDF_TYPE";
+        case VIC_BAD_INTEGRITY_TYPE:
+            return "VIC_BAD_INTEGRITY_TYPE";
+    };
+
+    /* Unreachable */
+    return "UNKNOWN";
```
letmaik: Why not add a `default` case which fails hard?

src/vicsetup/libvicsetup/ref/cpus.c
```diff
@@ -0,0 +1,15 @@
+#include "cpus.h"
+
+#include <unistd.h>
+
+#if 0
```
letmaik: Dead code?

src/vicsetup/libvicsetup/hexdump.c
```diff
@@ -0,0 +1,161 @@
...
+#include "raise.h"
+#include "malloc.h"
+
+static void _indent(size_t indent)
+{
+    for (size_t i = 0; i < indent; i++)
+        printf("  ");
+}
+
+void vic_hexdump_special(
```
letmaik: Special?

src/vicsetup/libvicsetup/crypto.c
```diff
@@ -0,0 +1,388 @@
...
+#include <mbedtls/cipher.h>
+#include <mbedtls/sha1.h>
+#include <mbedtls/sha256.h>
+#include <mbedtls/sha512.h>
+#include <mbedtls/base64.h>
+#include <mbedtls/entropy.h>
+#include <mbedtls/ctr_drbg.h>
+#include <pthread.h>
+
+#include "../../../third_party/argon2/include/argon2.h"
```
letmaik: Why don't you add argon2 to the include path and simply `#include "argon2.h"`?

src/vicsetup/libvicsetup/argon2_argon2.c
```diff
@@ -0,0 +1,3 @@
+#include "argon2.h"
```
letmaik: Why are all those files needed? Looks like a weird work-around/short-cut for something?

src/vicsetup/libvicsetup/Makefile
```diff
@@ -0,0 +1,16 @@
+include ../defs.mak
+
+ARCHIVE = libvicsetup.a
+
+CFLAGS += -Wno-type-limits
+
+INCLUDES += -I/usr/local/include
+INCLUDES += -I../../../third_party/argon2/include
```
letmaik: Don't hard-code this here. vicsetup is an independent library that should reach out. It can receive the path to argon2 via some make variable that's passed in.

src/vicsetup/libvicsetup/Makefile
```diff
@@ -0,0 +1,16 @@
+include ../defs.mak
+
+ARCHIVE = libvicsetup.a
+
+CFLAGS += -Wno-type-limits
+
+INCLUDES += -I/usr/local/include
```
letmaik: Is this needed?

src/vicsetup/defs.mak
```diff
@@ -0,0 +1,16 @@
+TOP=$(abspath $(dir $(word 2, $(MAKEFILE_LIST))))
+
+CC = gcc
+
+CFLAGS += -g
+CFLAGS += -O3
```
letmaik, line 10: Debug and optimization should be settable via options.

src/vicsetup/defs.mak
```diff
@@ -0,0 +1,16 @@
+TOP=$(abspath $(dir $(word 2, $(MAKEFILE_LIST))))
+
+CC = gcc
```
letmaik: Don't hard-code gcc.

src/lkl/setup.c
```diff
@@ -402,28 +384,43 @@ static void* lkl_activate_crypto_disk_thread(struct lkl_crypt_device* lkl_cd)
...
+// Recall that integrity formatting is performed by the dm-integrity module
+// within the kernel. The integrity superblock is initially zero-filled. When
+// activated for the first time, the kernel module formats the device.
+// #define ENABLE_INTEGRITY
+
 static void* lkl_create_crypto_disk_thread(struct lkl_crypt_device* lkl_cd)
 {
     int err;
+    /* ATTN: vicsetup only supports 512 bytes sectors for integrity */
+    static const size_t SECTOR_SIZE = 512;
```
letmaik, line 425: Is this a permanent limitation?

src/Makefile
```diff
@@ -29,9 +29,12 @@ CFLAGS_AUTO = -g -pipe -fno-unwind-tables -fno-asynchronous-unwind-tables -ffunc
 -Werror=implicit-function-declaration -Werror=implicit-int -Werror=pointer-sign -fPIC
 CFLAGS_C99FSE = -std=c99 -ffreestanding -fexcess-precision=standard -frounding-math -Wa,--noexecstack
 
+LIBVICSETUP_DIR = $(srcdir)/vicsetup/libvicsetup
+ARGON2_DIR = $(SGXLKL_ROOT)/third_party/argon2
+
 CFLAGS_SGXLKL_ALL = $(CFLAGS_C99FSE) $(CFLAGS_AUTO) $(CFLAGS_MAIN) $(MUSL_CFLAGS)
 CFLAGS_SGXLKL_ALL += -D_XOPEN_SOURCE=700
-CFLAGS_SGXLKL_ALL += $(SGXLKL_INCLUDES) -isystem $(CRYPTSETUP_BUILD)/include -isystem $(E2FSPROGS_BUILD)/include -I$(srcdir)/libjson/include
+CFLAGS_SGXLKL_ALL += $(SGXLKL_INCLUDES) -isystem $(E2FSPROGS_BUILD)/include -I$(srcdir)/libjson/include -I$(LIBVICSETUP_DIR)/include -I$(ARGON2_DIR)/include
```
letmaik, line 37: Argon2 should be using `-isystem` since it's not our code.

letmaik (commented):
I left some high-level comments but haven't looked deeply at the code.
There's a binary file that was accidently committed: `src/vicsetup/libvicsetup/stF0Dgjz`.

src/vicsetup/libvicsetup/argon2/Argon2.sln
mikbras: To resolve this, I added an argon2 submodule.

src/lkl/setup.c
```diff
@@ -402,28 +384,43 @@ static void* lkl_activate_crypto_disk_thread(struct lkl_crypt_device* lkl_cd)
...
+    static const size_t MIN_LUKS2_ITERATIONS = 1000;
```
prp, line 427: A comment explaining what this sets would be useful.

src/lkl/setup.c
```diff
@@ -1110,6 +1013,10 @@ static void init_random()
     }
 
     fd = open("/dev/random", O_RDONLY);
+    /* Define ioctl() rather than including <sys/ioctl.h> to work around
+     * duplicate definitions of the _IO* family of macros.
+     */
+    extern int ioctl (int, int, ...);
```
prp, line 1023: How will this change after re-layering, when this code moves to userspace?

src/vicsetup/Makefile
```diff
@@ -0,0 +1,22 @@
+.PHONY: tests
```
prp, line 1: Why does this need its own Makefile? Can't this be build recursively with the Makefile under `src/`?

src/vicsetup/README.md
```diff
@@ -0,0 +1,60 @@
...
+    - cryptsetup
+
+**Vicsetup** is an all-in-one replacement for these three Linux tools. It is
+intended for use within the SGX-LKL enclave image and has various limitations
+beyond this scope.
+
+libvicsetup
+===========
+
+**Libvicsetup** provides features similar to **libcryptsetup.a**. the SGX-LKL
```
prp:
```suggestion
**Libvicsetup** provides features similar to **libcryptsetup.a**. The SGX-LKL
```

src/vicsetup/libvicsetup/blockdev.c
```diff
@@ -0,0 +1,709 @@
...
+#include <sys/stat.h>
+#include <sys/ioctl.h>
+#include <unistd.h>
+#include <sys/mount.h>
+
+#include "strings.h"
+#include "raise.h"
+#include "malloc.h"
+
+#define MAGIC 0xf3fcef718ce744bd
```
prp, line 14: A comment would be useful here.

src/vicsetup/libvicsetup/crypto.c
```diff
@@ -0,0 +1,388 @@
...
+
+#include <mbedtls/pkcs5.h>
+#include <mbedtls/aes.h>
+#include <mbedtls/cipher.h>
+#include <mbedtls/sha1.h>
+#include <mbedtls/sha256.h>
+#include <mbedtls/sha512.h>
+#include <mbedtls/base64.h>
+#include <mbedtls/entropy.h>
+#include <mbedtls/ctr_drbg.h>
```
prp, line 14: What is our story for the mbedtls dependency when this moves to userspace? Currently this is satisfied by OE's mbedtls, right?

src/vicsetup/libvicsetup/crypto.h
```diff
@@ -0,0 +1,57 @@
+#ifndef _VIC_CRYPTO_H
+#define _VIC_CRYPTO_H
+
+#include <stdint.h>
+#include <stddef.h>
+
+#include "vic.h"
+
```
prp, line 8: Some comments documenting what the API calls do would be nice.

src/vicsetup/libvicsetup/dm.c
```diff
@@ -0,0 +1,917 @@
...
+
+    /* Set the device-mapper name */
+    if (!dm_task_set_name(dmt, name))
+        RAISE(VIC_FAILED);
+
+#ifdef TRACE_TARGET
+    printf("DM-NAME: %s\n", name);
+    printf("TARGET: start{%lu} size{%lu} target{%s} params{%s}\n",
+        start, size, "integrity", params);
+#endif
```
prp, line 305: Eventually this should be integrated with SGX-LKL tracing. We also need to decide what userspace tracing will look like, as this won't be able to use the kernel-level tracing macros.

src/vicsetup/libvicsetup/dm.c
```diff
@@ -0,0 +1,917 @@
...
+#include "integrity.h"
+#include "hexdump.h"
+#include "malloc.h"
+#include "trace.h"
+#include "uuid.h"
+
+#define DM_UUID_LEN 129
+
+// #define TRACE_TARGET
+// #define USE_UDEV
```
prp: Can these be documented?

src/vicsetup/libvicsetup/dm.c
```diff
@@ -0,0 +1,917 @@
...
+
+    dm_lib_release();
+
+    return result;
+}
+
+/*
+**==============================================================================
+**
+** Experimental raw dm-ioctl code (to bypass libdevmapper).
```
prp: What is the status of this? Can this be used? What is missing?

src/vicsetup/libvicsetup/hexdump.h
```diff
@@ -0,0 +1,25 @@
+#ifndef _VIC_HEXDUMP_H
```
prp, line 1: Some comments would be nice.

src/vicsetup/libvicsetup/trace.c
```diff
@@ -0,0 +1,33 @@
+#include <stdio.h>
```
prp, line 1: This should be integrated with SGX-LKL debug/trace output.

src/vicsetup/tests/Makefile
```diff
@@ -0,0 +1,12 @@
+include ../defs.mak
```
prp, line 1: Shouldn't tests live under `tests/`? This way they get picked up by the CI pipeline.

src/vicsetup/todo.txt
```diff
@@ -0,0 +1,17 @@
+    AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
```
prp: It would be good to open Github issues instead.

prp (commented):
Thanks. I left some minor comments.

tests/containers/encrypted/verify-crypt/Makefile
mikbras: I removed this test since it was needed only for my own development. There are three other tests that create encrypted disks so the test is redundant.

tests/containers/encrypted/verify-crypt/Makefile
mikbras: I removed the test as indicated above since it is redundant.

tests/containers/encrypted/create-disk/Makefile
mikbras: Removed. Thank you for noticing.

src/vicsetup/rules.mak
mikbras: Fixed: Good catch. Thanks for finding this.

src/vicsetup/libvicsetup/result.c
mikbras: Adding a default case would fail to flag unhanded enum tags (I want a compiler error here when new enum tags are added to the header but not handled here), but I did add a hard failure.

src/vicsetup/libvicsetup/ref/cpus.c
mikbras: Removed. Thanks.

src/vicsetup/libvicsetup/hexdump.c
mikbras: Renamed to **vic_hexdump()** to **vic_hexdump_formatted()** and renamed **vic_hexdump_special()** to **vic_hexdump()**.

src/vicsetup/libvicsetup/argon2_argon2.c
mikbras: argon2 does not compile cleanly so this was to workaround various warnings (vicsetup treats warnings as errors). I will extend the build script to handle this on a per-file basis and eliminate these files.

mikbras: This permitted us to override **malloc/free** with **vic_malloc/vic_free** but this appears to no longer be necessary so I will remove it.

mikbras: Removed custom vic allocator and then argon2 wrapper files.

src/vicsetup/libvicsetup/Makefile
mikbras: Added **LIBJSON_DIR** and **ARGON2_DIR** make variables and check whether these directories exist.

src/vicsetup/libvicsetup/Makefile
mikbras: Removed. Thanks for finding this.

src/vicsetup/defs.mak
mikbras: Good idea. Added **DEBUG** and **OPTIMIZE** make switches. Defaults to -03 and no-debug.

src/vicsetup/defs.mak
mikbras: Fixed.

src/lkl/setup.c
mikbras: It is something we should fix when we can. I created issue https://github.com/lsds/sgx-lkl/issues/605 to track.

src/Makefile
mikbras: Fixed. The gcc man pages is vague on what the different is.

src/vicsetup/libvicsetup/crypto.c
mikbras: Fixed. There used to be two argon2.h headers but the local one has been deleted.

src/lkl/setup.c
mikbras: This suppression will go away because we will no longer be mixing kernel and user headers (the warnings existed before).

src/vicsetup/Makefile
mikbras: This makefile is not used when building libsgxlkl.so. Instead we add rules to src/Makefile as you suggest.

src/lkl/setup.c
mikbras: Done. Thank you.

src/vicsetup/Makefile
mikbras: This makefile is for building the libvicsetup, the vicsetup test tool, and the tests.

src/vicsetup/README.md
mikbras: Thank you for finding this.

mikbras: Fixed. Thanks.

src/vicsetup/libvicsetup/blockdev.c
mikbras: Resolved. Thanks!

src/vicsetup/libvicsetup/crypto.c
mikbras: We are partitioning the symbol namespaces into **kernel** and **user** components. That is **libsgxlkl.so** will be created by linking two objects: **kernel.o** and **user.o**. Each will contain its own mbedtls copy without exporting any mbedtls symbols.

src/vicsetup/libvicsetup/crypto.h
mikbras: Perhaps but this is an internal API. Vicsetup provides only one public API used by SGX-LKL: **libcryptsetup.h** and this is documented elsewhere.

src/vicsetup/libvicsetup/dm.c
mikbras: I created an issue to track vicsetup tracing: https://github.com/lsds/sgx-lkl/issues/607. I pressume that **cryptsetup** was not integrated with sgx-lkl tracing either.

Good point about user-space tracing. Can we use syslog?

src/vicsetup/libvicsetup/dm.c
mikbras: Resolved. Thanks.

src/vicsetup/libvicsetup/dm.c
mikbras: I removed this experimental code for now. We could spend three days and complete the replacement but I don't see why we would do it now. I originally implemented this to figure out why the kernel was crashing when calling **dm-ioctl()** to format an integrity device. It turns out that kernel memory was exhausted by the **dm-integrity** module, which consumes massive amounts of memory when formatting integrity headers (> 64MB). Expanding kernel memory worked around this for now, but we might want to perform integrity formatting in user-space (rather than letting the dm-integrity kernel module do it). I captured this in the following issue: https://github.com/lsds/sgx-lkl/issues/608.

src/vicsetup/libvicsetup/hexdump.h
mikbras: Done. :)

src/vicsetup/libvicsetup/trace.c
mikbras: Yes, I think so. I created an issue to track this: https://github.com/lsds/sgx-lkl/issues/607.

src/vicsetup/tests/Makefile
mikbras: I struggled with this. These tests are for the **vicsetup** package and are only run manually at present. The tests directory runs tests against an installed **sgx-lkl**. And since the **vicsetup** tests use the **vicsetup** command-line tool, this assumes that the **vicesetup** tool has been installed as part of **sgx-lkl**. And installing the **vicsetup** tool invites usage and implies support.

Another alternative is for us to have a unit tests directory that can be run as part of the build. I'm not sure.

src/vicsetup/todo.txt
mikbras: Resolved (I added issues and deleted this file)

letmaik:
There are various comments related to integration with SGX-LKL. I think we should treat vicsetup as an external dependency like all the other dependencies and like the original cryptsetup library. This will provide a clean cut and will allow easier stand-alone testing and re-use in other projects (even though this is not a priority currently). I don't think there is much missing to make that happen.

src/vicsetup/tests/Makefile
letmaik: See my comment https://github.com/lsds/sgx-lkl/pull/577#issuecomment-655594164.

mikbras:
> There are various comments related to integration with SGX-LKL. I think we should treat vicsetup as an external dependency like all the other dependencies and like the original cryptsetup library. This will provide a clean cut and will allow easier stand-alone testing and re-use in other projects (even though this is not a priority currently). I don't think there is much missing to make that happen.

I agree. Both **libjson** and **vicsetup** should probably just go under **third_party** and since the tests run quickly, we should run the tests as we build them. I looked into what it would take, and we could probably do it in a day or two. I create the following issues to track this: https://github.com/lsds/sgx-lkl/issues/615.

davidchisnall:
@mikbras, please can you update this from the `oe_port` branch?  It now has a number of conflicts.  It would be good to merge this soon.

src/vicsetup/libvicsetup/crypto.h
davidchisnall: As long as the expectation is that no one will ever modify the vicsetup code, that's fine.  If we expect vicsetup to be maintained, please add doc comments for all functions.

src/vicsetup/libvicsetup/defs.h
```diff
@@ -0,0 +1,37 @@
+#ifndef _VIC_DEFS_H
+#define _VIC_DEFS_H
+
+#define VIC_INLINE static __inline__
```
davidchisnall, line 4: Please add doc comments for these macros.  Many are self-explanatory to me, some are not.

src/vicsetup/libvicsetup/defs.h
```diff
@@ -0,0 +1,37 @@
...
+#define VIC_PACK_BEGIN _Pragma("pack(push, 1)")
+#define VIC_PACK_END _Pragma("pack(pop)")
+
+#define VIC_COUNTOF(ARR) (sizeof(ARR) / sizeof(ARR[0]))
+
+#define VIC_OFFSETOF(TYPE, MEMBER) __builtin_offsetof(TYPE, MEMBER)
+
+#define VIC_UNUSED __attribute__((unused))
+
+#define __LUKS_CONCAT(X, Y) X##Y
```
davidchisnall, line 15: Why is this called `__LUKS_CONCAT` not `__VIK_CONCAT`?

Symbols that start with a double underscore are reserved for the implementation in C, so defining a macro that starts with double underscores is undefined behaviour, please change this to a valid C identifier.

src/vicsetup/libvicsetup/defs.h
```diff
@@ -0,0 +1,37 @@
...
+#define VIC_CONCAT(X, Y) __LUKS_CONCAT(X, Y)
+
+#define VIC_STATIC_ASSERT(COND) \
+    typedef char VIC_CONCAT(__vic_static_assert, __LINE__)[(COND) ? 1 : -1] \
```
davidchisnall, line 19: Same comment on double underscores.  For any vaguely modern compiler, this should be defined at `_Static_assert(COND)`.  See this complete implementation:

https://github.com/freebsd/freebsd/blob/6af4a8cb884975f65a51b3b6d766e084faf55d4f/sys/sys/cdefs.h#L285

(Note that this is part of libc and so is permitted to use double-underscore identifiers)

src/vicsetup/libvicsetup/dm.c
davidchisnall: syslog is probably a good idea.  At some point, we'll add hooks to redirect syslog to an external secure audit log.

src/vicsetup/libvicsetup/goto.h
```diff
@@ -0,0 +1,27 @@
...
+#include <stdio.h>
+#include <stdint.h>
+
+#include "vic.h"
+#include "trace.h"
+
+#define GOTO(LABEL)                               \
+    do                                            \
+    {                                             \
+        __goto(__FILE__, __LINE__, __FUNCTION__); \
```
davidchisnall, line 13: Same comment about invalid identifiers.

src/vicsetup/libvicsetup/include/libcryptsetup.h
```diff
@@ -0,0 +1,288 @@
...
+
+/*
+**==============================================================================
+**
+** Internal functions (please do not call directly)
+**
+**==============================================================================
+*/
+
+int __crypt_init(struct crypt_device** cd, const char* device);
```
davidchisnall, line 131: Double underscores.

src/vicsetup/libvicsetup/luks.c
```diff
@@ -0,0 +1,585 @@
...
+#include "byteorder.h"
+#include "luks2.h"
+#include "luks1.h"
+#include "raise.h"
+#include "hexdump.h"
+#include "integrity.h"
+#include "dm.h"
+#include "strings.h"
+
+VIC_STATIC_ASSERT(VIC_OFFSETOF(vic_luks_hdr_t, magic) == 0);
```
davidchisnall, line 20: Can we have a comment explaining why these fields need to be at these offsets? (Is this matching an on-disk structure?)

davidchisnall:
@mikbras, is it possibly to push vicsetup somewhere else and use it as a submodule in our tree?  Most of the code doesn't look SGX-LKL-specific.

SeanTAllen:
@mikbras can this be closed? i believe that another PR subsumed this, yes?

mikbras:
Closed since duplicated of merged PR.

mikbras:
Closing pull request.

PR #578, opened 2020-07-02
LTP: fix random failure in recvmsg01 test case

shaikshavali1:
socket file descriptors (“sfd” & “ufd”), used in server
thread is closed before exiting from server thread.
Also some clean-up of unused code is performed.

### DIFF
--- a/tests/ltp/patches/fix_recvmsg_recvmsg01.patch
+++ b/tests/ltp/patches/fix_recvmsg_recvmsg01.patch
@@ -15,7 +15,7 @@ EFAULT error behaviour is commented/disabled until github
 issue169 is fixed.
 
 diff --git a/testcases/kernel/syscalls/recvmsg/recvmsg01.c b/testcases/kernel/syscalls/recvmsg/recvmsg01.c
-index 13bcaa4e0..b5ea895f4 100644
+index 13bcaa4e0..9947f996d 100644
 --- a/testcases/kernel/syscalls/recvmsg/recvmsg01.c
 +++ b/testcases/kernel/syscalls/recvmsg/recvmsg01.c
 @@ -54,9 +54,11 @@
@@ -30,7 +30,7 @@ index 13bcaa4e0..b5ea895f4 100644
  
  char *TCID = "recvmsg01";
  int testno;
-@@ -83,7 +85,7 @@ void cleanup(void);
+@@ -83,10 +85,10 @@ void cleanup(void);
  void cleanup0(void);
  void cleanup1(void);
  void cleanup2(void);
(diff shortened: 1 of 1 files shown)


tests/ltp/patches/fix_recvmsg_recvmsg01.patch
```diff
@@ -15,7 +15,7 @@ EFAULT error behaviour is commented/disabled until github
 issue169 is fixed.
```
ashokchandran, line 16: In the original test case, the main process creates a child process
to write/read to/from sockets infinitely. SGX-LKL supports a single
process environment. The test case is modified to use a child

One of the subtest cases is designed to test the generation of EFAULT by accessing invalid address values. Currently, SGX behavior is to call enclave abort, if the address is not within the enclave address range. Because of this test program is causing enclave abort and exiting with non-zero exit code. 

is raised to fix this behavior. The subtest cases which test
EFAULT error behavior is commented/disabled until GitHub
issue169 is fixed.

tests/ltp/patches/fix_recvmsg_recvmsg01.patch
```diff
@@ -69,27 +73,39 @@ index 13bcaa4e0..b5ea895f4 100644
...
++pthread_t tid = NULL;
  char tmpsunpath[1024];
  
  void setup(void)
-@@ -268,8 +272,9 @@ void setup(void)
+@@ -263,13 +266,19 @@ void setup(void)
+ 
+ 	signal(SIGPIPE, SIG_IGN);
+ 
+-	pid = start_server(&sin1, &sun1);
```
ashokchandran, line 89: please check if sin1 and sun1 are global so is it needed to be passed in as params

tests/ltp/patches/fix_recvmsg_recvmsg01.patch
shaikshavali1: @ashokchandran, description is not modified/changed in this patch.

tests/ltp/patches/fix_recvmsg_recvmsg01.patch
shaikshavali1: @ashokchandran, it will avoid usage of & in start_server function. I feel there is no issue if we pass address of these global variable. Please let me know your input.

tests/ltp/patches/fix_recvmsg_recvmsg01.patch
shaikshavali1: done

tests/ltp/patches/fix_recvmsg_recvmsg01.patch
```diff
@@ -69,27 +73,39 @@ index 13bcaa4e0..b5ea895f4 100644
...
++	start_server(&sin1, &sun1);
+ }
  
  void cleanup(void)
  {
 -	if (pid > 0)
 -		(void)kill(pid, SIGKILL);	/* kill server */
-+	close(ufd);
```
hukoyu, line 86: @shaikshavali1 These three lines are very confusing. There is red and green at the same line. Is this a deleted line or added line?

hukoyu: It looks like these lines are deleted.

tests/ltp/patches/fix_recvmsg_recvmsg01.patch
ashokchandran: ok that will keep changes minimal. Approved.

hukoyu (approved):
LGTM

PR #580, opened 2020-07-02
Document compatiblity with normal Linux.

davidchisnall:


davidchisnall:
@AevaOnline, please can you take a look at this?

docs/Incompatibilities.md
```diff
@@ -0,0 +1,127 @@
+Incompatibilities between SGX-LKL and x86-64 Linux
+==================================================
+
+SGX-LKL is a port of Linux to a paravirtualised SGX environment, using LKL as a porting layer.
+SGX-LKL runs x86-64 programs and these programs execute the same machine instructions as they would outside of an enclave.
+The kernel; however, runs in quite a different environment to ring 0 on an x86-64 CPU and this can lead to some important differences.
```
letmaik, line 6: I noticed this in another place as well, you use a semicolon before "however". Is that correct? I would have used a comma instead.

docs/Incompatibilities.md
```diff
@@ -0,0 +1,127 @@
...
+
+A conventional x86-64 system exposes multiple privilege domains ('rings') and provides a memory management unit that allows each userspace process to have a separate virtual address space.
+This includes the ability to alias pages such that the same physical page is in several different locations in the virtual address space.
+In contrast, SGX exposes a single virtual address space and does not allow page mappings to be modified.
+
+At the system call layer, this means that there are some restrictions on `mmap`:
+
+ - Mapping anonymous memory works reliably.
+ - Shared mappings are not automatically written back.
+ - `MAP_FIXED` mappings should only be done over existing mappings: the kernel and userspace share and address space.
```
letmaik:
```suggestion
 - `MAP_FIXED` mappings should only be done over existing mappings: the kernel and userspace share an address space.
```

docs/Incompatibilities.md
```diff
@@ -0,0 +1,127 @@
...
+In a conventional Linux system, the Linux scheduler runs userspace threads, switching to another thread either when a thread makes a blocking system call or when an interrupt fires.
+SGX; however, does not provide a mechanism for receiving timer interrupts and so cannot efficiently run a preemptive scheduler.
+
+SGX-LKL uses the lthread scheduler to provide a cooperative threading implementation, multiplexing lthreads on top of ethreads.
+The lthread scheduler running on each ethread will switch the running lthread on any system call.
+The main impact of this on userspace code is that threads that run without any system calls will not be interrupted and can consume all of the CPU resources.
+Spin locks that do not have a fallback futex path for the contended case, for example, may fail to make progress when there is a single ethread and may consume excessive amounts of CPU time in all uses.
+Threads that block on a futex will be correctly scheduled.
+
+LKL runs Linux in single-core mode and delegates scheduling of userspace threads to the lthread scheduler.
```
letmaik, line 53: What is the impact of running in "single-core mode" for userspace applications? What will `std::thread::hardware_concurrency` return? And `sysconf(_SC_NPROCESSORS_ONLN)`? Or reading `/proc/cpuinfo`? Will those return 1? Or the number of ethreads? Or lthreads? Does it work "as expected" or do users have to configure parallelisation differently in their apps for SGX-LKL (I hope not)?

docs/Incompatibilities.md
```diff
@@ -0,0 +1,127 @@
...
+As a result, the kernel's CPU affinity modes are not supported.
+Attempting to set CPU affinity will report success but silently fail.
+
+### Calling system calls
+
+On x86-64 Linux, userspace makes system calls by placing the arguments and system call number in registers and then issuing a `syscall` instruction.
+With SGX, the `syscall` instruction is not permitted and so will raise an illegal instruction trap, which will be delivered to the host, resulting in an enclave exit and an enclave resume with the host.
+We do not yet emulate the syscall instruction and so any code attempts to make system calls directly will fail.
```
letmaik:
```suggestion
We do not yet emulate the syscall instruction and so any code that attempts to make system calls directly will fail.
```

docs/Incompatibilities.md
```diff
@@ -0,0 +1,127 @@
...
+
+### Working set limits
+
+SGX, in current implementations, uses an encrypted page cache (EPC) that is directly accessible and guarantees confidentiality and integrity.
+Accesses in the EPC are slightly slower than normal memory.
+If the EPC is exhausted, the OS must swap to the rest of memory.
+This involves encrypting a page, removing it from the enclave, and providing a different page, which is then decrypted and validated.
+This process takes tens of thousands of cycles.
+
+Any workload that does not fix into the EPC will incur a significant (often a factor of ten) slowdown as a result of EPC swapping overheads.
```
letmaik:
```suggestion
Any workload that does not fit into the EPC will incur a significant (often a factor of ten) slowdown as a result of EPC swapping overheads.
```

docs/Incompatibilities.md
```diff
@@ -0,0 +1,127 @@
...
+The userpace signal handler will reflect this trap back into the SGX enclave.
+SGX-LKL will check if the faulting instruction is `cpuid` or `rdtsc` and emulate them but will generate a `SIGILL` for any other instruction.
+
+A malicious host can cause spurious illegal instruction, floating point, or segmentation fault signals to be delivered.
+
+### Imprecise time
+
+SGX does not give a trusted time source and, as described above, the `rdtsc` instruction is not allowed.
+SGX-LKL uses a variable in untrusted memory to provide a monotonic counter.
+The host updates this periodically (approximately every 5ms) and the enclave code advances it by 1ns on every query if the host has not changed the update.
```
letmaik: What happens if the host updates it to a value in the past? I guess it's ignored and the usual 1ns increment continues without update.

letmaik (commented):
I like it. Certainly for the advanced user and we need another guide & tooling to help the average user.
Should this doc say anything about musl/glibc? I'm trying to figure out what the level of abstraction for this doc should be.

docs/Incompatibilities.md
davidchisnall: That's what my copyeditor tells me to use for this construct.

docs/Incompatibilities.md
davidchisnall: `sysconf` will return return the number of ethreads, because we override them.  Not sure about `/proc/cpuinfo`.  Only one thread can be executing a system call at a time.

docs/Incompatibilities.md
davidchisnall: Yes, exactly.

docs/Incompatibilities.md
davidchisnall:
```suggestion
LKL runs Linux in single-core mode and delegates scheduling of userspace threads to the lthread scheduler.
This means that only one core can be executing system calls at any time, though userspace code can be executing on other cores at the same time.
```

docs/Incompatibilities.md
davidchisnall:
```suggestion
The host updates this periodically (approximately every 5ms) and the enclave code advances it by 1ns on every query if the host has not changed the update.
The in-enclave code ensures that time always goes forwards.
```

docs/Incompatibilities.md
letmaik: OK that makes sense now and is clear enough.

docs/Incompatibilities.md
```diff
@@ -0,0 +1,129 @@
...
+SGX-LKL is a port of Linux to a paravirtualised SGX environment, using LKL as a porting layer.
+SGX-LKL runs x86-64 programs and these programs execute the same machine instructions as they would outside of an enclave.
+The kernel; however, runs in quite a different environment to ring 0 on an x86-64 CPU and this can lead to some important differences.
+
+Compatibility differences
+-------------------------
+
+Some features are not supported at all and so applications that depend on them will not work.
+
+### Single process
```
SeanTAllen, line 13: The title seems weird given context. I am expecting a list of things to are not supported, and it says "single process" which is the limitation rather than a feature that isn't supported or  thing that won't work.

docs/Incompatibilities.md
```diff
@@ -0,0 +1,129 @@
...
+
+These restrictions are close to those of uCLinux and SGX-LKL will eventually use the no-MMU code from Linux.
+
+Because SGX does not grant control over the MMU, it is impossible on SGX1 hardware to securely change page permissions.
+SGX-LKL delegates this to the untrusted host.
+As a result, any `mprotect` call, or any `mmap` with permissions weaker than read-write-execute, should be considered advisory.
+These calls will work correctly *in the absence of an attacker* but a malicious host can simply ignore these calls.
+This will be addressed on hardware with Enclave Dynamic Memory Management (EDMM) support.
+
+### Cooperative threading
```
SeanTAllen, line 41: See single process comment.

Perhaps a small rewrite of the paragraph that leads into this "list" would be better.

docs/Incompatibilities.md
```diff
@@ -0,0 +1,129 @@
...
+The userpace signal handler will reflect this trap back into the SGX enclave.
+SGX-LKL will check if the faulting instruction is `cpuid` or `rdtsc` and emulate them but will generate a `SIGILL` for any other instruction.
+
+A malicious host can cause spurious illegal instruction, floating point, or segmentation fault signals to be delivered.
+
+### Imprecise time
+
+SGX does not give a trusted time source and, as described above, the `rdtsc` instruction is not allowed.
+SGX-LKL uses a variable in untrusted memory to provide a monotonic counter.
+The host updates this periodically (approximately every 5ms) and the enclave code advances it by 1ns on every query if the host has not changed the update.
```
SeanTAllen: 5 ms is incorrect at the moment. Currently it is 500 micros.

docs/Incompatibilities.md
```diff
@@ -0,0 +1,129 @@
...
+The in-enclave code ensures that time always goes forwards.
+This means that time within the enclave will always go forward but may jump forward.
+All of the Linux clocks are driven from this time source, so anything depending on smooth time (e.g. video / audio playback) would not be reliable.
+
+Additionally, the time exposed in the enclave is untrusted.
+The enclave may communicate with an external trusted time source but that can give only a lower bound on the current time: a malicious host could cause the enclave to sleep for an unbounded amount of time.
+
+Summary of security implications
+--------------------------------
+
```
SeanTAllen, line 115: Some headings in this section would be good as there are a couple different topics. That would also better fit the format from earlier.

docs/Incompatibilities.md
```diff
@@ -0,0 +1,129 @@
...
+SGX-LKL will check if the faulting instruction is `cpuid` or `rdtsc` and emulate them but will generate a `SIGILL` for any other instruction.
+
+A malicious host can cause spurious illegal instruction, floating point, or segmentation fault signals to be delivered.
+
+### Imprecise time
+
+SGX does not give a trusted time source and, as described above, the `rdtsc` instruction is not allowed.
+SGX-LKL uses a variable in untrusted memory to provide a monotonic counter.
+The host updates this periodically (approximately every 5ms) and the enclave code advances it by 1ns on every query if the host has not changed the update.
+The in-enclave code ensures that time always goes forwards.
```
SeanTAllen: I think this should be "The in-enclave code ensures that the monotonic counter always goes forward".

docs/Incompatibilities.md
```diff
@@ -0,0 +1,129 @@
...
+This means that time within the enclave will always go forward but may jump forward.
```
SeanTAllen: This could be read by someone as "I can't set the realtime clock to the past in the enclave" which would be incorrect.

I think rather than time here, I think it should be monotonic counter or just counter.

"This means that the counter within the enclave will always go forward..."

docs/Incompatibilities.md
```diff
@@ -0,0 +1,129 @@
...
+All of the Linux clocks are driven from this time source, so anything depending on smooth time (e.g. video / audio playback) would not be reliable.
```
SeanTAllen: "this time source" could be ambiguous. I'd suggest something like "All of the Linux clocks are driven by the monotonic counter"

docs/Incompatibilities.md
prp: I also haven't seen this before. I would use a comma.

docs/Incompatibilities.md
prp: Agreed. Perhaps say "Limitation to single process environment" or "Single process environment only".

docs/Incompatibilities.md
```diff
@@ -0,0 +1,129 @@
...
+The kernel; however, runs in quite a different environment to ring 0 on an x86-64 CPU and this can lead to some important differences.
+
+Compatibility differences
+-------------------------
+
+Some features are not supported at all and so applications that depend on them will not work.
+
+### Single process
+
+The SGX-LKL environment provides a single process.
```
prp: Perhaps explain why.
```suggestion
The SGX-LKL environment provides a single process due to the lack of certain virtual memory features inside SGX enclaves.
```

docs/Incompatibilities.md
```diff
@@ -0,0 +1,129 @@
...
+Any application that depends on spawning child processes will not work.
+
+In some future version, we expect to be able to support multiple processes but there will be no isolation between them.
+This means that using multiple processes for privilege separation will not be possible.
+
+### No MMU support
+
+A conventional x86-64 system exposes multiple privilege domains ('rings') and provides a memory management unit that allows each userspace process to have a separate virtual address space.
```
prp:
```suggestion
A conventional x86-64 system exposes multiple privilege domains ('rings') and provides a memory management unit (MMU) that allows each userspace process to have a separate virtual address space.
```

docs/Incompatibilities.md
```diff
@@ -0,0 +1,129 @@
...
+This includes the ability to alias pages such that the same physical page is in several different locations in the virtual address space.
+In contrast, SGX exposes a single virtual address space and does not allow page mappings to be modified.
+
+At the system call layer, this means that there are some restrictions on `mmap`:
+
+ - Mapping anonymous memory works reliably.
```
prp: This doesn't sound like a limitation. Remove?

docs/Incompatibilities.md
```diff
@@ -0,0 +1,129 @@
...
+ - Shared mappings are not automatically written back.
```
prp:
```suggestion
 - Shared mappings (`MAP_SHARED`) are not automatically written back.
```

docs/Incompatibilities.md
```diff
@@ -0,0 +1,129 @@
...
+ - `MAP_FIXED` mappings should only be done over existing mappings: the kernel and userspace share an address space.
```
prp:
```suggestion
 - Fixed mappings (`MAP_FIXED`) should only be done over existing mappings: the kernel and userspace share an address space.
```

docs/Incompatibilities.md
```diff
@@ -0,0 +1,129 @@
...
+   It is currently possible to do `MAP_FIXED` over kernel mappings, this will be fixed in a future version.
+
+These restrictions are close to those of uCLinux and SGX-LKL will eventually use the no-MMU code from Linux.
```
prp, line 33:
```suggestion
These restrictions are close to those of uCLinux, and SGX-LKL will eventually use the no-MMU code from Linux.
```

docs/Incompatibilities.md
```diff
@@ -0,0 +1,129 @@
...
+SGX-LKL delegates this to the untrusted host.
+As a result, any `mprotect` call, or any `mmap` with permissions weaker than read-write-execute, should be considered advisory.
+These calls will work correctly *in the absence of an attacker* but a malicious host can simply ignore these calls.
+This will be addressed on hardware with Enclave Dynamic Memory Management (EDMM) support.
+
+### Cooperative threading
+
+The SGX-LKL runtime uses an enclave thread (ethread) abstraction that roughly corresponds to a virtual CPU (VCPU) in a conventional VM environment.
+In a conventional Linux system, the Linux scheduler runs userspace threads, switching to another thread either when a thread makes a blocking system call or when an interrupt fires.
+SGX; however, does not provide a mechanism for receiving timer interrupts and so cannot efficiently run a preemptive scheduler.
```
prp, line 45: See above.

docs/Incompatibilities.md
```diff
@@ -0,0 +1,129 @@
...
+
+SGX-LKL uses the lthread scheduler to provide a cooperative threading implementation, multiplexing lthreads on top of ethreads.
```
prp: Has the term "lthread" been introduced?

docs/Incompatibilities.md
```diff
@@ -0,0 +1,129 @@
...
+
+LKL runs Linux in single-core mode and delegates scheduling of userspace threads to the lthread scheduler.
+This means that only one core can be executing system calls at any time, though userspace code can be executing on other cores at the same time.
+As a result, the kernel's CPU affinity modes are not supported.
+Attempting to set CPU affinity will report success but silently fail.
+
+### Calling system calls
+
+On x86-64 Linux, userspace makes system calls by placing the arguments and system call number in registers and then issuing a `syscall` instruction.
+With SGX, the `syscall` instruction is not permitted and so will raise an illegal instruction trap, which will be delivered to the host, resulting in an enclave exit and an enclave resume with the host.
```
prp: Long sentence -- I would split it.

prp (approved):
LGTM, left some minor comments.

docs/Incompatibilities.md
davidchisnall:
```suggestion
```

docs/Incompatibilities.md
davidchisnall: This is not a list, so doesn't need an Oxford comma.  For separating clauses, you need either a comma or a conjunction.

docs/Incompatibilities.md
davidchisnall: This paragraph is introducing lthreads.
```suggestion
SGX-LKL uses a cooperative scheduler ('lthreads') to provide a cooperative threading implementation, multiplexing cooperative lthreads on top of hardware enclave threads ('ethreads').
```

docs/Incompatibilities.md
davidchisnall:
```suggestion
With SGX, the `syscall` instruction is not permitted and so will raise an illegal instruction trap.
The trap will be delivered to the host, resulting in an enclave exit and an enclave resume with the host.
```

docs/Incompatibilities.md
davidchisnall:
```suggestion
The host updates this periodically (approximately every 500μs, subject to host scheduler variation) and the enclave code advances it by 1ns on every query if the host has not changed the update.
```

docs/Incompatibilities.md
davidchisnall:
```suggestion
The in-enclave code ensures that the monotonic counter always goes forward.
```

docs/Incompatibilities.md
davidchisnall:
```suggestion
This means that the clock source provided to the in-enclave kernel will always go forward but may jump forward.
Software running on the kernel is still free to set the wall-clock time to any value.
```

docs/Incompatibilities.md
davidchisnall:
```suggestion
All of the Linux clocks are driven from the monotonic counter, so anything depending on smooth time (e.g. video / audio playback) would not be reliable.
```

docs/Incompatibilities.md
davidchisnall: This section is a summary.

PR #583, opened 2020-07-03
LTP: Fix for bad file descriptor issue - recvfrom01

ashokchandran:
socket file descriptors used in
a thread is closed before exiting from thread.


tests/ltp/patches/fix_recvfrom_recvfrom01.patch
```diff
@@ -14,7 +14,7 @@ EFAULT error behaviour is commented/disabled until github
 issue169 is fixed.
```
shaikshavali1, line 14: Could you please check and update the patch description if needed.

shaikshavali1 (commented):
@ashokchandran, please check the patch description.

hukoyu (approved):
LGTM

PR #585, opened 2020-07-03
LTP:sendmsg01 Fix for bad file descriptor issue 

ashokchandran:
socket file descriptors used in
a thread is closed before exiting from the thread.


shaikshavali1 (approved):
@ashokchandran, LGTM

hukoyu (approved):
LGTM

PR #587, opened 2020-07-03
LTP:Fix for file descriptor issues in sendto01. Also it fixes issue 405

ashokchandran:
Make sure that the file descriptors used by test are closed properly and also enable the local endpoint shutdown test as it was failing because the socket was closed early.

Also fixes issue : https://github.com/lsds/sgx-lkl/issues/405 (Socket shutdown failure)


hukoyu (approved):
LGTM

PR #589, opened 2020-07-03
Output lthread stack traces on sgxlkl_fail and unhandled exceptions

### LINKED ISSUE #321: Print enclave backtrace on unhandled exception in kernel-level SGX-LKL code
Like https://github.com/lsds/sgx-lkl/issues/205 but for enclave kernel-side code.

EDIT: This is useful in combination with #170 for RELEASE builds as well.

prp:
This PR outputs stack traces for all threads when `sgxlkl_fail()` is called for a DEBUG build. It also outputs a stacktrace when a signal handler cannot pass an exception to LKL.

The output looks as follows:
```
...
[[  SGX-LKL ]] startmain(): enter
[[  SGX-LKL ]] lkl_start_init(): calling register_lkl_syscall_overrides()
[[  SGX-LKL ]] lkl_start_init(): fetching configuration from environment variables
[[  SGX-LKL ]] lkl_start_init(): calling initialize_enclave_event_channel()
[[  SGX-LKL ]] lkl_start_init(): calling lkl_virtio_console_add()
[[  SGX-LKL ]] WARN: Stack trace before the signal handler was invoked:
[[  SGX-LKL ]]     #0: 7fde005a58a1 in startmain(...)
[[  SGX-LKL ]]     #1: 7fde005b1430 in _exec(...)
[[  SGX-LKL ]] FAIL: Exception SIGSEGV (page fault) received before LKL is running (lt->tid=1 [kernel] code=5 addr=0x0 opcode=0x8b ret=0)
[[  SGX-LKL ]] =============================================================
[[  SGX-LKL ]] Stack traces for all lthreads:
[[  SGX-LKL ]] -------------------------------------------------------------
[[  SGX-LKL ]] *1: tid=1 [kernel]
[[  SGX-LKL ]]     #0: 7fde00b445d0 in <unknown>(...)
[[  SGX-LKL ]] -------------------------------------------------------------
[[  SGX-LKL ]] 2: tid=2 [vio-0]
[[  SGX-LKL ]]     #0: 7fde005a4eba in vio_wait_for_host_event(...)
[[  SGX-LKL ]]     #1: 7fde005a508a in vio_enclave_process_host_event(...)
[[  SGX-LKL ]]     #2: 7fde005b1430 in _exec(...)
[[  SGX-LKL ]] -------------------------------------------------------------
[[  SGX-LKL ]] 3: tid=3 [vio-1]
[[  SGX-LKL ]]     #0: 7fde005a4eba in vio_wait_for_host_event(...)
[[  SGX-LKL ]]     #1: 7fde005a508a in vio_enclave_process_host_event(...)
[[  SGX-LKL ]]     #2: 7fde005b1430 in _exec(...)
[[  SGX-LKL ]] -------------------------------------------------------------
[[  SGX-LKL ]] 4: tid=4 [vio-2]
[[  SGX-LKL ]]     #0: 7fde005a4eba in vio_wait_for_host_event(...)
[[  SGX-LKL ]]     #1: 7fde005a508a in vio_enclave_process_host_event(...)
[[  SGX-LKL ]]     #2: 7fde005b1430 in _exec(...)
[[  SGX-LKL ]] =============================================================
2020-07-03T22:08:26.000000Z [(H)ERROR] tid(0x7fdefccdc700) | :OE_ENCLAVE_ABORTING [/home/prp/sgx-lkl/openenclave/host/calls.c:oe_call_enclave_function_by_table_id:91]
[   SGX-LKL  ] init (0: 19 exit=0) [   SGX-LKL  ] FAIL: sgxlkl_ethread_init() failed (ethread_id=0 result=19 (OE_ENCLAVE_ABORTING))
Makefile:32: recipe for target 'run-sw' failed
make: *** [run-sw] Error 1
```

A current limitation is that for all but the lthread that calls `lthread_dump_all_threads()`, the stack traces reflect the stack at the time of the last context switch by the lthread scheduler.

The PR fleshes out the POC from https://github.com/lsds/sgx-lkl/pull/584.

### DIFF
--- a/src/enclave/enclave_signal.c
+++ b/src/enclave/enclave_signal.c
@@ -134,7 +134,7 @@ static uint64_t sgxlkl_enclave_signal_handler(
 
     memset(&trap_info, 0, sizeof(trap_info));
     ret = get_trap_details(exception_record->code, &trap_info);
-    if (ret != -1 && lkl_is_running())
+    if (ret != -1)
     {
         SGXLKL_TRACE_SIGNAL(
             "Exception %s received (code=%d address=0x%lx opcode=0x%x)\n",
@@ -142,6 +142,32 @@ static uint64_t sgxlkl_enclave_signal_handler(
             exception_record->code,
             exception_record->address,
             opcode);
+
+        /**
+         * If LKL has not yet been initialised, we cannot handle the
+         * exception and fail instead.
+         */
+        if (!lkl_is_running())
+        {
+#ifdef DEBUG
+            sgxlkl_error("Unhandled exception. Printing stack trace before the signal handler was invoked:\n");
+            sgxlkl_print_backtrace((void*)oe_ctx->rbp);
+#endif
+
+            struct lthread* lt = lthread_self();
+            sgxlkl_fail(
+                "Exception %s received before LKL is running (lt->tid=%i [%s] "
+                "code=%i "
+                "addr=0x%lx opcode=0x%x "
+                "ret=%i)\n",
+                trap_info.description,
+                lt ? lt->tid : -1,
+                lt ? lt->funcname : "(?)",
+                exception_record->code,
+                (void*)exception_record->address,
+                opcode,
+                ret);
+        }
 
         memset(&uctx, 0, sizeof(uctx));
         serialize_ucontext(oe_ctx, &uctx);
@@ -157,27 +183,18 @@ static uint64_t sgxlkl_enclave_signal_handler(
     else
     {
         struct lthread* lt = lthread_self();
-
-        sgxlkl_warn(
-            "Unhandled exception %s received (lt->funcname=%s lt->tid=%i "
+        sgxlkl_fail(
+            "Unknown exception %s received (lt->tid=%i [%s]"
             "code=%i "
             "addr=0x%lx opcode=0x%x "
             "ret=%i)\n",
             trap_info.description,
-            lt ? lt->funcname : "(?)",
             lt ? lt->tid : -1,
+            lt ? lt->funcname : "(?)",
             exception_record->code,
             (void*)exception_record->address,
             opcode,
             ret);
-
-        if (!lkl_is_running())
-        {
-#ifdef DEBUG
-            lthread_dump_all_threads();
-#endif
-            sgxlkl_fail("Aborting because LKL is not running yet\n");
-        }
     }
 
     return OE_EXCEPTION_CONTINUE_EXECUTION;
--- a/src/enclave/enclave_signal.c
+++ b/src/enclave/enclave_signal.c
@@ -150,7 +150,13 @@ static uint64_t sgxlkl_enclave_signal_handler(
         if (!lkl_is_running())
         {
 #ifdef DEBUG
-            sgxlkl_error("Unhandled exception. Printing stack trace before the signal handler was invoked:\n");
+            sgxlkl_error("Exception received before LKL can handle it. "
+                         "Printing stack trace saved by exception handler:\n");
+            /**
+             * Since we cannot unwind the frames in the OE exception handler,
+             * we need to print a backtrace with the frame pointer from the
+             * saved exception context.
+             */
             sgxlkl_print_backtrace((void*)oe_ctx->rbp);
 #endif
 
@@ -177,6 +183,10 @@ static uint64_t sgxlkl_enclave_signal_handler(
         info.si_addr = (void*)exception_record->address;
         info.si_signo = trap_info.signo;
 
+        /**
+         * The trap is is passed to LKL. If it can be handled, excecution will continue,
+         * otherwise LKL will abort the process.
+         */
         lkl_do_trap(trap_info.trapnr, trap_info.signo, NULL, &uctx, 0, &info);
         deserialize_ucontext(&uctx, oe_ctx);
     }
--- a/src/enclave/enclave_signal.c
+++ b/src/enclave/enclave_signal.c
@@ -143,6 +143,13 @@ static uint64_t sgxlkl_enclave_signal_handler(
             exception_record->address,
             opcode);
 
+#ifdef DEBUG
+        if (sgxlkl_trace_signal)
+        {
+            sgxlkl_print_backtrace((void*)oe_ctx->rbp);
+        }
+#endif
+
         /**
          * If LKL has not yet been initialised, we cannot handle the
          * exception and fail instead.
--- a/src/enclave/enclave_util.c
+++ b/src/enclave/enclave_util.c
@@ -1,9 +1,11 @@
 #include "enclave/enclave_util.h"
+#include "enclave/lthread.h"
 
 #include <stdarg.h>
 
 #include "openenclave/internal/print.h"
 #include "openenclave/corelibc/oemalloc.h"
+#include "openenclave/internal/backtrace.h"
 
 #define OE_STDERR_FILENO 1
 
@@ -13,6 +15,11 @@ void sgxlkl_fail(const char* msg, ...)
     oe_host_fprintf(OE_STDERR_FILENO, "[[  SGX-LKL ]] FAIL: ");
     va_start(args, msg);
     oe_host_vfprintf(OE_STDERR_FILENO, msg, args);
+
+#ifdef DEBUG
+    lthread_dump_all_threads();
+#endif
+
     oe_abort();
 }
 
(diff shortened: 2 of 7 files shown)


src/enclave/enclave_util.c
```diff
@@ -66,6 +73,30 @@ void* oe_calloc_or_die(size_t nmemb, size_t size, const char* fail_msg, ...)
     return ptr;
 }
 
+#ifdef DEBUG
+// Provide access to an internal OE function
+extern int oe_backtrace_impl(void** start_frame, void** buffer, int size);
```
letmaik, line 83: Add a comment why the public function cannot be used and maybe create an issue in OE.

src/enclave/enclave_util.c
```diff
@@ -1,9 +1,11 @@
 #include "enclave/enclave_util.h"
+#include "enclave/lthread.h"
 
 #include <stdarg.h>
 
 #include "openenclave/internal/print.h"
 #include "openenclave/corelibc/oemalloc.h"
+#include "openenclave/internal/backtrace.h"
```
letmaik, line 9: `#ifdef` this as well, just for consistency.

src/enclave/enclave_signal.c
```diff
@@ -143,6 +143,32 @@ static uint64_t sgxlkl_enclave_signal_handler(
...
 
+        /**
+         * If LKL has not yet been initialised, we cannot handle the
+         * exception and fail instead.
+         */
+        if (!lkl_is_running())
+        {
+#ifdef DEBUG
+            sgxlkl_error("Unhandled exception. Printing stack trace before the signal handler was invoked:\n");
+            sgxlkl_print_backtrace((void*)oe_ctx->rbp);
```
letmaik, line 167: If I understand correctly this trace is from the ethread. Maybe make that clear in the output since you also mention "lthreads" in the other traces.

src/enclave/enclave_signal.c
```diff
@@ -143,6 +143,32 @@ static uint64_t sgxlkl_enclave_signal_handler(
...
+                trap_info.description,
+                lt ? lt->tid : -1,
+                lt ? lt->funcname : "(?)",
+                exception_record->code,
+                (void*)exception_record->address,
+                opcode,
+                ret);
+        }
+
         memset(&uctx, 0, sizeof(uctx));
```
letmaik, line 185: I think for auditing it would be nice if there was a comment saying where unhandled exceptions are processed in case LKL *is* running. The code below is likely hard to follow to the final location.

letmaik (approved):

src/enclave/enclave_signal.c
prp: It is the trace for the lthread, as all of our stacks are associated with lthreads. The difference here is that I use the frame pointer that was used when the exception happened, as opposed to the frame pointer used by the signal handler.

prp:
I would also like to provide backtraces for all currently running lthreads (see limitation above), but it is not clear to me what the cleanest way of doing this is. We need to ensure that `sgxlkl_fail()` will abort no matter what, even when an lthread is stuck in a userspace busy loop. The only idea that I came up with is to send them signals, but the synchronisation for this will be messy. Thoughts?

src/enclave/enclave_signal.c
letmaik: OK, and dumping both is necessary because exception handling involves roundtrips that lose the original stacktrace? I'm trying to figure out what the difference to a normal environment is.

davidchisnall:
> I would also like to provide backtraces for all currently running lthreads (see limitation above), but it is not clear to me what the cleanest way of doing this is. We need to ensure that sgxlkl_fail() will abort no matter what, even when an lthread is stuck in a userspace busy loop. The only idea that I came up with is to send them signals, but the synchronisation for this will be messy. Thoughts?

Sending a signal with LKL won't interrupt the threads, we need to force an ethread abort, so we'd need an ocall that sends a signal to all of the host threads running lthreads, forces enclave exit, and then restarts at a custom stack-trace handler.  

Since this debug mode, I think that's all fine, but there are significant security concerns about using libUnwind (ideally we wouldn't even link it in non-debug modes).  It trusts everything on the stack and in the binary and then follows pointers with no checking.  Since our exception frames are writeable, a bug could potentially do all manner of things (the unwinder also invokes a personality function for each frame, which can also write to memory addresses based on data that might be corrupt, so there are lots of ways to leak secrets from SGX).

src/enclave/enclave_signal.c
prp: I think that OE uses a separate stack for its exception handler inside the enclave. When I use the compiler generated frame pointer, I didn't see the lthread frames, which is why I retrieve it from the saved exception context. In contrast, `sgxlkl_print_backtrace()` iterates over the lthread scheduler state and dumps out the saved stacks for all lthreads.

I can add comments to explain this better.

letmaik:
> Since this debug mode, I think that's all fine, but there are significant security concerns about using libUnwind (ideally we wouldn't even link it in non-debug modes). It trusts everything on the stack and in the binary and then follows pointers with no checking. Since our exception frames are writeable, a bug could potentially do all manner of things (the unwinder also invokes a personality function for each frame, which can also write to memory addresses based on data that might be corrupt, so there are lots of ways to leak secrets from SGX).

@prp This should be mentioned in the code along one of the `#ifdef`s.

src/enclave/enclave_util.c
prp: Done

src/enclave/enclave_signal.c
davidchisnall: libUnwind can't unwind out of exception frames because exception frames don't contain DWARF unwind tables that tell them how to use the exception context to get back to the host.

src/enclave/enclave_util.c
prp: Done.

prp:
> > I would also like to provide backtraces for all currently running lthreads (see limitation above), but it is not clear to me what the cleanest way of doing this is. We need to ensure that sgxlkl_fail() will abort no matter what, even when an lthread is stuck in a userspace busy loop. The only idea that I came up with is to send them signals, but the synchronisation for this will be messy. Thoughts?
> 
> Sending a signal with LKL won't interrupt the threads, we need to force an ethread abort, so we'd need an ocall that sends a signal to all of the host threads running lthreads, forces enclave exit, and then restarts at a custom stack-trace handler.

Yes, I mean trigger a signal for the ethread to cause an AEX and invoke the OE exception handler. The problem is that this adds quite a bit of complexity. Unless somebody can see a cleaner way, I won't do this as part of this PR.

prp:
> Since this debug mode, I think that's all fine, but there are significant security concerns about using libUnwind (ideally we wouldn't even link it in non-debug modes). It trusts everything on the stack and in the binary and then follows pointers with no checking. Since our exception frames are writeable, a bug could potentially do all manner of things (the unwinder also invokes a personality function for each frame, which can also write to memory addresses based on data that might be corrupt, so there are lots of ways to leak secrets from SGX).

Doesn't OE always link in backtrace functionality?

src/enclave/enclave_signal.c
prp: Done

src/enclave/enclave_signal.c
```diff
@@ -151,33 +183,28 @@ static uint64_t sgxlkl_enclave_signal_handler(
         info.si_addr = (void*)exception_record->address;
         info.si_signo = trap_info.signo;
 
+        /**
+         * The trap is is passed to LKL. If it can be handled, excecution will continue,
```
letmaik, line 194: In the abort case, would it print a backtrace? If not, then this PR only partially solves #321, right?

prp: Ah, I now understand what you meant in #321. No, you won't get a backtrace in that case. The problem is that oe_backtrace() cannot resolve application symbols.

(If we figured out how to get application symbols, we could always print a backtrace when a signal is received and `SGXLKL_TRACE_SIGNAL=1`. This would be easier than figuring out when LKL is aborting a process due to a signal.)

prp: @letmaik I had a quick look at how `oe_backtrace()` works: it uses about 10 lines of code (not `libunwind`) to walk the stack inside the enclave (with some access checks), and then does an ocall to resolve addresses to symbols. The ocall resolves the symbols based on the `libsgxlkl.so` ELF image that was loaded into the enclave.

This isn't sufficient to resolve application symbols, which are loaded dynamically from ELF images inside the enclave. Another complication is that application symbols would have to be resolved in userspace, whereas the mechanism in this PR is implemented in kernel space. In general, I would say that producing stack traces for applications is an orthogonal problem to what SGX-LKL does. It should be possible to do this entirely within the enclave in userspace, without help from SGX-LKL.

You may want to update #321.

letmaik: I updated #321 to be kernel-only. Just to clarify though, if an exception in userspace within the user application is raised and unhandled by the application, will that surface as an unhandled kernel-level exception (with unresolved symbols) as in this PR? A similar question, since we're splitting SGX-LKL into kernel code and userspace code at the moment, what will happen to exceptions in the userspace code that are from SGX-LKL and not from the application? I can't remember exactly what the split is, but I think mounting disks would move to userspace, right?

davidchisnall (approved):

src/enclave/enclave_signal.c
prp: No, because if the signal is not handled by the enclave application, the SGX-LKL exception handler will not see this but only LKL. LKL will then abort the thread/process but this happens entirely within LKL code. The abort will result in a graceful SGX-LKL shutdown that returns an error code.

You will get valid stack traces for userspace code as long as it is part of `libsgxlkl.so`. There was some discussion about splitting userspace into a separate file, and things may change then. It all depends if `oe_backtrace` still knows how to find the debug symbols.

PR #591, opened 2020-07-06
Fix race condition in initialize_enclave_event_channel()

prp:
This PR fixes a race condition in the VIO event channel initialisation.

When creating enclave threads for processing host events via the event channel mechanism, the event channels are distinguished based on the associated lthread pointer. This pointer is only updated after `lthread_create()` has returned, which means that a thread may access an uninitialised value. 

Fixes https://github.com/lsds/sgx-lkl/issues/564.

### DIFF
--- a/src/enclave/enclave_event_channel.c
+++ b/src/enclave/enclave_event_channel.c
@@ -179,14 +179,15 @@ void initialize_enclave_event_channel(
 
         *dev_id = enc_dev_config[i].dev_id;
 
-        struct lthread* lt = NULL;
         if (lthread_create(
-                &lt, NULL, vio_enclave_process_host_event, (void*)dev_id) != 0)
+                &vio_tasks[i],
+                NULL,
+                vio_enclave_process_host_event,
+                (void*)dev_id) != 0)
         {
             oe_free(vio_tasks);
             sgxlkl_fail("Failed to create lthread for event channel\n");
         }
-        vio_tasks[i] = lt;
     }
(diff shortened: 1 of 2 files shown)


letmaik:
Could you add a description on what the issue was and how you fixed it?

src/enclave/enclave_event_channel.c
```diff
@@ -172,14 +179,15 @@ void initialize_enclave_event_channel(
 
         *dev_id = enc_dev_config[i].dev_id;
 
-        struct lthread* lt = NULL;
         if (lthread_create(
-                &lt, NULL, vio_enclave_process_host_event, (void*)dev_id) != 0)
+                &vio_tasks[i],
```
wintersteiger, line 183: Does that actually solve the underlying problem? `lthread_create` follows the same pattern (`*new_lt` vs `lt`), without locking as far as I can tell.

wintersteiger (approved):
LGTM, except I'm not 100% sure it solves the root cause of the problem.

src/enclave/enclave_event_channel.c
prp: I think that it should be fine. The race only happens when `vio_enclave_process_host_event()` starts running, which only happens after `lthread_create()` has updated `*new_lt`.

wintersteiger: Yeah, you're right, that looks fine!

SeanTAllen:
nice

Issue #555, opened 2020-06-26
[Tests] Rare failure of basic/global_vars_test due to wrong errno


prp:
The following test "basic-global_vars_test-(nonrelease)-(run-sw)-(8-ethreads)" sometimes fails as follows:
```
[    0.135952] EXT4-fs (vda): mounted filesystem without journal. Opts: 
fopen failed errno(2)
TEST FAILED
test: 1 failed
[    0.144929] reboot: Restarting system
```

This is odd because the code for the test expects the first `fopen` to return an `errno` of 2, which is what happens: https://github.com/lsds/sgx-lkl/blob/c09fde496d58fe4227bac2a5f0575ae7e602c027/tests/basic/global_vars_test/global_vars_test.c#L89

One guess would be that something else is changing `errno`?

(cc: @davidchisnall @SeanTAllen)

prp:
```
Execution timeout: 60
Will run tests for run-sw
/opt/sgx-lkl-debug/bin/sgx-lkl-disk create --size=5M --docker=./Dockerfile sgx-lkl-rootfs.img
Using detected host DNS server in /etc/resolv.conf: 168.63.129.16
Creating sgx-lkl-rootfs.img from Dockerfile ./Dockerfile...
Building Docker image...
Creating disk image file from Docker container...
Succesfully created sgx-lkl-rootfs.img.
SGXLKL_VERBOSE=1 SGXLKL_KERNEL_VERBOSE=1 /opt/sgx-lkl-debug/bin/sgx-lkl-run-oe --sw-debug sgx-lkl-rootfs.img global_vars_test -a -b -c hello
[   SGX-LKL  ] SGX-LKL (OE) 0.2.0-dev (58fa34b) LKL 5.3.0 [DEBUG build (-O0)] [SOFTWARE DEBUG]
[   SGX-LKL  ] nproc=8 ETHREADS=8 CMDLINE="mem=32M"
[   SGX-LKL  ] HW TLS support: conf->fsgsbase=0
[   SGX-LKL  ] Registering disk 0 (path='sgx-lkl-rootfs.img', mnt='/', [RW    ])
[   SGX-LKL  ] No tap device specified, networking will not be available.
[   SGX-LKL  ] get_signed_libsgxlkl_path... result=/opt/sgx-lkl-debug/bin/../lib/libsgxlkl.so.signed
[   SGX-LKL  ] oe_create_enclave... result=0 (OE_OK)
[...]

PR #592, opened 2020-07-06
Fix globals test and make more verbose

prp:
This PR updates the globals test to save `errno` earlier and make it more verbose.

This is to address the issue in #555.

### DIFF
--- a/tests/basic/global_vars_test/global_vars_test.c
+++ b/tests/basic/global_vars_test/global_vars_test.c
@@ -87,13 +87,27 @@ void perform_errno_test(void)
     _do_file_ops("file_fake.txt");
     l_err_no = errno;
     if (l_err_no != 2)
+    {
         prog_exec_status |= (1 << en_test_errno_fail_index);
+        fprintf(
+            stdout,
+            "fake test failed: errno=%i l_err_no=%i\n",
+            errno,
+            l_err_no);
+    }
 
     // reset the errno to verify the valid scenario
     _do_file_ops("/helloworld.txt");
     l_err_no = errno;
     if (l_err_no != 0)
+    {
         prog_exec_status |= (1 << en_test_errno_pass_index);
+        fprintf(
+            stdout,
+            "valid test failed: errno=%i l_err_no=%i\n",
+            errno,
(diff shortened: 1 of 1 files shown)


letmaik:
I find this test kind of badly written. `errno` should be captured immediately if it's used in later analysis. Also, there's no `fprintf` for the case where `fclose` fails.

prp:
> I find this test kind of badly written. `errno` should be captured immediately if it's used in later analysis. Also, there's no `fprintf` for the case where `fclose` fails.

I agree, but do you see a reason why the test would sometimes fail? I'm working under the assumption that the issue in #555 indicates a bug in SGX-LKL.

letmaik:
The only other function called is `fprintf` but I guess that's not changing `errno`? If that's the case, then yes, looks like a weird bug.

prp:
I think that it may change it. Let me update the test.

tests/basic/global_vars_test/global_vars_test.c
```diff
@@ -84,16 +87,30 @@ void perform_errno_test(void)
...
-    l_err_no = errno;
+    l_err_no = _do_file_ops("file_fake.txt");
+
     if (l_err_no != 2)
+    {
         prog_exec_status |= (1 << en_test_errno_fail_index);
+        fprintf(
+            stdout,
+            "fake test failed: errno=%i l_err_no=%i\n",
+            errno,
```
letmaik, line 98: Do you want to keep printing errno now that you changed the test?

prp: I think that there is no harm in doing it. I'm still not entirely sure if this will fix the problem or not.

letmaik (approved):

davidchisnall:
`fclose` will call `close`, which will change `errno`.  `fprintf` is a fun one, it may just buffer internally, but it may call `fwrite` that will set `errno`.  The value of `errno` is undefined across the call, but if the buffer is empty and the stream is buffered then it may not actually modify `errno`.

prp:
...which would explain the non-deterministic behaviour that we have seen.

PR #595, opened 2020-07-06
LTP:fix bad file descriptor error in sendfile tests

shaikshavali1:
Fixed bad file descriptor error in sendfile02, sendfile05
sendfile06 test cases and also performed code cleanup.


ashokchandran (commented):
changes look good to me.

hukoyu (approved):

PR #596, opened 2020-07-06
LTP: fcntl cleanup fcntl24 fcntl25 fcntl32

ashokchandran:
Issue: These tests were mounting root file system again though it is already available
Fix : Remove mount / unmount so tests use already mounted root file system.


hukoyu (approved):
LGTM

PR #604, opened 2020-07-07
Add memory system call replacements.

davidchisnall:
These system calls are used for heap setup and so we can't yet remove
the musl wrappers, but this provides the kernel support that we'll need
when we do.


src/enclave/enclave_mem.c
```diff
@@ -390,7 +270,7 @@ void* enclave_mmap(
         if (prot != -1 && (!zero_pages || found_only_fresh_pages))
         {
             // Set requested page permission
-            mprotect(ret, length, prot);
+			sgxlkl_host_syscall_mprotect(&mprotect_ret, ret, length, prot);
```
prp: Is there some weird indentation here?

src/include/enclave/enclave_mem.h
```diff
@@ -29,13 +29,15 @@ void* enclave_mremap(
     size_t new_length,
     int mremap_fixed);
 
-int enclave_mmap_flags_supported(int flags, int fd);
+extern int mmap_files; // Allow MAP_PRIVATE or MAP_SHARED?
```
prp, line 27: Why not define `mmap_files` here?

prp (approved):
LGTM

src/include/enclave/enclave_mem.h
davidchisnall: And make it common linkage?  I try to avoid common linkage whenever possible because it causes all manner of subtle linker behaviours.

src/enclave/enclave_mem.c
davidchisnall: I forgot to run clang-format before I committed this.  Fixed now.

vtikoo (approved):

src/enclave/enclave_mem.c
```diff
@@ -98,98 +89,6 @@ static size_t addr_to_index(void* addr)
...
-    }
-    // File-backed mapping (if allowed)
-    else if (fd >= 0 && enclave_mmap_flags_supported(flags, fd))
-    {
-        mem =
-            enclave_mmap(addr, length, flags & MAP_FIXED, prot | PROT_WRITE, 0);
-        if (mem > 0)
-        {
-            // Read file into memory
-            lseek(fd, offset, SEEK_SET);
```
SeanTAllen, line 130: Did you intend to remove this lseek when this function was moved and partially rewritten?

davidchisnall: Yes.  lseek changes the offset in the fd, which is something mmap should not do.

SeanTAllen: David replied elsewhere that it was intentional

SeanTAllen:
I'm working on this PR now.

SeanTAllen:
This is currently blocked waiting for #577 to land. In the meantime, I've done some related work in #655.

Once #655 and #577 are merged to oe_port, I'm going to pick up the work on this again and remove the syscall overrides for mmap functions from sgx-lkl-musl and into system call replacements like this PR has.

Some of the changes needed to make this change possible are part of #655.

SeanTAllen:
Closed in favor of #710. I'm leaving the branch around until #710 is merged.

PR #606, opened 2020-07-07
LTP: fixed issues in listxattr 01,02 test cases

shaikshavali1:
Tests were failing as no xattr support in a loop device
So modified the tests to use the root filesystem.


hukoyu (approved):
LGTM

PR #610, opened 2020-07-08
LTP:Fix for mknodat02

ashokchandran:
Issue: Test requires a test file system. This was failing due to memory of SGXs set to 32MB and test file system needs256 MB.
Fix: Use root file system for the test


hukoyu (approved):
LGTM

shaikshavali1 (commented):
@ashokchandran, LGTM

PR #614, opened 2020-07-08
Rebase OE with master for new EEID changes

jhand2:
Update OE submodule after rebase of branch.

Note that this PR also contains fixes for:
* Name change in sgx/attestation.edl
* Some instances where oe_free was being used to free memory not allocated by oe_malloc. This was caught by new debugmalloc functionality in OE.


jhand2:
Looks like this was updated already. Closing

PR #616, opened 2020-07-09
LTP: fix test case mkdir09 failure

shaikshavali1:
Test case was using multiple processes to create and delete
directories. SGX-LKL environment multiple processes are not supported.
Hence modified the test case to use multiple threads instead of multiple
processes.


prp:
@shaikshavali1 The test fails in the CI.

shaikshavali1:
> @shaikshavali1 The test fails in the CI.

@prp, I am investigating the failure.

hukoyu (approved):
LGTM

PR #619, opened 2020-07-10
[Test] add a base test for self host network

praenubilus:
sgxlkl has a self host ipv4 network gateway address 10.0.1.254, that can be used by application inside enclave to forward packets from tap device. This test is created to verify whether the service is available.

### DIFF
--- a/tests/basic/network_self_host/Makefile
+++ b/tests/basic/network_self_host/Makefile
@@ -1,0 +1,36 @@
+# SGX-LKL has a self host ipv4 network gateway address 10.0.1.254, that can 
+# be used by application inside sgxlkl enclave to forward packets from tap
+# device. This test ping the self host gateway 10.0.1.254 with a 2 second timeout. 
+
+include ../../common.mk
+
+CC_APP=/bin/ping
+
+CC_APP_CMDLINE=${CC_APP} -c 1 -W 2 10.0.1.254
+
+CC_IMAGE_SIZE=50M
+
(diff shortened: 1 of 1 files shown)


praenubilus:
can you help to review @hukoyu @jxyang

letmaik:
PR description is missing. Also, in the test itself there should be some documentation as it's not clear what this is testing.

praenubilus:
> PR description is missing. Also, in the test itself there should be some documentation as it's not clear what this is testing.

@letmaik I have add the description. For the documentation, where should I put it? Because I just follow the format of basic/networking test. Should I include a README.md file?

letmaik:
@prp I don't know much about this. Is it self-explanatory or does it need a readme file or similar?

prp:
> @prp I don't know much about this. Is it self-explanatory or does it need a readme file or similar?

I think that the Makefile should include a comment at the beginning that explains the test and its purpose.

SeanTAllen:
@prp it has always been my assumption with this project that the README.md at the same level as the Makefile is where I would find that information. I believe that is in general how most tests that have any such info operate.

jxyang:
@SeanTAllen That's not true for tests. A readme is usually required for samples but not tests.

@praenubilus Can you add a comment like @prp suggested?

SeanTAllen:
@jxyang thanks

praenubilus:
I have added a description at the begin of the Make file. @jxyang @prp , any other comments I should address for merge?

jxyang (approved):

PR #618, opened 2020-07-09
Implementing nightly build only tests mechanism

hukoyu:
With this PR we can mark any test (long running ones for example) as nightly build only 
by adding them to file .azure-pipelines/other/nighlty_run_only_tests.txt

If SGXLKL_NIGHTLY_RUN  environment variable is NOT defined, the tests in this file will be skipped. 
These tests will run only if SGXLKL_NIGHTLY_RUN env variable is defined.

PR #620, opened 2020-07-10
MAA Remote Attestation test scenario

hukoyu:
Test Scenario:
- we have two enclaves: sgxlkl_enclave and oe_enclave.
- oe_enclave is the tls server and runs and waits for connection from sgxlkl_enclave
- sgxlkl_enclave starts and connects to oe_enclave
- oe_enclave waits for 60 seconds after startup to give enough time to sgxlkl_enlcave to start and connect
- oe_enclave sends a query request to sgxlkl_enclave at which point sgxlkl_enclave attests the oe_enclave through MAA (Microsoft Azure attestation Service)
- sgxlkl_enclave gets the report from oe_enclave and then gets authentication token from AAD and then calls MAA for JWT token and gets and verifies it and after that returns result to oe_enclave

**This PR depends on these 3 PRs:** 
https://github.com/lsds/sgx-lkl/pull/618 to run this test with nightly build only. 
https://github.com/lsds/sgx-lkl/pull/571 for ADO secrets to map to env variables
https://github.com/lsds/sgx-lkl/pull/108 for cert generation. For now untrusted channel used as work around


hukoyu:
cc @praenubilus

PR #621, opened 2020-07-10
Add missing host config auto-generation dependency

wintersteiger:
Without the dependency rule, the host config was not auto-generated at the right time in all settings.

Signed-off-by: Christoph M. Wintersteiger <cwinter@microsoft.com>


letmaik (approved):

davidchisnall (approved):
Thanks

PR #622, opened 2020-07-10
Disable shallow submodule update

letmaik:
There are repeated CI failures like
> fatal: shallow file has changed since we read it
Fetched in submodule path 'openenclave/3rdparty/optee/optee_os', but it did not contain d1634ce8ff4a39242d4d333392e260e00405e471. Direct fetching of that commit failed.

I think this happens when a submodule points to a commit that is not a tag. GitHub disallows direct fetching of commits. Cloning submodules in full will not have that issue at the cost of slightly longer checkout time.


prp (approved):
LGTM

PR #623, opened 2020-07-11
Implementing nightly build only tests mechanism

hukoyu:
With this PR we can mark any test (long running ones for example) as nightly build only
by adding them to file .azure-pipelines/other/nighlty_run_only_tests.txt

If SGXLKL_NIGHTLY_BUILD environment variable is set to false, the tests in this file will be skipped.
These tests will run only if SGXLKL_NIGHTLY_BUILD env variable is true.


hukoyu:
Dupe of https://github.com/lsds/sgx-lkl/pull/624

PR #624, opened 2020-07-11
Implement nighly build test mechanism

hukoyu:
With this PR we can mark any test (long running ones for example) as nightly build only
by adding them to file .azure-pipelines/other/nighlty_run_only_tests.txt

If SGXLKL_NIGHTLY_BUILD environment variable is set to false, the tests in this file will be skipped.
These tests will run only if SGXLKL_NIGHTLY_BUILD env variable is true.

### DIFF
--- a/.azure-pipelines/scripts/junit_utils.sh
+++ b/.azure-pipelines/scripts/junit_utils.sh
@@ -136,12 +136,12 @@ function JunitTestFinished()
         fi
 	# Replace STACK_TRACE place holder with stack trace file
         AddLogFileToJunit "$stack_trace_file_path" "$junit_file_path" "STACK_TRACE"
-    fi
 
-    # Add stdout file to junit if exist
-    if [[ -f "$stdout_file_path" ]]; then
-        echo "      <system-out>STD_OUT_MESSAGE</system-out>" >> "$junit_file_path"
-        AddLogFileToJunit "$stdout_file_path" "$junit_file_path" "STD_OUT_MESSAGE"
+        # Add stdout file to junit if exist
+        if [[ -f "$stdout_file_path" ]]; then
+            echo "      <system-out>STD_OUT_MESSAGE</system-out>" >> "$junit_file_path"
+            AddLogFileToJunit "$stdout_file_path" "$junit_file_path" "STD_OUT_MESSAGE"
+        fi
     fi
 
     {
--- a/.azure-pipelines/scripts/test_runner.sh
+++ b/.azure-pipelines/scripts/test_runner.sh
@@ -200,7 +200,7 @@ suite_test_start_time=$(date +%s)
 
 failure_identifiers=()
 while IFS= read -r line; do failure_identifiers+=("$line"); done < "$failure_identifiers_file"
-
+echo "Nightly build: $SGXLKL_NIGHTLY_BUILD"
 for file in "${file_list[@]}";
 do
     SkipTestIfDisabled
--- a/.azure-pipelines/scripts/test_runner.sh
+++ b/.azure-pipelines/scripts/test_runner.sh
@@ -19,7 +19,7 @@ if [ -z "$SGXLKL_BUILD_MODE" ]; then
     exit 1
 fi
 if [ -z "$SGXLKL_NIGHTLY_BUILD" ]; then
-    echo "ERROR: 'SGXLKL_NIGHTLY_BUILD' is undefined. Please export SGXLKL_NIGHTLY_BUILD_MODE=true|false"
+    echo "ERROR: 'SGXLKL_NIGHTLY_BUILD' is undefined. Please export SGXLKL_NIGHTLY_BUILD_MODE=0|1"
     exit 1
 fi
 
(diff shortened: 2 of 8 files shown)


.azure-pipelines/scripts/test_runner.sh
```diff
@@ -18,6 +18,10 @@ if [ -z "$SGXLKL_BUILD_MODE" ]; then
     echo "ERROR: 'SGXLKL_BUILD_MODE' is undefined. Please export SGXLKL_BUILD_MODE=<mode>"
     exit 1
 fi
+if [ -z "$SGXLKL_NIGHTLY_BUILD" ]; then
```
jxyang, line 21: If not defined, we should treat as non-nightly build.

.azure-pipelines/scripts/test_runner.sh
```diff
@@ -134,12 +138,24 @@ function GetReadyToRunNextTest()
...
+    if [[ $is_test_disabled -eq 0 && $SGXLKL_NIGHTLY_BUILD -eq 0 ]]; then
+        is_test_nightly_only=$(grep -c "$file" "$nightly_tests_file")
+        if [[ $is_test_nightly_only -ge 1 ]]; then
+            echo "Test $file is marked nightly build only. Skipping test..."
+            echo "To enable the test remove $file from $nightly_tests_file"
+        fi
+    fi
+
+    # if this test is disabled set counters and skip to next test
+    if [[ $is_test_disabled -ge 1 || $is_test_nightly_only -ge 1 ]]; then
```
jxyang: `is_test_nightly_only -ge 1` is confusing. You can remove it by setting `is_test_disabled ` on line 152.

.azure-pipelines/scripts/test_runner.sh
```diff
@@ -183,7 +200,7 @@ suite_test_start_time=$(date +%s)
 
 failure_identifiers=()
 while IFS= read -r line; do failure_identifiers+=("$line"); done < "$failure_identifiers_file"
-
+echo "This is Nightly Build: $SGXLKL_NIGHTLY_BUILD"
```
jxyang: Confusing message. Suggest only printing when $SGXLKL_NIGHTLY_BUILD is set to 1.

.azure-pipelines/scripts/junit_utils.sh
```diff
@@ -136,12 +136,12 @@ function JunitTestFinished()
         fi
 	# Replace STACK_TRACE place holder with stack trace file
         AddLogFileToJunit "$stack_trace_file_path" "$junit_file_path" "STACK_TRACE"
-    fi
 
-    # Add stdout file to junit if exist
-    if [[ -f "$stdout_file_path" ]]; then
-        echo "      <system-out>STD_OUT_MESSAGE</system-out>" >> "$junit_file_path"
-        AddLogFileToJunit "$stdout_file_path" "$junit_file_path" "STD_OUT_MESSAGE"
+        # Add stdout file to junit if exist
```
jxyang, line 140: This is not related to nightly build, right?

.azure-pipelines/scripts/test_runner.sh
jxyang, line 21: Why do we force another environment? Can we treat not defined as 0?

.azure-pipelines/scripts/test_runner.sh
jxyang: `$is_test_nightly_only -ge 1` is confusing. You can get rid of it by setting `$is_test_disabled` on line 152.

.azure-pipelines/scripts/test_runner.sh
jxyang: Confusing message. Suggest rewording.

.azure-pipelines/scripts/junit_utils.sh
hukoyu: Yes this not directly related but there is an issue with downloaded html page encoding with basic/networking test that blocks junit xml file and all core test junit file publish. To be able to see test results I need to fix this. This can be a separate PR but if it is OK I want to fix it as part of this PR.

.azure-pipelines/scripts/test_runner.sh
hukoyu: Done removed. This message is not necessary actually.

.azure-pipelines/scripts/test_runner.sh
hukoyu: Done. Got rid of is_test_nightly_only  and used is_test_disabled

.azure-pipelines/scripts/test_runner.sh
hukoyu: Done

.azure-pipelines/scripts/test_runner.sh
hukoyu: Make sense. Printed message for 1 only. Did rewording.

.azure-pipelines/scripts/test_runner.sh
hukoyu: Done

.azure-pipelines/scripts/test_runner.sh
hukoyu: Done

jxyang (approved):

.azure-pipelines/template.yml
```diff
@@ -47,6 +47,9 @@ parameters:
 - name: 'publish'
   type: boolean
   default: false
+- name: 'nightly_build'
+  type: number
```
letmaik, line 51: Why not a boolean? This would be cleaner I think. You could then conditionally include the env var:
```yaml
            - bash: .azure-pipelines/scripts/test_runner.sh ${{ test_suite }}
              displayName: Run Tests
              env:
                ..
                {{ if parameters.nightly_build }}:
                  SGXLKL_NIGHTLY_BUILD: 1
```

PR #625, opened 2020-07-13
LTP: fix oom error in mknod07 test case

shaikshavali1:
One of the test cases is to create a special file in read-only fs.
To perform this subtest case, it mounts a file system in read-only mode.
There is an out of memory error while it creating the loop device and mount.

This subtest case is removed.


hukoyu (approved):
LGTM

PR #631, opened 2020-07-14
LTP: Fix test case splice02 failure

shaikshavali1:
As per the man page, one of the file descriptor passed
to splice system call should refer to a pipe else it will generate
EINVAL error.
In this test case, both file descriptors passed to splice
system call referred to the file. This test case is modified
to pass one pipe and one file descriptors to splice system call.


hukoyu (approved):
LGTM

PR #632, opened 2020-07-15
LTP: Fix setregid02 test cases failure

shaikshavali1:
"nobody" user cannot read other user details from /etc/passwd file.
code related to setting UID & GID to "nobody" is moved after getting
group details of other users.


hukoyu (approved):
LGTM

PR #636, opened 2020-07-17
LTP: fixed chroot02 testcase

vamsikrishna935:
Removed the wait in the parent as child process is not available.
 Tests were failing to delete the temporary directory created for testing.
 It is happening because after doing chroot on temp directory, the path will be changed and 
 we cannot exit from chroot because child process is not available.
 so created a temp directory for testing with fixed name which is being delted everytime before test runs.


tests/ltp/patches/ltp_chroot_chroot02_fix.patch
```diff
@@ -0,0 +1,56 @@
...
+diff --git a/testcases/kernel/syscalls/chroot/chroot02.c b/testcases/kernel/syscalls/chroot/chroot02.c
+index e483ca4b5..11a7217f4 100644
+--- a/testcases/kernel/syscalls/chroot/chroot02.c
++++ b/testcases/kernel/syscalls/chroot/chroot02.c
+@@ -84,7 +84,7 @@ int main(int ac, char **av)
+                if (pid == 0) {
+                        retval = 0;
+
+-                       if (chroot(tst_get_tmpdir()) == -1) {
++                       if (chroot("/tmp/chroot02_temp_dir") == -1) {
```
shaikshavali1: Could you please define a macro for a temp directory name?

shaikshavali1 (changes requested):
@vamsikrishna935, Please use the macro for temp dir name.

hukoyu (approved):
LGTM

PR #637, opened 2020-07-17
LTP: fixed readv02 testcase

vamsikrishna935:
Tests will be enabled once git issue 297 is fixed.

### DIFF
--- a/tests/ltp/patches/ltp_readv_readv02_fix.patch
+++ b/tests/ltp/patches/ltp_readv_readv02_fix.patch
@@ -1,0 +1,61 @@
++ Test will be enabled once git isue 297 is fixed.
++ Issue 297: [Tests] lkl_access_ok() should return -1 on invalid access
++ https://github.com/lsds/sgx-lkl/issues/297
+diff --git a/testcases/kernel/syscalls/readv/readv02.c b/testcases/kernel/syscalls/readv/readv02.c
+index aa40e2c7d..d42e72517 100644
+--- a/testcases/kernel/syscalls/readv/readv02.c
++++ b/testcases/kernel/syscalls/readv/readv02.c
+@@ -127,23 +127,23 @@ int main(int ac, char **av)
+                }
+
+ //test2:
+-               l_seek(fd[0], CHUNK * 6, 0);
+-               if (readv(fd[0], (rd_iovec + 6), 3) < 0) {
+-                       if (errno != EFAULT) {
+-                               tst_resm(TFAIL, "expected errno = EFAULT, "
+-                                        "got %d", errno);
+-                       } else {
+-                               tst_resm(TPASS, "got EFAULT");
+-                       }
+-                       if (memcmp((buf_list[0] + CHUNK * 6),
+-                                  (buf_list[1] + CHUNK * 6), CHUNK * 3) != 0) {
+-                               tst_resm(TFAIL, "Error: readv() partially "
+-                                        "overlaid buf[2]");
+-                       }
+-               } else {
+-                       tst_resm(TFAIL, "Error: readv returned a positive "
+-                                "value");
+-               }
++//             l_seek(fd[0], CHUNK * 6, 0); // TODO: Enable once git issue 297 is fixed
++//             if (readv(fd[0], (rd_iovec + 6), 3) < 0) {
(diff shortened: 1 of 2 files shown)


tests/ltp/patches/ltp_readv_readv02_fix.patch
```diff
@@ -0,0 +1,61 @@
...
+@@ -222,11 +222,7 @@ void setup(void)
+
+        fd[1] = -1;             /* Invalid file descriptor */
+
+-       bad_addr = mmap(0, 1, PROT_NONE,
+-                       MAP_PRIVATE_EXCEPT_UCLINUX | MAP_ANONYMOUS, 0, 0);
+-       if (bad_addr == MAP_FAILED) {
+-               tst_brkm(TBROK, cleanup, "mmap failed");
+-       }
++       bad_addr = 0;
```
shaikshavali1, line 35: How many sub tests using this bad_addr?

davidchisnall:
This is incredibly hard to review.  A load of the diff is commenting things out using line comments instead of block comments, so it is impossible to see whether there are any changes other than commenting the blocks out.  There is no clear explanation of why the block is commented out.  Please use auto-linking to refer to issues rather than 'git issue 297', which is also semantically incorrect: this is an SGX-LKL issue, not a git issue.

We should not be patching tests to pass when they are actually indicating real errors in SGX-LKL.  This is *completely* missing the point of enabling LTP.  Our goal is not to get a high number of LTP passes, our goal is to fix bugs.  

The correct thing to do here is fix #297, not patch the test to work around the fact that #297 exists.  The lkl_access_ok function should return -1 on failure (though we may wish to retain the `sgxlkl_fail` behaviour as something that we can turn on in the attested configuration).

hukoyu:
@davidchisnall  just to add that in the LTP test readv02 there are 4 test cases as seen in the original test code:
https://github.com/linux-test-project/ltp/blob/master/testcases/kernel/syscalls/readv/readv02.c

There are LTP tests that have multiple sub tests. In such cases that most of them passing and some of them failing, we preferred disabling the failing sub test with a TODO note saying the related github issue(bug). We were aiming increasing test coverage by enabling test with passing sub tests and will enable back TODO comments when the blocking bug/issue fixed. This is the judgement call we made and we used this approach in many enabled test patches.  If you search for TODO in ltp/patches folder you can see them.

hukoyu (approved):
LGTM

tests/ltp/patches/ltp_readv_readv02_fix.patch
```diff
@@ -0,0 +1,38 @@
...
+diff --git a/testcases/kernel/syscalls/readv/readv02.c b/testcases/kernel/syscalls/readv/readv02.c
+index aa40e2c7d..cbdf35eb3 100644
+--- a/testcases/kernel/syscalls/readv/readv02.c
++++ b/testcases/kernel/syscalls/readv/readv02.c
+@@ -126,7 +126,7 @@ int main(int ac, char **av)
+                                 "value");
+                }
+
+-//test2:
++/*test2: // TODO: Enable once git issue 297 is fixed
```
davidchisnall, line 13: Please use `#if 0` to disable entire chunks or you potentially trigger nested comment errors on a future update.

tests/ltp/patches/ltp_readv_readv02_fix.patch
```diff
@@ -0,0 +1,38 @@
...
+-               }
++               }*/
+
+ //test3:
+                if (readv(fd[1], (rd_iovec + 9), 1) < 0) {
+@@ -222,11 +222,7 @@ void setup(void)
+
+        fd[1] = -1;             /* Invalid file descriptor */
+
+-       bad_addr = mmap(0, 1, PROT_NONE,
```
davidchisnall, line 30: Why is this disabled?  What is the value of `MAP_PRIVATE_EXCEPT_UCLINUX`?  

This appears to be attempting to map one byte of anonymous memory.  This should work, but at a guess it is failing because we are not doing anonymous mappings when `fd` is not `-1`.  This looks like an actual bug and we should be tracking it.  Where is the issue associated with this?

davidchisnall (commented):
Please submit a follow-on PR addressing the issues in this.

PR #641, opened 2020-07-17
Remove bad munmap03 LTP test patch

SeanTAllen:
The patch is wrong. It hides a bug in munmap. On failure, munmap is
supposed to return -1. The patch as noted in its comment allows for
0 aka success to be returned and assumes because the errno was set
correctly, that everything is correct.

It isn't. The return value has to be -1 when failure occurs. Anyone
using this munmap interface would get surprising results and the test
patch hides that. munmap is currently broken.


SeanTAllen:
I've created issue #642 specifically for the munmap implementation being incorrect. I'll be working on fixing that in the very near-term.

davidchisnall (approved):
Looks sensible.  We shouldn't run invalid tests, we should track (and fix) the issues that cause them to fail.

hukoyu:
@SeanTAllen  @davidchisnall  I am just curious why didn;t you fix the test instead of deleting? 

cc @paulcallen  @jxyang

davidchisnall:
Having a test that expects the wrong result is actively harmful and prevents fixing the underlying bug.  The fix for the underlying bug will also reenable the test.

PR #647, opened 2020-07-20
LTP: fixed truncate03 testcase

vamsikrishna935:
Modified the tests by changing the temporary directory creation/deletion with root user.

### DIFF
--- a/tests/ltp/patches/ltp_truncate_truncate03_fix.patch
+++ b/tests/ltp/patches/ltp_truncate_truncate03_fix.patch
@@ -1,0 +1,49 @@
++ Modified the tests by changing the temporary directory creation/deletion with root user.
++ Issue 297: [Tests] lkl_access_ok() should return -1 on invalid access
++ https://github.com/lsds/sgx-lkl/issues/297
+diff --git a/testcases/kernel/syscalls/truncate/truncate03.c b/testcases/kernel/syscalls/truncate/truncate03.c
+index 401c16a8e..316fb882c 100644
+--- a/testcases/kernel/syscalls/truncate/truncate03.c
++++ b/testcases/kernel/syscalls/truncate/truncate03.c
+@@ -76,7 +76,7 @@ static struct test_case_t {
+ } test_cases[] = {
+        { TEST_FILE1, TRUNC_LEN, EACCES },
+        { TEST_FILE2, TRUNC_LEN, ENOTDIR },
+-       { NULL, TRUNC_LEN, EFAULT },
++//     { NULL, TRUNC_LEN, EFAULT }, TODO: Enable once git issue 297 is fixed
+        { long_pathname, TRUNC_LEN, ENAMETOOLONG },
+        { "", TRUNC_LEN, ENOENT },
+        { TEST_DIR1, TRUNC_LEN, EISDIR },
+@@ -121,14 +121,12 @@ void setup(void)
+        tst_sig(NOFORK, DEF_HANDLER, cleanup);
+
+        tst_require_root();
+-
++       tst_tmpdir();
+        ltpuser = SAFE_GETPWNAM(cleanup, "nobody");
+        SAFE_SETEUID(cleanup, ltpuser->pw_uid);
+
+        TEST_PAUSE;
+
+-       tst_tmpdir();
+-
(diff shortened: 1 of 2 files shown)


tests/ltp/patches/ltp_truncate_truncate03_fix.patch
```diff
@@ -0,0 +1,49 @@
...
++//     { NULL, TRUNC_LEN, EFAULT }, TODO: Enable once git issue 297 is fixed
+        { long_pathname, TRUNC_LEN, ENAMETOOLONG },
+        { "", TRUNC_LEN, ENOENT },
+        { TEST_DIR1, TRUNC_LEN, EISDIR },
+@@ -121,14 +121,12 @@ void setup(void)
+        tst_sig(NOFORK, DEF_HANDLER, cleanup);
+
+        tst_require_root();
+-
++       tst_tmpdir();
```
SeanTAllen, line 22: why does this need to be moved earlier in the execution?

SeanTAllen: is this related to having it created by the root user?

tests/ltp/patches/ltp_truncate_truncate03_fix.patch
```diff
@@ -0,0 +1,49 @@
...
+-       tst_tmpdir();
+-
+        SAFE_TOUCH(cleanup, TEST_FILE1, NEW_MODE, NULL);
+
+        SAFE_TOUCH(cleanup, "t_file", FILE_MODE, NULL);
+@@ -154,7 +152,7 @@ void setup(void)
+
+        for (n = 0; n < TST_TOTAL; n++) {
+                if (!test_cases[n].pathname)
+-                       test_cases[n].pathname = tst_get_bad_addr(cleanup);
```
SeanTAllen, line 37: what does this change do? I don't know the test very well, but it seems like this would make it not do much of anything for test.

can you explain what is going on?

tests/ltp/patches/ltp_truncate_truncate03_fix.patch
vamsikrishna935: nobody user is failing to create directory under /tmp, with root we can create. As it does not effect test purpose so move it to earlier.

rsrinivas01 (approved):
Changes look fine to me

tests/ltp/patches/ltp_truncate_truncate03_fix.patch
vamsikrishna935: @SeanTAllen Here tst_get_bad_addr will get you bad address using mmap. As mmap is not supported the test fails here, so replaced the bad address with 0 ( as anything other than good is a bad address) and is serving our purpose.

hukoyu (approved):
LGTM

tests/ltp/patches/ltp_truncate_truncate03_fix.patch
SeanTAllen: Sorry, what do you mean "mmap is not supported"? mmap is supported. We use it in many places, we have mmap ltp tests that run.

SeanTAllen: Why this change was done isn't documented as part of the patch. Can you do a PR to update the comment to explain this change?

SeanTAllen: Can you please let me know what "mmap is not supported" means in this context?

bodzhang: mmap() is partially supported. The [compatibility doc](https://github.com/lsds/sgx-lkl/blob/oe_port/docs/Incompatibilities.md) has a high level description. Details on the exact limitation of each of the mmap() API parameters would be even more helpful. @SeanTAllen, do you have that level of details?

@vamsikrishna935 , does the high level description in the compatibility doc give me enough info to analyze "tst_get_bad_addr(cleanup)"?

SeanTAllen: I'm actively working on fixing mmap bugs in sgx-lkl. Yes, I would like the lower level details of what "mmap is not supported" means in this case as it might be indicative of an mmap bug rather than a fundamental  incompatibility.

PR #635, opened 2020-07-17
Adding credentials as env variables for remote attestation test

hukoyu:
We need some credentials for remote attestation scenario tests.
These secrets are defined on ADO and stored encrypted as variable group in library and used in pipeline. 

With this PR we are mapping these credentials to some environment variables on the running VM so that can be consumed by scenario tests. 

These credentials as env variables are will be needed for several scenario tests that will be added soon.

PR #650, opened 2020-07-21
[Test] MAA - Remote Attestation Test Scenario

hukoyu:
This PR depends on https://github.com/lsds/sgx-lkl/pull/635 to be first merged for credentials for AAD and MAA services to be available during nightly run.

This PR is adding a complex test scenario for remote attestation with MAA which requires trusted TLS conenction (cert_type =0 is used)
In this scenario we have two enclaves: sgx-lkl enclave and oe-enclave. oe-enclave is the TLS server and sgx-lkl enclave establishes a trusted connection to oe-enclave and then verifies the oe-enclave's report against MAA with remote attestation.  

Exiting tests in sgx-lkl have one enclave. This scenario test has two enclaves and some tricks used to make this scenario test work with existing test framework such as : 
1) Used a shared log file for stdout and stderr logs of both enclaves
2) OE Enclave runs as a backround process and exists after 45 seconds to give sgx-lkl enclave enough time to connect and verify
3) OE Enclave depends on OE SDK, that's why OE sdk is installed if doesn't exits under /opt/openenclave
4) makefile (with lower case 'M') is used for nested make files so that they are not treated as a separate test file by sgx-lkl test framework.  
5) While reviewing start from Makefile and run_scenario.sh 

Notes: 
1) We keep credentials in ADO as encrypted. These AAD and MAA credentials mapped to environment variables. Since we don't want to enable these env variables for PRs submitted from forks for security reasons, we will run this scenario test in nightly builds only .This test is verified many times localy that works as expected when required env variables available. Also verified with sgx-lkl nightly pipeline as well. Passes for HW but fails for SW. Trying to get answer if SW failure is expected or not. 

2) There are 48 files in this PR. This looks a lot but in this scenario we have sgx-lkl enclave, oe enclave, trusted TLS connection, cert verify, AAD bearer token fetch, MAA attestation, JWT token parse code, ... exists and this is a complex  test scenario and this is the reason for so many files.


letmaik:
If I understand correctly, then this test verifies an attested TLS certificate coming from an OE enclave. It does that by copying a bunch of private sources from OE that do the validation. My main question would be: Why do we need to do that? Shouldn't OE provide cert validation as part of its "hostverify" library? If not, why don't we fix it there first?
(cc @ad-l)

wintersteiger:
In general adding this type of test is a good idea, but I think this needs a good bit of cleanup. We can't copy OE source/header files into the repo; OE is always available during every build, we should either use the existing one, or, if we need a custom version for this test, a fresh copy as a second OE submodule or a `git clone`.  A second copy of OE is perfectly fine, it's truly the application that requires an accessible copy of OE and the existing OE is behind the kernel, i.e. not accessible by the application. (Same for mbedTLS, this is completely in the application layer/user space.) All of `maa/oe_enclave` is a pure OE enclave, it shouldn't have any references to SGX-LKL and it doesn't need to use SGX-LKL's dependencies or third-party libraries.

Conversely, `maa/sgxlkl_enclave` is an SGX-LKL _application_, that means an unmodified Linux application that now runs in a container, i.e. it shouldn't have any references to OE, except the host verification library and it should use it's own copy of OE for that too (in future the host verification library will hopefully be a self-contained library). It seems to me that `_oe_verify_attestation_certificate` is a copy of OE's existing `oe_verify_attestation_certificate`, which is in the host verification library. We should definitely use the existing version and fix it (if necessary) as @letmaik said.

The call to `oe_generate_attestation_certificate` may have to run behind the kernel and we should set that up correctly layered right away, i.e. generate a key pair and a quote and put them into auxv, where the application can pick it up later. 

I'm not sure whether `_maa_verify_report` is needed at all. Shouldn't it be possible to verify the quote without an external service? In any case, I think this should all go into `enclave_identity_callback`, so that OE's `oe_verify_attestation_certificate` runs it.

`verifier` in `oe_enclave/enc/server.cpp` seems rather incomplete, that should be extended.

tests/remote_attestation/maa/common/curl_helper.cpp
```diff
@@ -0,0 +1,178 @@
...
+
+        // Reponse reader callback
+        curl_error =
+            curl_easy_setopt(curl, CURLOPT_WRITEFUNCTION, response_data_cb);
+        goto_if_curl_error(
+            curl_error, "Failed to set WRITEFUNCTION option", error);
+        curl_error = curl_easy_setopt(curl, CURLOPT_WRITEDATA, bufdata);
+        goto_if_curl_error(curl_error, "Failed to set WRITEDATA option", error);
+
+        /* Perform the request, res will get the return code */
```
Pengpeng-Microsoft: Style: use "//" for comments to be consistent.
Change the comment to be: "Perform curl request and set response code per response."

tests/remote_attestation/maa/common/gencreds.cpp
```diff
@@ -0,0 +1,254 @@
...
+    if (public_key)
+        oe_free_key(public_key, public_key_size, NULL, 0);
+
+    if (cert)
+        oe_free_attestation_certificate(cert);
+
+    return result;
+}
+
+int generate_attested_credentials(
```
Pengpeng-Microsoft: I could not find usage of this function in your changes. Can you remove it and all the functions in its call chains that are not referenced elsewhere?

tests/remote_attestation/maa/common/gencreds.cpp
```diff
@@ -0,0 +1,254 @@
+#include "gencreds.h"
```
Pengpeng-Microsoft: See my comments below for generate_attested_credentials. My hunch is that this whole file is not needed. Can you check? Thanks!

tests/remote_attestation/maa/common/gencreds.h
```diff
@@ -0,0 +1,21 @@
+#ifndef _OE_GENCREDS_H
```
Pengpeng-Microsoft: Can you check if this file is needed? Thanks!

tests/remote_attestation/maa/common/json_utils.c
```diff
@@ -0,0 +1,181 @@
...
+    case '{':
+        end_ch = '}';
+    default:
+        // for payload w/o enclosing symbols, just set the start, end to empty
+        start_ch = '\0';
+    }
+    char* start_pos = NULL;
+    int value_len = 0;
+    if (start_ch == '\0')
+    {  // handle situation w/o encloser, e.g. product-id
```
Pengpeng-Microsoft: Please remove ", e.g. product-id".

tests/remote_attestation/maa/common/message.h
```diff
@@ -0,0 +1,83 @@
+#ifndef CLIENT_COMMON_H
```
Pengpeng-Microsoft: I could not find references to the struct's in this header file. Can you check if this header file is needed?

tests/remote_attestation/maa/common/settings.c
```diff
@@ -0,0 +1,52 @@
...
+    const size_t buf_size = 128;
+    char str[buf_size];
+    int res = 0;
+
+    if (name == NULL)
+    {
+        goto fail;
+    }
+
+    if (getenv(name) == NULL)
```
Pengpeng-Microsoft: We should remove this if section entirely. Otherwise, if env variables do not exist, it will halt the pipeline waiting for user input into console.

tests/remote_attestation/maa/oe_enclave/enc/private.pem
```diff
@@ -0,0 +1,39 @@
+-----BEGIN RSA PRIVATE KEY-----
```
Pengpeng-Microsoft: Do you need this file? I could not find reference.

tests/remote_attestation/maa/oe_enclave/enc/public.pem
```diff
@@ -0,0 +1,39 @@
+-----BEGIN RSA PRIVATE KEY-----
```
Pengpeng-Microsoft: Do you need this file? I could not find reference.

tests/remote_attestation/maa/oe_enclave/enc/server.cpp
```diff
@@ -0,0 +1,296 @@
+#include <mbedtls/certs.h>
```
Pengpeng-Microsoft: Can you merge this file with tlssrv.cpp as we discussed?

Pengpeng-Microsoft (changes requested):

jxyang:
> If I understand correctly, then this test verifies an attested TLS certificate coming from an OE enclave. It does that by copying a bunch of private sources from OE that do the validation. My main question would be: Why do we need to do that? Shouldn't OE provide cert validation as part of its "hostverify" library? If not, why don't we fix it there first?
> (cc @ad-l)

`oe_private` and `oe_public` give us headers for handling self-signed certificates. The actual functionality is implemented in `oe_private/cert.c`. I agree we should expose a function in oe hostverify that extract public key and extension (in this case, a quote) from the self-signed certificate. That should clean up this PR significantly.

jxyang:
> In general adding this type of test is a good idea, but I think this needs a good bit of cleanup. We can't copy OE source/header files into the repo; OE is always available during every build, we should either use the existing one, or, if we need a custom version for this test, a fresh copy as a second OE submodule or a `git clone`. A second copy of OE is perfectly fine, it's truly the application that requires an accessible copy of OE and the existing OE is behind the kernel, i.e. not accessible by the application. (Same for mbedTLS, this is completely in the application layer/user space.) All of `maa/oe_enclave` is a pure OE enclave, it shouldn't have any references to SGX-LKL and it doesn't need to use SGX-LKL's dependencies or third-party libraries.
> 
> Conversely, `maa/sgxlkl_enclave` is an SGX-LKL _application_, that means an unmodified Linux application that now runs in a container, i.e. it shouldn't have any references to OE, except the host verification library and it should use it's own copy of OE for that too (in future the host verification library will hopefully be a self-contained library). It seems to me that `_oe_verify_attestation_certificate` is a copy of OE's existing `oe_verify_attestation_certificate`, which is in the host verification library. We should definitely use the existing version and fix it (if necessary) as @letmaik said.
> 
> The call to `oe_generate_attestation_certificate` may have to run behind the kernel and we should set that up correctly layered right away, i.e. generate a key pair and a quote and put them into auxv, where the application can pick it up later.
> 
`oe_generate_attestation_certificate` is called from oe enclave side only. We don't need to worry about it in sgx-lkl for this test case.

> I'm not sure whether `_maa_verify_report` is needed at all. Shouldn't it be possible to verify the quote without an external service? In any case, I think this should all go into `enclave_identity_callback`, so that OE's `oe_verify_attestation_certificate` runs it.

We had some discussions of using MAA vs. using OE hostverify before. There were some obstacles in using hostverify IIRC. If we plan to submit a PR related to hostverify, maybe we could overcome the obstacles and ditch MAA.

> 
> `verifier` in `oe_enclave/enc/server.cpp` seems rather incomplete, that should be extended.

jxyang:
To summarize the possible reductions of complexities in the test case suggested by Maik and Christoph:

* Explore moving functionalities into OE  hostverify, meaning submit a PR to OE
* Explore whether MAA could be replaced with OE hostverify

In addition, The TLS server doesn't need to verify the TLS client in this test case, thus consider getting rid of:
* `verifier` in `oe_enclave/enc/Server.cpp`
* `sgxlkl_cert.der` and `sgxlkl_private_key.pem` in `sgxlkl_enclave`, instead, use the default certificate from mbedtls
* message exchanges after tls connection is established. This means no need for `clientAgent_msg.*` and `common/message.h`.
* `common/settings.*` because we can simply call `getenv`
* other files under `common` that are no longer needed after the above clean ups.

I will do a line-by-line review after the clean up is done.

tests/remote_attestation/maa/oe_enclave/enc/private.pem
hukoyu: Yes this is needed for signing oe_enclave app.
oe_enclave/enc/makefile:   oesign sign -e oeApp_enc -c enc.conf -k private.pem

tests/remote_attestation/maa/oe_enclave/enc/server.cpp
hukoyu: I think it will be better to keep them separate for readability and maintainability

tests/remote_attestation/maa/oe_enclave/enc/public.pem
hukoyu: This is the matching public.pem which is used for private.pem for signing oe_enclave app. I also didn't find code  using it directly. Should we keep it just in case.

tests/remote_attestation/maa/common/settings.c
hukoyu: Removed.

tests/remote_attestation/maa/common/json_utils.c
hukoyu: Done

tests/remote_attestation/maa/common/gencreds.h
hukoyu: Yes this is used by oe_enclave/enc/tlssrv.cpp

tests/remote_attestation/maa/common/gencreds.cpp
hukoyu: Yes this is used by oe_enclave/enc/tlssrv.cpp

tests/remote_attestation/maa/common/message.h
hukoyu: Actually Query_t and QueryResult_t and some constants are used on bith sgx-lkl and oe enclave codes. I will look if I can remove them but they are referenced ad used at the moment.

tests/remote_attestation/maa/common/gencreds.cpp
hukoyu: This is used in oe_enclave/enc/tlssrv.cpp

tests/remote_attestation/maa/common/curl_helper.cpp
hukoyu: Done

tests/remote_attestation/maa/oe_enclave/enc/private.pem
wintersteiger: We really shouldn't check in any private/public keys if they are not necessary. The SGX-LKL build already sets up a random key pair that can be reused, or the key generation can be duplicated to generate different ones.

tests/remote_attestation/maa/oe_enclave/enc/public.pem
wintersteiger: When in doubt, remove unnecessary key files, don't keep them "just in case", you can easily generate new key pairs.

hukoyu: Verified that this file is not needed. Removed it.

tests/remote_attestation/maa/oe_enclave/enc/private.pem
hukoyu: Checked this with @jxyang  as well. private.pem is private key for signing oe_enclave app to be verified by sgxlkl_enclave against a MR signer. If we generate random key pairs, the test and verification will fail. That is why we need a preknown MRSIGNER and need to sign oe_enclave with matching private key (private.pem) so that verification can pass and TLS connection can be established. So this is required.

hukoyu:
> To summarize the possible reductions of complexities in the test case suggested by Maik and Christoph:
> 
> * Explore moving functionalities into OE  hostverify, meaning submit a PR to OE
> * Explore whether MAA could be replaced with OE hostverify
> 
> In addition, The TLS server doesn't need to verify the TLS client in this test case, thus consider getting rid of:
> 
> * `verifier` in `oe_enclave/enc/Server.cpp`
> * `sgxlkl_cert.der` and `sgxlkl_private_key.pem` in `sgxlkl_enclave`, instead, use the default certificate from mbedtls
> * message exchanges after tls connection is established. This means no need for `clientAgent_msg.*` and `common/message.h`.
> * `common/settings.*` because we can simply call `getenv`
> * other files under `common` that are no longer needed after the above clean ups.
> 
> I will do a line-by-line review after the clean up is done.

Made this cleanup and submitted/pushed changes based on @jxyang 's comments:
1) Simplified verifier in oe_enclave which returns true always
2) Removed clientAgent_msg.* code and related codes
3) Got rid of common/settings.* and common/message.h
4) As we discussed with Xuejun removing sgxlkl_cert.der and sgxlkl_private_key.pem in sgxlkl_enclave will be complicated, we will just keep them.
@jxyang  can you review please? Thanks

tests/remote_attestation/maa/common/message.h
hukoyu: common/message.h and common/settings.* removed totally.

hukoyu:
@letmaik  I ran sgx-lkl-nightly pipeline manually as you recommended
https://dev.azure.com/sgx-lkl/sgx-lkl/_build/results?buildId=1455&view=results 
But it doesn't have my MAA Remote Attestation test since my PR is not merged yet. is there another trick I need to do?

letmaik:
@hukoyu You triggered the right pipeline but I think you didn't select your PR branch, so it used the default branch `oe_port` instead.

hukoyu:
@letmaik I always submit my PRs from my fork. Not from branches on this repo. Should I resubmit this PR from a branch in this repo for this manual test pipeline to work?

letmaik:
@hukoyu Right, I didn't think of that. You can still do it with a trick and keep your branch in your fork. For the branch name, use `refs/pull/650/merge`. This is a special git ref that GitHub creates for each PR.

hukoyu:
Thanks @letmaik  I am able to run sgx-lkl nightly pipeline for my PR with refs/pull/650/merge

hukoyu:
@davidchisnall @prp @letmaik @mikbras 
I have a question about passing environment variables to sgx-lkl enclave.
We are already passing env variables like SGXLKL_ETHREADS to sgx-lkl enclave and it is available in it, 
For MAA remote attestation test scenario I am trying to pass a few env variables like MAA_CLIENT_ID. I am verifying env variables are available in VM and is available in makefile calling the sgx-lkl-run-oe , when I run it in my personal VM it is working fine, but in sgx-lkl (nightly) pipeline, env variable doesn't reach to sgx-lkl enclave. It is NULL. 
I tried adding to Dockerfile with ENV, passing in line like "MAA_CLIENT_ID=vaue sgx-lkl-run-oe ..." 
nothing works in pipeline. 
Why the result is different in pipeline than local run, and why env variables not available in sgx-lkl enclave. What am I missing? I need your help. Thanks
cc  @paulcallen  @Pengpeng-Microsoft  @jxyang

hukoyu:
When I run the nightly build for my PR 
in sgxlkl_enclave/sgxLklApp.c I get this error:
it throws exception since dependent env variables are empty/NULL.

sgxLklApp: Lack the credential(s) to run. Hint: you must set environment variables for MAA_CLIENT_ID, and MAA_CLIENT_SECRET

hukoyu:
> When I run the nightly build for my PR
> in sgxlkl_enclave/sgxLklApp.c I get this error:
> it throws exception since dependent env variables are empty/NULL.
> 
> sgxLklApp: Lack the credential(s) to run. Hint: you must set environment variables for MAA_CLIENT_ID, and MAA_CLIENT_SECRET

Thanks to @praenubilus  , I learned that after recent change I need to set SGXLKL_HOST_IMPORT_ENV to host environment variables I need to pass to sgx-lkl enclave. After setting this now sgx-lkl enclave gets the needed env variables for credentials. Also the reason it was working locally because I was using an old version of /opt/sgx-lkl-debug (dated July 9th). After installing latest debian package I was ablet to see issue locally as well.

hukoyu:
Ran the sgx-lkl nightly build for this PR:
https://tfsproduks1.visualstudio.com/Ab5352d9a-5ed5-48f2-b767-7f273b55b53c/sgx-lkl/_build/results?buildId=1498&view=results

Test passes on hw mode for debug and nonrelease. However, fails for sw mode for both debug and nonrelease.
@jxyang @Pengpeng-Microsoft  @praenubilus  @triston996  Is TLS self signed certificate supposed to work on sw mode? Or does it only work for hw mode? Is this expected result?

Results is:
core hw debug 1 thread: PASS
core hw debug 4 threads: PASS
core hw debug 8 threads: PASS

core hw nonrelease 1 thread: PASS
core hw nonrelease 4 threads: PASS
core hw nonrelease 8 threads: PASS

core sw debug 1 thread: FAIL
core sw debug 4 threads: FAIL
core sw debug 8 threads: FAIL

core sw nonrelease 1 thread: FAIL
core sw nonrelease 4 threads: FAIL
core sw nonrelease 8 threads: FAIL

Failure log:
```
SGXLKL_VERBOSE=0 SGXLKL_GETTIME_VDSO=0 SGXLKL_TAP=sgxlkl_tap0 "/opt/sgx-lkl-debug/bin"/sgx-lkl-run-oe  --sw-debug sgxlkl-image.img /sgxLklApp 10.0.1.254  >> /agent/_work/1/s/remote_attestation_maa_test.log 2>&1
makefile:29: recipe for target 'lkl-run' failed
make[1]: *** [lkl-run] Error 139
make[1]: Leaving directory '/agent/_work/1/s/tests/remote_attestation/maa/sgxlkl_enclave'
Mon Jul 27 06:40:04 UTC 2020
Host: Enclave path = ./enc/oeApp_enc.signed, server port = 13999
2020-07-27T06:40:05.000000Z [(H)ERROR] tid(0x7ff617c68780) | enclave_create with ENCLAVE_TYPE_SGX1 type failed (err=0x1) (oe_result_t=OE_PLATFORM_ERROR) [../host/sgx/sgxload.c:oe_sgx_create_enclave:480]
2020-07-27T06:40:05.000000Z [(H)ERROR] tid(0x7ff617c68780) | :OE_PLATFORM_ERROR [../host/sgx/create.c:oe_sgx_build_enclave:816]
2020-07-27T06:40:05.000000Z [(H)ERROR] tid(0x7ff617c68780) | :OE_PLATFORM_ERROR [../host/sgx/create.c:oe_create_enclave:964]
Host: Enclave creation failed
Host: oe_enclave app exited successfully
Client Agent failed! tlscli_connect with cert file: ./sgxlkl_cert.der and pkey file: ./sgxlkl_private_key.pem
Segmentation fault
[    0.105475] reboot: Restarting system
TEST FAILED
Makefile:6: recipe for target 'run-sw' failed
make: *** [run-sw] Error 1
make run-sw: FAILED WITH EXIT CODE: 2
Test run completed with EXIT CODE 2
----------output-end-------------
Failure : 'TEST FAILED' observed in 'report/remote_attestation-maa-(debug)-(run-sw)-(1-ethreads).stdout.txt'
Failure : 'Segmentation fault' observed in 'report/remote_attestation-maa-(debug)-(run-sw)-(1-ethreads).stdout.txt'
```
cc @letmaik @prp @davidchisnall @mikbras @paulcallen

tests/remote_attestation/maa/common/auth.cpp
```diff
@@ -0,0 +1,82 @@
+#include "auth.h"
```
jxyang: This file should be shared by all remote attestation test cases. Suggest putting it under `tests/remote_attestation/common`.

tests/remote_attestation/maa/common/curl_helper.cpp
```diff
@@ -0,0 +1,178 @@
+#include "curl_helper.h"
```
jxyang: same suggestion as auth.cpp and auth.h

tests/remote_attestation/maa/common/gencreds.cpp
jxyang: gencreds.cpp and gencreds.h should be moved to `tests/remote_attestation/maa/oe_enclave`, where it is exclusively used.

tests/remote_attestation/maa/common/json_utils.c
```diff
@@ -0,0 +1,181 @@
+#include <stdio.h>
```
jxyang: Suggest putting it under `tests/remote_attestation/common`.

tests/remote_attestation/maa/common/log_utils.h
```diff
@@ -0,0 +1,26 @@
+#ifndef LOG_UTILS_H
```
jxyang: Don't see the function `log_hex_data` used any where. Suggest removing log_utils.c.

tests/remote_attestation/maa/common/maa.cpp
```diff
@@ -0,0 +1,109 @@
+#include "maa.h"
```
jxyang: Confusing file name. Suggest `maa_helper.cpp` or `maa_utils.cpp`.

tests/remote_attestation/maa/common/transport_utils.c
```diff
@@ -0,0 +1,266 @@
+#include "transport_utils.h"
```
jxyang: Suggest putting it under `tests/remote_attestation/common`.

tests/remote_attestation/maa/oe_enclave/host/makefile
```diff
@@ -0,0 +1,38 @@
...
+endif
+
+ifeq ($(USE_GCC),)
+        CC = clang-7
+        CXX = clang++-7
+        C_COMPILER=clang
+        CXX_COMPILER=clang++
+endif
+
+CFLAGS=$(shell pkg-config oehost-$(C_COMPILER) --cflags)
```
jxyang: This requires an OE installation. Can we just use the settings from the OE submodule in SGX-LKL?

tests/remote_attestation/maa/run_scenario.sh
```diff
@@ -0,0 +1,117 @@
...
... 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds 120 seconds for connection from sgxlkl_enclave.
```
jxyang: Do we need all the `120 seconds` repeated N times? :-)

tests/remote_attestation/maa/run_scenario.sh
```diff
@@ -0,0 +1,117 @@
...
+# sgxlkl_enclave starts and connects to oe_enclave
+# sgxlkl_enclave gets the report from oe_enclave and then gets authentication token from AAD
+# sgxlkl_enclave calls MAA for JWT token and gets, parses and verifies it
```
jxyang: AAD is a prerequisite for calling MAA. Suggesting `then gets authentication token from AAD before calling MAA for verification. sgxlkl_enclave then parses and verifies the JWT token returned from MAA`

tests/remote_attestation/maa/run_scenario.sh
```diff
@@ -0,0 +1,117 @@
...
+    echo "Dependent environment variables for this scenario not available."
+    echo "SGXLKL_ROOT: $SGXLKL_ROOT"
+    echo "SGXLKL_PREFIX: $SGXLKL_PREFIX"
+    echo "SGXLKL_RUN_MODE: $SGXLKL_RUN_MODE"
+    exit 1
+fi
+
+export SGXLKL_HOST_IMPORT_ENV="MAA_CLIENT_ID,MAA_CLIENT_SECRET,MAA_APP_ID,MAA_ADDR,MAA_ADDR_APP"
+
+# If openenclave is not installed, install it
```
jxyang: I think we should use the OE submodule instead of forcing a build of OE from scratch.

tests/remote_attestation/maa/run_scenario.sh
```diff
@@ -0,0 +1,117 @@
...
+. /opt/openenclave/share/openenclave/openenclaverc
+
+# Set log file that will be shared between two enclaves in the test scenario for stdout and stderr logs
+# In sgx-lkl test framework can have one log file for stdout and stderr. This log file is called with cat to be included in test's output
+# This file is also searched for some key sentences to verify TLS connection and MAA attestation results
+export MAA_TEST1_LOG_FILE="$SGXLKL_ROOT/remote_attestation_maa_test.log"
+date > "$MAA_TEST1_LOG_FILE"
+
+# Kill hanging processes if exist
+process="sgx-lkl-run-oe"
```
jxyang: This should belong to one of the test scripts, and be available to all test cases.

tests/remote_attestation/maa/run_scenario.sh
```diff
@@ -0,0 +1,117 @@
...
+    # shellcheck disable=SC2009
+    ps -aux | grep $process
+    echo "Trying to kill hanging $process process"
+    sudo pkill -9 $process
+    # pkill does not block.
+    sleep 5
+fi
+
+process="oeApp_host"
+if pgrep -x $process >/dev/null; then
```
jxyang: This should belong to one of the test scripts, and be available to all test cases. The process name could be an argument.

tests/remote_attestation/maa/run_scenario.sh
```diff
@@ -0,0 +1,117 @@
...
+    cd "$home" || exit 1
+fi
+
+# shellcheck disable=SC1091
+. /opt/openenclave/share/openenclave/openenclaverc
+
+# Set log file that will be shared between two enclaves in the test scenario for stdout and stderr logs
+# In sgx-lkl test framework can have one log file for stdout and stderr. This log file is called with cat to be included in test's output
+# This file is also searched for some key sentences to verify TLS connection and MAA attestation results
+export MAA_TEST1_LOG_FILE="$SGXLKL_ROOT/remote_attestation_maa_test.log"
```
jxyang: We probably shouldn't put the log under the root. Suggegst `$SGXLKL_ROOT/report`

tests/remote_attestation/maa/sgxlkl_enclave/makefile
```diff
@@ -0,0 +1,30 @@
...
+all: lkl-image
+
+clean:
+	rm -f clientAgent.o alpine_sgxLklApp sgxlkl-image.img*
+
+alpine-container:
+	docker build -t alpine3.10.build ./alpine
+
+alpine-binaries: alpine-container
+	docker run --rm  -v $(CURDIR)/..:/src alpine3.10.build bash -c "$(build-alpine-exe)"
```
jxyang: we can run `$(build-alpine-exe)` in step `lkl-image`, instead of creating a separate docker container for building the executable. This way we can get rid of `alpine-container`, `alpine-binaries`, and the Dockerfile under `alpine`.

tests/remote_attestation/maa/sgxlkl_enclave/oe_private/cert.c
```diff
@@ -0,0 +1,901 @@
+#include <mbedtls/ctr_drbg.h>
```
jxyang: As discussed, the functionalities of this file should ideally be exposed through OE host verify.

tests/remote_attestation/maa/sgxlkl_enclave/oe_private/oe_defs.h
```diff
@@ -0,0 +1,150 @@
+#ifndef OE_DEFS_H
```
jxyang: This file should go away if we expose the cert-related APIs through OE host verify.

tests/remote_attestation/maa/sgxlkl_enclave/oe_private/oe_header.h
```diff
@@ -0,0 +1,211 @@
+#ifndef OE_HEADER_H
```
jxyang: This file should go away if we expose the cert-related APIs through OE host verify.

tests/remote_attestation/maa/sgxlkl_enclave/oe_private/verify_cert.h
```diff
@@ -0,0 +1,12 @@
+#ifndef VERIFY_CERT_H
```
jxyang: This file should go away if we expose the cert-related APIs through OE host verify.

tests/remote_attestation/maa/sgxlkl_enclave/oe_public/bits/defs.h
```diff
@@ -0,0 +1,182 @@
+#ifndef _OE_BITS_DEFS_H
```
jxyang: This file should go away if we expose the cert-related APIs through OE host verify.

tests/remote_attestation/maa/sgxlkl_enclave/oe_public/bits/report.h
```diff
@@ -0,0 +1,207 @@
+/**
+ * @file report.h
```
jxyang: This file should go away if we expose the cert-related APIs through OE host verify.

tests/remote_attestation/maa/sgxlkl_enclave/oe_public/bits/result.h
```diff
@@ -0,0 +1,374 @@
+/**
```
jxyang: This file should go away if we expose the cert-related APIs through OE host verify.

tests/remote_attestation/maa/sgxlkl_enclave/oe_public/bits/types.h
```diff
@@ -0,0 +1,180 @@
+/**
+ * @file types.h
```
jxyang: This file should go away if we expose the cert-related APIs through OE host verify.

tests/remote_attestation/maa/sgxlkl_enclave/tlscli.c
```diff
@@ -0,0 +1,561 @@
...
+        mbedtls_ssl_config_free(&cli->conf);
+        mbedtls_x509_crt_free(&cli->crt);
+        mbedtls_pk_free(&cli->pk);
+        free(cli);
+    }
+
+    return ret;
+}
+
+int tlscli_read(tlscli_t* cli, void* data, size_t size, tlscli_err_t* err)
```
jxyang: I don't think we need tlscli_write or tlscli_read

jxyang (changes requested):

tests/remote_attestation/maa/sgxlkl_enclave/tlscli.c
hukoyu: Removed

tests/remote_attestation/maa/common/log_utils.h
hukoyu: Actually used in sgxlkl_enclave/tlscli.c

tests/remote_attestation/maa/common/gencreds.cpp
hukoyu: Moved to oe_enclave/enc

jxyang:
To show the structural relationship between this test case and Triston's. Consider placing both tests under `tests/attestation`, and this one named `oe_attests_to_sgxlkl`. The other should be named `sgxlkl_attests_to_oe`.

tests/remote_attestation/maa/common/curl_helper.cpp
hukoyu: Moved to test/remote_attestation/common

tests/remote_attestation/maa/common/json_utils.c
hukoyu: Moved to test/remote_attestation/common

tests/remote_attestation/maa/common/auth.cpp
hukoyu: Moved to test/remote_attestation/common

tests/remote_attestation/maa/sgxlkl_enclave/oe_public/bits/defs.h
hukoyu: There is no action item for now. if this happens we can submit a follow up PR. Let's don't block the PR with uncertainity for now.

tests/remote_attestation/maa/sgxlkl_enclave/oe_public/bits/report.h
hukoyu: There is no action item for now. if this happens we can submit a follow up PR. Let's don't block the PR with uncertainity for now.

tests/remote_attestation/maa/sgxlkl_enclave/oe_public/bits/result.h
hukoyu: There is no action item for now. if this happens we can submit a follow up PR. Let's don't block the PR with uncertainity for now.

tests/remote_attestation/maa/sgxlkl_enclave/oe_public/bits/types.h
hukoyu: There is no action item for now. if this happens we can submit a follow up PR. Let's don't block the PR with uncertainity for now.

hukoyu:
@jxyang I applied all of your feedback. Please let me know if you have other feedback or this is ready for merging. @jxyang  also TLS trusted connection fails with SW mode. Passes with HW mode. Is this expected behavior?

hukoyu:
> To show the structural relationship between this test case and Triston's. Consider placing both tests under `tests/attestation`, and this one named `oe_attests_to_sgxlkl`. The other should be named `sgxlkl_attests_to_oe`.

@jxyang , @triston996 's test doesn't do attestation. It just verifies TLS trusted connection but doesn't do atetstation agains MAA at all. I moved my comon files to /tests/remote_attestation/common and used from /tests/remote_attestation/maa/

hukoyu:
@jxyang said that this test must pass for both HW and SW mode. Currently it is passing for HW mode, but failing for SW mode with segmentation faut. Also it passes in local run but fails on pipeline which makes it harder to debug. Need to solve this SW fail issue. Other than that this PR looks in good state.

hukoyu:
Thanks to @letmaik , OE Enclave app in this scenario was always running in HW mode.
```
result = oe_create_oeApp_enclave(
        enclave_path,
        OE_ENCLAVE_TYPE_SGX,
        OE_ENCLAVE_FLAG_DEBUG,
        NULL,
        0,
        &enclave);
```
That's why it was passing locally since my dev VM has SGX harwdare support. And it was failing on pipeline since SW mode tests run on Non-SGX regular VMs. 

To repro this issue locally, I added simulation mode flag to oe_create_*_enclave method and it fails locally as well when we start OE Enclave app.
```
esult = oe_create_oeApp_enclave(
        enclave_path,
        OE_ENCLAVE_TYPE_SGX,
        OE_ENCLAVE_FLAG_DEBUG | OE_ENCLAVE_FLAG_SIMULATE,
        NULL,
        0,
        &enclave);
```
Exception got when we start OE_ENCLAVE with SW/Simulation Mode:
```
Host: Enclave path = ./enc/oeApp_enc.signed, Run mode = sw, server port = 13999
Host: starting_server
makefile:12: recipe for target 'run' failed
make: *** [run] Segmentation fault (core dumped)
```

CONCLUSION: TLS Trusted connection server is OE_ENCLAVE in this test scenario and self-signed certificate is used which depends on SGX Hardware which is not available on Non-SGX VMs. So technically we cannot run this scenario on SW mode. We can run it only for HW mode. 
So solution will be to add a mechansim to mark a test as HW only, SW only , or both HW and SW.

@jxyang @davidchisnall @prp @paulcallen @mikbras @Pengpeng-Microsoft @triston996 @praenubilus any objection to this conclusion and not running this test for SW mode?

jxyang:
Any reason why the OE enclave failed to load in sw mode?

paulcallen:
so why is this actually failing? Is it because there is a call to an OE API that is not supported when run in simulation mode? There are a couple!
What is the new call stack where it is failing?

hukoyu:
@jxyang @paulcallen @triston996 here is the call stack for oe enclave sw mode error:
```
Program received signal SIGSEGV, Segmentation fault.
0x00007fffee181ae9 in sgx_create_report (report_data=0x0, report_data_size=0, target_info=0x0, target_info_size=0, report=0x7fffeec0b388)
    at ../enclave/core/sgx/report.c:133
133         asm volatile("ENCLU"
(gdb) bt
#0  0x00007fffee181ae9 in sgx_create_report (report_data=0x0, report_data_size=0, target_info=0x0, target_info_size=0, report=0x7fffeec0b388)
    at ../enclave/core/sgx/report.c:133
#1  0x00007fffee180885 in _get_default_key_request_attributes (sgx_key_request=0x7fffeec0b688) at ../enclave/core/sgx/keys.c:199
#2  0x00007fffee18033c in _oe_get_seal_key_by_policy_internal (seal_policy=OE_SEAL_POLICY_UNIQUE, key_buffer=0x7fffee40cd10 "", key_buffer_size=0x7fffeec0b9d0,
    key_info=0x7fffee40ce70 "", key_info_size=0x7fffeec0b9c8) at ../enclave/core/sgx/keys.c:252
#3  0x00007fffee17fee6 in oe_get_seal_key_by_policy_v2 (seal_policy=OE_SEAL_POLICY_UNIQUE, _key_buffer=0x7fffeec0bb10, _key_buffer_size=0x7fffeec0bb08,
    _key_info=0x7fffeec0bb00, _key_info_size=0x7fffeec0baf8) at ../enclave/core/sgx/keys.c:343
#4  0x00007fffee03303b in _load_seal_key_by_policy (policy=OE_SEAL_POLICY_UNIQUE, key_buffer=0x7fffeec0bd70, key_buffer_size=0x7fffeec0bd68, key_info=0x0,
    key_info_size=0x0) at ../enclave/asym_keys.c:46
#5  0x00007fffee030ec8 in _load_asymmetric_key_by_policy (policy=OE_SEAL_POLICY_UNIQUE, key_params=0x7fffeec0bec0, is_public=true, key_buffer=0x7fffeec0beb8,
    key_buffer_size=0x7fffeec0beb0, key_info=0x0, key_info_size=0x0) at ../enclave/asym_keys.c:299
#6  0x00007fffee03096a in oe_get_public_key_by_policy (policy=OE_SEAL_POLICY_UNIQUE, key_params=0x7fffeec0bec0, key_buffer=0x7fffeec0beb8,
    key_buffer_size=0x7fffeec0beb0, key_info=0x0, key_info_size=0x0) at ../enclave/asym_keys.c:409
#7  0x00007fffee007a6c in _generate_key_pair(unsigned char**, unsigned long*, unsigned char**, unsigned long*) ()
#8  0x00007fffee0075f9 in _generate_cert_and_private_key(char const*, unsigned long, unsigned char**, unsigned long*, unsigned char**, unsigned long*) ()
#9  0x00007fffee006fdf in generate_attested_credentials ()
#10 0x00007fffee006767 in _get_cert_and_private_key(mbedtls_x509_crt*, mbedtls_pk_context*, _tlssrv_err*) ()
#11 0x00007fffee005e4c in _configure_srv(_tlssrv*, _tlssrv_err*) ()
#12 0x00007fffee005c30 in tlssrv_create(char const*, char const*, _oe_result (*)(void*, unsigned char const*, unsigned long, unsigned char const*, unsigned long, unsigned char const*, unsigned long, unsigned long), void*, _tlssrv**, _tlssrv_err*) ()
#13 0x00007fffee004bdb in setup_tls_server ()
#14 0x00007fffee0081eb in ecall_setup_tls_server ()
#15 0x00007fffee1796ac in oe_handle_call_enclave_function (arg_in=140737488346056) at ../enclave/core/sgx/calls.c:289
#16 0x00007fffee17b73e in _handle_ecall (td=0x7fffeec13000, func=2, arg_in=140737488346056, output_arg1=0x7fffeec0cfc8, output_arg2=0x7fffeec0cfc0)
    at ../enclave/core/sgx/calls.c:417
#17 0x00007fffee17b07f in __oe_handle_main (arg1=281483566645248, arg2=140737488346056, cssa=0, tcs=0x7fffeec0e000, output_arg1=0x7fffeec0cfc8,
    output_arg2=0x7fffeec0cfc0) at ../enclave/core/sgx/calls.c:996
#18 0x00007fffee17c940 in oe_enter () at ../enclave/core/sgx/enter.S:197
#19 0x0000000000490510 in oe_enter_sim (tcs=0x7fffeec0e000, aep=<optimized out>, arg1=140737198992304, arg2=140737198991872, arg3=0x7fffffffda28, arg4=0x7fffffffda20,
    enclave=0x6d5320) at ../host/sgx/enter.c:267
#20 0x000000000048152e in _enter_sim (enclave=0x6d5320, tcs_=0x7fffeec0e000, aep=4784032, arg1=281483566645248, arg2=140737488346056, arg3=0x7fffffffda28,
    arg4=0x7fffffffda20) at ../host/sgx/calls.c:124
#21 0x000000000047fe5b in _do_eenter (enclave=0x6d5320, tcs=0x7fffeec0e000, aep=4784032, code_in=OE_CODE_ECALL, func_in=2, arg_in=140737488346056,
    code_out=0x7fffffffda1c, func_out=0x7fffffffda1a, result_out=0x7fffffffda18, arg_out=0x7fffffffda10) at ../host/sgx/calls.c:189
#22 oe_ecall (enclave=0x6d5320, func=2, arg=140737488346056, arg_out_ptr=0x7fffffffdbc0) at ../host/sgx/calls.c:627
#23 0x000000000045455c in oe_call_enclave_function_by_table_id (enclave=0x6d5320, table_id=18446744073709551615, function_id=0, input_buffer=0x6d5c70,
    input_buffer_size=48, output_buffer=0x6d5ca0, output_buffer_size=32, output_bytes_written=0x7fffffffde20) at ../host/calls.c:86
#24 0x0000000000454a85 in oe_call_enclave_function (enclave=0x6d5320, function_id=0, input_buffer=0x6d5c70, input_buffer_size=48, output_buffer=0x6d5ca0,
    output_buffer_size=32, output_bytes_written=0x7fffffffde20) at ../host/calls.c:124
#25 0x0000000000406711 in setup_tls_server ()
#26 0x0000000000405e96 in main ()
(gdb)
```

wintersteiger:
This segfaults while trying to get a report/quote, it tries to invoke the `ENCLU[EREPORT]` instruction, which is not available outside an enclave. Since this test depends on quotes, I don't think we can run this in software mode.

letmaik:
> 
> 
> This segfaults while trying to get a report/quote, it tries to invoke the `ENCLU[EREPORT]` instruction, which is not available outside an enclave. Since this test depends on quotes, I don't think we can run this in software mode.

I agree. But OE shouldn't even try to do that in simulation mode and rather print a useful error message... @jhand2

wintersteiger:
Yeah, +1!

hukoyu:
As @wintersteiger  and @letmaik confirmed SW mode is not supported. @jxyang recommended a workaround by using predefined cert and private key pair for SW mode on OE Enclave. Made necessary changes but eventhough all mbedtls functions build fine, two functions for reading cert file and private key failing and quite picky : mbedtls_x509_crt_parse_file, mbedtls_pk_parse_keyfile. Working with @jxyang to overcome these build errors and get it working. Will try to solve this issue this week, if can't, I will go with disabling this test for SW mode which is not supported anyways.

hukoyu:
@jxyang @paulcallen @Pengpeng-Microsoft @letmaik @wintersteiger 
mbedtls_x509_crt_parse_file, mbedtls_pk_parse_keyfile functions of mbedtls library is disabled in OE SDK. (https://github.com/openenclave/openenclave/blob/v0.10.x/docs/MbedtlsSupport.md)
Basically SW mode is not supported for this test.

I added a simple solution for skipping  a test for sw or hw mode.
test_runner.sh change for skipping based on skip-run-sw in Makefile
https://github.com/lsds/sgx-lkl/pull/650/commits/daf86e1c859402bd8443a7d02e66ead30cd35954#diff-2c52904eb84fc5c1c90163d977d97d62

How to define skip-run-sw in Makefile
https://github.com/lsds/sgx-lkl/pull/650/commits/daf86e1c859402bd8443a7d02e66ead30cd35954#diff-d0e12ce16b624be61648a533aaa2e5b3

I ran sgx-lkl nightly build (https://dev.azure.com/sgx-lkl/sgx-lkl/_build/results?buildId=1615&view=results) for refs/pull/650/merge and verified that core hw tests passes with 38 tests (including this remote attestation maa test), and core sw tests pass with 37 tests (and remote attestation maa is skipped).

In the run test log you can see this log:
```tests/remote_attestation/maa/Makefile is disabled for run-sw in Makefile. Skipping test... Remove skip-run-sw from Makefile to enable it.```

There are several failures in sgx-lkl nightly pipeline due to known issues and there are github issues about them. I verified this test is passing for HW mode and skipped for SW mode as expected.  This PR is in good state and ready to be approved and merged.

@jxyang @paulcallen @Pengpeng-Microsoft @letmaik @wintersteiger can you approve if looks good and there is no other feedback.

**Comment by letmaik** [comment]
> 
> 
> @jxyang @paulcallen @Pengpeng-Microsoft @letmaik @wintersteiger
> mbedtls_x509_crt_parse_file, mbedtls_pk_parse_keyfile functions of mbedtls library is disabled in OE SDK. (https://github.com/openenclave/openenclave/blob/v0.10.x/docs/MbedtlsSupport.md)
> Basically SW mode is not supported for this test.

What do these two functions have to do with sw mode?

**Comment by letmaik** [comment]
A detailed README for this test is still missing.

**Comment by wintersteiger** [comment]
@letmaik: they are trying to load a pre-built certificate to avoid having to get a quote for a new one, but all the file system functions are disabled. 

You can also load the keys and data from a buffer/string in a `.c` file, e.g. via `mbedtls_x509_crt_parse` and ` mbedtls_pk_parse_key`.

**Comment by hukoyu** [comment]
> A detailed README for this test is still missing.

@letmaik , @SeanTAllen  added README for the test scenario. 
If it looks good I will appriciate if you can approve the PR.

**Comment by hukoyu** [comment]
@Pengpeng-Microsoft , @jxyang I addressed all feedback and this PR is in good state and ready to be merged. You requested changes for this PR and blocked merging :) If everything looks good and don't have anyother feedback can you approve this PR please so that we can merge. Thanks

**Review by Pengpeng-Microsoft — APPROVED** [review]

File: tests/remote_attestation/maa/sgxlkl_enclave/oe_public/bits/report.h (hunk 30 above)
Comment (jxyang) [reply tests/remote_attestation/maa/sgxlkl_enclave/oe_public/bits/report.h]: This is the one of the first two design suggestions I gave. I feel strongly we shouldn't introduce so many oe_private / oe_public files for this test case.

**Review by jxyang — CHANGES_REQUESTED** [review]

**Comment by jxyang** [comment]
> > To show the structural relationship between this test case and Triston's. Consider placing both tests under `tests/attestation`, and this one named `oe_attests_to_sgxlkl`. The other should be named `sgxlkl_attests_to_oe`.
> 
> @jxyang , @triston996 's test doesn't do attestation. It just verifies TLS trusted connection but doesn't do atetstation agains MAA at all. I moved my comon files to /tests/remote_attestation/common and used from /tests/remote_attestation/maa/

An attestation could happen without MAA. @triston996 's test is another form of attestation by verifying MRSIGNER etc. in the enclave. Some of my comments suggested `tests/remote_attestation/common` for sharing files between your test and Triston's test. But now I feel `/tests/attestation/common` is a better name. You can work out the name/path between you two. But the bottom line is that there should be no duplicated files between your test and Trinston's test.

File: tests/remote_attestation/maa/oe_enclave/enc/server.cpp [hunk 34]
```diff
@@ -0,0 +1,110 @@
...
+#define MRSIGNER_SIZE 32
+#define ISVPRODID_SIZE 16
+#define BUFSIZE 32
+
+using namespace std;
+
+extern "C"
+{
+    int setup_tls_server(const char* server_port);
+    int run_server(void* ptr);
```
letmaik: Doesn't exist?

tests/remote_attestation/maa/oe_enclave/enc/private.pem
letmaik: You could generate the key, build the OE app, run `oesign dump` on the enclave image to get MRSIGNER, and then use that value as env var for running the SGX-LKL app. Would that not work?

tests/remote_attestation/maa/oe_enclave/host/host.cpp
```diff
@@ -0,0 +1,60 @@
...
+    printf("Host: starting_server\n");
+    ret = setup_tls_server(enclave, &ecall_ret, SERVER_PORT);
+    if (ecall_ret != 0) {
+        printf("Host: setup_tls_server failed with ecall return val: %d\n", ecall_ret);
+        ret = -1;
+        goto exit;
+    }
+
+    // Sleep 120 seconds to give time to sgxlkl enclave app to start up and connect
+    sleep(120);
```
letmaik: I couldn't see any threading here, so I'm assuming the server actually blocks and waits for the first connection in `setup_tls_server` and then exits. Is that right? In that case, sleeping here doesn't make sense. I may have missed something!

tests/remote_attestation/maa/run_scenario.sh
letmaik: @hukoyu Could you please respond to all comments from @jxyang?

tests/remote_attestation/maa/run_scenario.sh
```diff
@@ -0,0 +1,115 @@
...
+    echo "SGXLKL_RUN_MODE: $SGXLKL_RUN_MODE"
+    exit 1
+fi
+
+# If openenclave is not installed, install it
+if [[ ! -d "/opt/openenclave" ]]; then
+    home=$(pwd)
+    cd ~ || exit 1
+    sudo rm -rf openenclave/
+    git clone --recursive -b feature/sgx-lkl-support https://github.com/openenclave/openenclave.git
```
letmaik: This is not sustainable. We would have to install OE repeatedly if we add more tests that need it. @Pengpeng-Microsoft Can you recommend something?

tests/remote_attestation/maa/run_scenario.sh
```diff
@@ -0,0 +1,115 @@
...
+
+# Build and run OE enclave app
+cd oe_enclave || exit 1
+make
+exit_code=$?
+if [[ $exit_code -ne 0 ]]; then
+    echo "Failed to build OE_ENCLAVE"
+    exit 1
+fi
+make run
```
letmaik: Why don't you first build both components and *then* run?

tests/remote_attestation/maa/sgxlkl_enclave/makefile
```diff
@@ -0,0 +1,29 @@
+.PHONY: all build clean alpine-container alpine-binaries lkl-image lkl-run
+
+SGXLKL_BIN_ROOT="$(SGXLKL_PREFIX)/bin"
+BUILD_DIR="/src/maa/sgxlkl_enclave"
+SGXLKLAPP_COMMON="/src/common"
+
+build-alpine-exe = sudo gcc -g -o $(BUILD_DIR)/alpine_sgxLklApp -fshort-wchar $(BUILD_DIR)/sgxLklApp.c $(BUILD_DIR)/clientAgent_init.c $(BUILD_DIR)/tlscli.c $(SGXLKLAPP_COMMON)/log_utils.c $(SGXLKLAPP_COMMON)/auth.cpp $(SGXLKLAPP_COMMON)/maa.cpp $(SGXLKLAPP_COMMON)/transport_utils.c $(SGXLKLAPP_COMMON)/curl_helper.cpp $(SGXLKLAPP_COMMON)/json_utils.c $(BUILD_DIR)/oe_private/verify_cert.c $(BUILD_DIR)/oe_private/cert.c -lmbedtls -lmbedx509 -lmbedcrypto -ldl -lcurl
```
letmaik: As I said earlier, please break this into multiple lines.

tests/remote_attestation/maa/sgxlkl_enclave/makefile
```diff
@@ -0,0 +1,29 @@
...
+	docker run --rm  -v $(CURDIR)/../..:/src alpine3.10.build bash -c "$(build-alpine-exe)"
+	sudo chown $(USER):$(GROUP) alpine_sgxLklApp
+
+lkl-image: alpine-binaries
+	rm -rf sgxlkl-image.img
+	$(SGXLKL_BIN_ROOT)/sgx-lkl-disk create --size=100M --docker=./Dockerfile sgxlkl-image.img
+
+# This runs sgxlkl_enclave app inside enclave with SGX-LKL
+lkl-run:
+	SGXLKL_VERBOSE=0 SGXLKL_GETTIME_VDSO=0 SGXLKL_TAP=sgxlkl_tap0 $(SGXLKL_BIN_ROOT)/sgx-lkl-run-oe --enclave-config=enclave-config.json --$(SGXLKL_RUN_MODE)-debug sgxlkl-image.img /sgxLklApp 10.0.1.254  >> $(MAA_TEST1_LOG_FILE) 2>&1
```
letmaik: Line breaks.

tests/remote_attestation/maa/sgxlkl_enclave/makefile
```diff
@@ -0,0 +1,29 @@
...
+all: lkl-image
+
+clean:
+	rm -f clientAgent.o alpine_sgxLklApp sgxlkl-image.img*
+
+alpine-container:
+	docker build -t alpine3.10.build ./alpine
+
+alpine-binaries: alpine-container
+	docker run --rm  -v $(CURDIR)/../..:/src alpine3.10.build bash -c "$(build-alpine-exe)"
```
letmaik: Why are you doing this in two separate images? Is there any reason why you're not using multi-stage builds like in the other tests?

tests/remote_attestation/maa/sgxlkl_enclave/oe_private/verify_cert.h
letmaik: Any when the hostverify lib can easily be built on Alpine...
@hukoyu This needs to be documented. It is a huge amount of code that's being duplicated from OE. Please add a note to the README file about this.

tests/remote_attestation/maa/sgxlkl_enclave/tlscli.c
```diff
@@ -0,0 +1,455 @@
+#define _GNU_SOURCE
```
letmaik: I can't comment on binary files, so I'll do it here.
How were the sgxlkl_cert.der and sgxlkl_private_key.pem files generated? If it can't be easily automated, please add a README explaining why and how to create the files manually.

tests/remote_attestation/maa/README.md
```diff
@@ -0,0 +1,26 @@
...
+Here is the test flow:
+1) oe-enclave is the tls server and runs and sleeps 45 seconds and exits
+2) sgxlkl-enclave starts and connects to oe-enclave via TLS trusted connection
+3) sgxlkl-enclave gets report of oe-enclave at TLS connection
+3) sgxlkl-enclave gets authentication token from AAD (Azure Active Directory) in order to access MAA
+4) sgxlkl-enclave calls MAA remote attestation service for JWT token and gets, parses and verifies
+
+This scenario test depends on passed environment variables for credentials and urls accessing MAA service. 
+These are injected by ADO in pipeline:
+- MAA_CLIENT_ID,
```
letmaik: Could you add a short sentence on what each of them is? This has to be reproducible by any external developer.

tests/remote_attestation/maa/run_scenario.sh
hukoyu: That is not true @letmaik 
if you look at containing if block, it will only do that if /opt/openenclave doesn't exits

```
if [[ ! -d "/opt/openenclave" ]]; then
```

tests/remote_attestation/maa/oe_enclave/host/host.cpp
hukoyu: If I don't put sleep here, oe-enclave app exist immediately and no time for sgx-lkl enclave to start and connect.

tests/remote_attestation/maa/run_scenario.sh
letmaik: OK, I see it now. I don't think that's good though as you will never update the OE version then if a CI VM is not torn down. It would be better to install OE in a folder relative to the sgx-lkl repository, like sgx-lkl itself does it. It still doesn't seem right to have all this heavy setup machinery in this particular test.

tests/remote_attestation/maa/oe_enclave/host/host.cpp
letmaik: OK, then I guess it does use threads somehow.

tests/remote_attestation/maa/sgxlkl_enclave/tlscli.c
hukoyu: This is temp solution. We will use sgx-lkl to generate those certificates later. @jxyang  is working on  a solution.

tests/remote_attestation/maa/sgxlkl_enclave/makefile
hukoyu: Done

tests/remote_attestation/maa/oe_enclave/enc/server.cpp
hukoyu: Removed run_server. Not needed.

hukoyu:
> > > To show the structural relationship between this test case and Triston's. Consider placing both tests under `tests/attestation`, and this one named `oe_attests_to_sgxlkl`. The other should be named `sgxlkl_attests_to_oe`.
> > 
> > 
> > @jxyang , @triston996 's test doesn't do attestation. It just verifies TLS trusted connection but doesn't do atetstation agains MAA at all. I moved my comon files to /tests/remote_attestation/common and used from /tests/remote_attestation/maa/
> 
> An attestation could happen without MAA. @triston996 's test is another form of attestation by verifying MRSIGNER etc. in the enclave. Some of my comments suggested `tests/remote_attestation/common` for sharing files between your test and Triston's test. But now I feel `/tests/attestation/common` is a better name. You can work out the name/path between you two. But the bottom line is that there should be no duplicated files between your test and Trinston's test.

@jxyang I applied your feedback:
1) Changed the test folder name to attestation
2) Put common files under attestation/common and test under attestation/maa
3) Moved oe_public and oe_private folders under common/host_verify/public and common/host_verify/private
4)  Changed auth.* to auth_helper.* and maa.* to maa_helper.*
5) Changed log file path to tests/attestation/maa/ instead of $SGXLKL_ROOT/
6) Agreed that need to install OE SDK since oe enclave depends on it
7) We need to keep host_verify code to get the report from certificate until OE supports this
8) We need to keep predefined certificate on sgx-lkl enclave until sgx-lkl supports it for which Xuejun is working on

tests/remote_attestation/maa/run_scenario.sh
hukoyu: @Pengpeng-Microsoft @jxyang @paulcallen any idea what is the best solution here? To me it is acceptable if OE SDK is not updated with every run but it is also good to use the latest OE SDK. How should we proceed about that?

tests/remote_attestation/maa/sgxlkl_enclave/makefile
hukoyu: Will fix that as Maik recommended.

tests/remote_attestation/maa/README.md
hukoyu: Will do that

hukoyu:
@letmaik I addressed all @jxyang 's feedback. I also addressed some of your feedback. Here are the remaining feedback I need to address:
1) Use oesign dump to output MRSIGNER of signed enclave and set to env variable and use from enclaves
2) Explain credential env variables in README.md
3) Explain common/host_verify and why needed in this test code in README.md
4) Use multi-stage build instead of two images for sgx-lkl enclave
5) Clarify best way and frequency to install OE SDK for the test

tests/remote_attestation/maa/run_scenario.sh
wintersteiger: @letmaik: the heavy machinery is fine here, but we should be careful to reuse all this in future tests. Applications will have to bring their own libraries, in this case it just happens to be a (user-mode) OE. 

Definitely don't install anything system-wide, always locally. The `git clone` here will always get the latest version, i.e. this can change at random points in time, so don't keep any state, otherwise it will be hell to figure out why the CI machine behaves differently to our dev machines.

hukoyu:
> @letmaik I addressed all @jxyang 's feedback. I also addressed some of your feedback. Here are the remaining feedback I need to address:
> 
> 1. Use oesign dump to output MRSIGNER of signed enclave and set to env variable and use from enclaves
> 2. Explain credential env variables in README.md
> 3. Explain common/host_verify and why needed in this test code in README.md
> 4. Use multi-stage build instead of two images for sgx-lkl enclave
> 5. Clarify best way and frequency to install OE SDK for the test

@letmaik  @jxyang  @paulcallen 1, 2, and 3 are done and commit submitted. 
Working on 4, and there i sno action item for 5 for now. After I submit 4 , this PR should be good for merging.

hukoyu:
Tried to use multi-stage build instead of two images for sgx-lkl enclave, but couldn't get it working. 
It would be nice to have this done but it cause issues and fails. May be we can spend more time on that later but this is not a prerequisite for this test for now. The test is working fine as is.

@jxyang @letmaik  @paulcallen  I addressed all other feedback and I strongly believe this PR is in good shape to be merged to have test coverage for this remote attestation scenario. Could you please approve so that we can merge and enable this test in nightly builds.

tests/attestation/maa/oe_enclave/enc/makefile
```diff
@@ -0,0 +1,45 @@
...
+	$(CXX) -c $(CXXFLAGS) $(INCLUDES) -std=c++11 server.cpp
+	$(CXX) -c $(CFLAGS) $(CINCLUDES) tlssrv.cpp
+	$(CXX) -c $(CFLAGS) $(CINCLUDES) gencreds.cpp
+	$(CC) -c $(CFLAGS) $(CINCLUDES) -I$(COMMONS) $(COMMONS)/transport_utils.c
+	$(CC) -c $(CFLAGS) $(CINCLUDES) -I$(COMMONS) $(COMMONS)/log_utils.c
+	$(CC) -c $(CFLAGS) $(CINCLUDES) oeApp_t.c
+	$(CXX) -o oeApp_enc server.o tlssrv.o gencreds.o oeApp_t.o log_utils.o transport_utils.o $(LDFLAGS) -lmbedtls -lmbedcrypto -loehostsock -loehostresolver -loecore -loelibc -loesyscall
+
+sign:
+	oesign sign -e oeApp_enc -c enc.conf -k private.pem
```
jxyang, line 43: I had a similar comments with Triston's PR: since you are exporting the MRSIGNER of oe enclave as an env variable, we could use openssl to generate a random private key and get rid of the below private key file.

hukoyu: Removed the hardcoded private.pem and generating with openssl at runtime as @jxyang  recommended.

hukoyu:
Used the gencreds under common/ an dremoved the duplicates. Also, used openssl to generate a random private key and get rid of the checked in private key file. @jxyang this PR ready to be approved :)

jxyang (approved):

hukoyu:
@davidchisnall  @prp  @wintersteiger  can I get an approval from one of you? I addressed all feedback and this test is ready to be merged and enabled to increase test coverage for remote attestation.

hukoyu:
Verified test passing in nightly build for hw mode and skipped for sw mode.

PR #653, opened 2020-07-22
Remove unneeded 'lkl_run_in_kernel_stack'

SeanTAllen:
Encrypted disk setup works fine without it. Why this was previously needed
isn't noted. However, all tests pass so it can only be assumed that whatever
was the cause of the additional complexity has been fixed/addressed.


SeanTAllen:
@davidchisnall looks like it might indeed be needed as this looks like the error from #604.

PR #655, opened 2020-07-22
Don't fallback to lkl mmap

SeanTAllen:
the fallback from enclave mmap to lkl mmap is a source of possible subtle bugs and is being removed. this is part 1 of a few incremental changes to mmap usage in order to address layering issues and bugs.


SeanTAllen:
@prp @davidchisnall CI is running without the fallback to lkl mmap.

src/lkl/posix-host.c
```diff
@@ -681,6 +681,13 @@ static void *host_malloc(size_t size)
 	{
 		SGXLKL_ASSERT(kernel_mem == NULL);
 		kernel_mem = enclave_mmap(0, size, 0, PROT_READ | PROT_WRITE, 0);
+        if (((intptr_t)kernel_mem) < 0)
```
SeanTAllen: the formatting is off here due to #665. I am planning on addressing after this PR. I didn't want to muddy up the diff with a ton of formatting.

src/lkl/syscall-overrides-mem.c
```diff
@@ -0,0 +1,96 @@
...
+    {
+        void* mem =
+            enclave_mmap(addr, length, flags & MAP_FIXED, prot | PROT_WRITE, 0);
+
+        if (mem > 0)
+        {
+            // Read file into memory
+            size_t readb = 0;
+            ssize_t ret = 0;
+            while ((ret = pread(
```
SeanTAllen: if I use pread_fn I get an unbalanced lkl_cpu_put error from the java hello world error.

I wasn't expecting that @davidchisnall and would like to know what faulty assumption I made about this half-way point to the changes in #604. 

I can move things back out of the syscall-overrides-mem etc as without using pread_fn, all this can go back into enclave_mem.c

davidchisnall: That's odd.  This should be `lkl_sys_pread`, not the libc wrapper, but I don't understand why calling a syscall implementation function would cause problems but making `lkl_syscall` reentrant (which is *really* wasn't designed to be) would work.

davidchisnall (approved):

src/lkl/syscall-overrides-mem.c
SeanTAllen: updated.

SeanTAllen:
Because CI round-trip integration is broken with GH at the moment, passing CI is here: https://dev.azure.com/sgx-lkl/sgx-lkl/_build/results?buildId=1547&view=results

SeanTAllen:
ill need to rebase sgx-lkl-musl before this is merged. that can be part of the bumping.

all the same code is on both, but there was an issue with it not pointing to a squashed very of some commits when I branched.

src/enclave/enclave_mem.c
```diff
@@ -97,59 +98,59 @@ static size_t addr_to_index(void* addr)
...
+        void* mem =
             enclave_mmap(addr, length, flags & MAP_FIXED, prot | PROT_WRITE, 0);
+
         if (mem > 0)
         {
             // Read file into memory
-            lseek(fd, offset, SEEK_SET);
-            size_t readb = 0, ret;
-            while ((ret = read(fd, ((char*)mem) + readb, length - readb)) > 0)
+            size_t readb = 0;
```
mikbras, line 128: Just wondering whether readb was supposed contain something non-zero. Otherwise, I'm not sure what its purposes is.

mikbras (approved):
Sean, looks great. I couldn't find any problems. Thank you for doing this work.

SeanTAllen:
Ok because of reverts to sgx-lkl-musl. This PR looks to be messed up

src/enclave/enclave_mem.c
SeanTAllen: It's how many bytes your have gotten back from pread.

in case pread returns less than legnth and you need to read again.

PR #656, opened 2020-07-22
LTP: fixed lgetxattr02 and preadv203 tests

vamsikrishna935:
Tests were failing because of kernel panic in loop device.
So Modified the tests to use root file system.


SeanTAllen:
Is there an issue open for "Tests were failing because of kernel panic in loop device."? 

Is that the referenced #297? If it is, that isn't clear.

hukoyu (changes requested):
@vamsikrishna935  preadv203 is failing. It doesn't look fixed.

vamsikrishna935:
> 
> 
> Is there an issue open for "Tests were failing because of kernel panic in loop device."?
> 
> Is that the referenced #297? If it is, that isn't clear.
@SeanTAllen 
We have memory limit of 32MB in loop device, if the tests use more than 32MB tests will fail with out of memory (kernel panic). So in that case we are using root filesystem for running the tests as suggested by peter.

Isuue 297 is different, whenever a test tests the invalid address, it was supposed to return EFAULT but here it is causing enclave abort.

vamsikrishna935:
> 
> 
> @vamsikrishna935 preadv203 is failing. It doesn't look fixed.

@hukoyu Could you please rerun the checks once, it is passing in local with status code 0

hukoyu:
> > @vamsikrishna935 preadv203 is failing. It doesn't look fixed.
> 
> @hukoyu Could you please rerun the checks once, it is passing in local with status code 0

@vamsikrishna935 I cannot do that. If it failed twice in the pipeline run that it means it is not fixed properly. Can you look at the failure and understand why it is failed?

vamsikrishna935:
> 
> 
> > > @vamsikrishna935 preadv203 is failing. It doesn't look fixed.
> > 
> > 
> > @hukoyu Could you please rerun the checks once, it is passing in local with status code 0
> 
> @vamsikrishna935 I cannot do that. If it failed twice in the pipeline run that it means it is not fixed properly. Can you look at the failure and understand why it is failed?

Sure Huseyin

SeanTAllen:
@vamsikrishna935 someone else is going to come along and not know about the combination of 32 megs and kernel panic etc. Can you add that little bit of extra detail to the patch? It will make someone's life in the future must better and make us as a collective team less dependent on oral history to understand the patch (and other that might have the same issue).

Thanks.

vamsikrishna935:
> 
> 
> @vamsikrishna935 someone else is going to come along and not know about the combination of 32 megs and kernel panic etc. Can you add that little bit of extra detail to the patch? It will make someone's life in the future must better and make us as a collective team less dependent on oral history to understand the patch (and other that might have the same issue).
> 
> Thanks.

@SeanTAllen Sure i will update patch description

SeanTAllen:
@vamsikrishna935 Awesome. Thank you. If you can keep context like that in mind when creating other patches and flesh the descriptions out a bit more like you are going to with this one, that would be great. Thank you.

hukoyu:
This will be resubmitted to lsds/ltp repo.

PR #658, opened 2020-07-22
Set memory as PROT_NONE when unmapping memory

SeanTAllen:
To see what goes boom.


SeanTAllen:
@davidchisnall @prp here's the PROT_NONE on munmap results. 

For the "core tests" that have failed,  a quick glance suggests at the moment, it's mostly the same problem. I didn't see any that weren't crashing at the same point in output.

LTP tests are still running as I write this.

PR #662, opened 2020-07-23
Fix descriptions and initializers of enums in config

wintersteiger:
Descriptions of enum-valued config settings were not printed with `--help-enclave-config` and initializers were missing.


SeanTAllen (approved):
The python code itself looks reasonable as python, but I don't really understand the context it is used in.

I'm approving with those caveats.

wintersteiger:
Yeah, this is a very specific json schema to C source translation, the semantics really depend on the contents and structure of the schema files.

PR #664, opened 2020-07-23
[Test] New Test Case Database-ODBC-attestation

praenubilus:
This is the test case for establishing database connection and sending query from an sgxlkl application to a remote SQL Server database with always encrypted feature. After connection between the app and the database is established with attestation, the test will do the following queries and operations:

1. Create a CMK (Column Master Key) with random key name suffix. Column Master Keys are keys used to encrypt column encryption keys. [More details can be found here](https://docs.microsoft.com/en-us/sql/relational-databases/security/encryption/overview-of-key-management-for-always-encrypted?view=sql-server-ver15).
2. Create CEK (Column Encryption Key) with random key name suffix. Note: Column encryption keys are content-encryption keys used to encrypt data in database columns. [More details can be found here](https://docs.microsoft.com/en-us/sql/relational-databases/security/encryption/overview-of-key-management-for-always-encrypted?view=sql-server-ver15).
3. Create data table with random table name suffix.
4. Insert data into the table created.
5. Query data from the data table, and compare the decrypted query result, compare it with the expected result.
6. Drop all generated data table, CEK(Column Encrypted Key), CMK(Column Master Key).

The database have a registry table, for each successful create/delete database object(Table, CMK, CEK), a record will be insert/delete from the registry table. The initialization of the registry table can be configured in `enclave-config.json` file.


tests/database/odbc-w-enclave/odbc_proxy.c
```diff
@@ -0,0 +1,388 @@
...
+    OK_CHECK(clean_up(stmt));
+
+    // 2. Create CMK (Column Master Key)
+    OK_CHECK(create_CMK(stmt));
+
+    // 3. create CEK (Column Encryption Key)
+    OK_CHECK(create_ecek(stmt));
+
+    // 4. Create Data table
+    OK_CHECK(create_table(stmt));
```
Pengpeng-Microsoft: Would this fail if table already exits?

tests/database/odbc-w-enclave/cksp.c
```diff
@@ -0,0 +1,155 @@
+/***************************************************************************
+  Custome Key Store Provider
+
+  This module maingly interacts with ODBC driver and VCR Agent. It is
```
Pengpeng-Microsoft: The following comments are wrong. Can you review and replace?

Pengpeng-Microsoft (changes requested):

tests/database/odbc-w-enclave/README.md
```diff
@@ -0,0 +1,39 @@
+# Database Test Case (ODBC Attestation)
+
+## Description
+
+This is the test case for database connection and query from sgxlkl application to a remote SQLServer database with always encrypted V2. After connection with attestation established, the test will do following queries and operations:
+
+1. before start, drop the test data table, CEK(Column Encrypted Key), CMK(Column Master Key)
```
letmaik: Since this test may run concurrently in different configurations in CI (debug, non-release, etc), does this actually work? Should the test table name be generated randomly?

tests/database/odbc-w-enclave/README.md
```diff
@@ -0,0 +1,39 @@
...
+- Application programming language: **_C_**. 
+- Database Server: **_Azure SQL Database_**.
+- Database driver: **_ODBC_**.
+- Docker builder: **_alpine:3.10_**
+
+## To Run the Test
+
+### Prerequists
+
+- You should have a SQL Server database created and having **read permission**
```
letmaik: How can you write to the db with read permissions?

tests/database/odbc-w-enclave/README.md
```diff
@@ -0,0 +1,39 @@
...
+- Clean up existing binaries or image files
+    ```bash
+        make clean
+    ```
+- set the following environment variables before run the test
+    - **MAA_SERVICE_ADD**: the http address of the attestation provider
```
letmaik: Why is this env var name so weird? Why not just `MAA_URL`?

tests/database/odbc-w-enclave/msodbcsql.h
```diff
@@ -0,0 +1,730 @@
+#ifndef __msodbcsql_h__
+#define __msodbcsql_h__
+
+/*
+ *-----------------------------------------------------------------------------
+ * File:        msodbcsql.h
+ *
+ * Copyright:   Copyright (c) Microsoft Corporation
```
letmaik: What's the license of this file? Why does it have to be copied to the repo?

tests/database/odbc-w-enclave/odbc_app.c
```diff
@@ -0,0 +1,74 @@
...
+    char* db_uid = getenv("DB_USERID");
+    char* db_name = getenv("DB_NAME");
+    char* maa_add = getenv("MAA_SERVICE_ADD");
+
+    if (db_password == NULL || db_server == NULL || db_uid == NULL ||
+        db_name == NULL || maa_add == NULL)
+    {
+        fprintf(
+            stderr,
+            "\n-------------------- Failure --------------------"
```
letmaik: Can we make this shorter and just prefix lines with "Error: "?

tests/database/odbc-w-enclave/cksp.c
```diff
@@ -0,0 +1,155 @@
...
+{
+    return 1;
+}
+
+void KeyStoreFree()
+{
+}
+
+CEKEYSTOREPROVIDER2 MyCustomKSPName_desc = {
+    // pretend to be AKV, so SQL server will forward the ECEK decryption
```
letmaik: Is this a hack? Will this break in the future?

tests/database/odbc-w-enclave/odbc_proxy.c
```diff
@@ -0,0 +1,388 @@
+#include <stdlib.h>
```
letmaik: Why is this file called proxy?

tests/database/odbc-w-enclave/odbc_proxy.c
```diff
@@ -0,0 +1,388 @@
...
+static int create_CMK(SQLHSTMT stmt)
+{
+    int result = FAILURE;
+    SQLRETURN rc;
+    char* stmt_str =
+        "CREATE COLUMN MASTER KEY " CMK_NAME " WITH"
+        "(KEY_STORE_PROVIDER_NAME = N'AZURE_KEY_VAULT',"
+        "KEY_PATH = "
+        "N'https://acckeyvault.vault.azure.net/keys/TestCMK1/"
+        "2a89d8e0cb834009827d588a32709e39',"
```
letmaik: Please document where all those magic strings come from. This has to be reproducible with a custom SQL server / AKV instance.

tests/database/odbc-w-enclave/README.md
praenubilus: That is a good point I have not considered when writing the test case. That also raise other concerns, this test case will also include CEK(Column Encrypted Key), CMK(Column Master Key) creation and drop. What is your opinion @jxyang @Pengpeng-Microsoft

tests/database/odbc-w-enclave/msodbcsql.h
praenubilus: SQL AE feature is still in public preview. The release version of the latest ODBC header will not support all the features, so we have to modify the release version of ODBC header and copy it to app code

tests/database/odbc-w-enclave/cksp.c
praenubilus: @jxyang can you comment on this?

tests/database/odbc-w-enclave/odbc_proxy.c
praenubilus: Since we are using custom key store provider, this AKV url is just a place holder. But it is required by SQL Server at least for now. The key value will not be actually accessed by the Database Server. @jxyang any other comments?

letmaik: It looks very real to me. Can you not replace that with a dummy url?

tests/database/odbc-w-enclave/msodbcsql.h
letmaik: Then there *must* be a big fat TODO somewhere that says exactly that and it should also say that this custom patched header can be removed once it's in GA.

tests/database/odbc-w-enclave/odbc_proxy.c
praenubilus: Yes. But in step 1, we drop data table first, if it exists

tests/database/odbc-w-enclave/odbc_proxy.c
praenubilus: any suggestion for naming?

tests/database/odbc-w-enclave/odbc_proxy.c
praenubilus: Theoretically yes, but that requires us to reverse engineer the enclave signature hashing algorithm(based on the given dummy url).

tests/database/odbc-w-enclave/msodbcsql.h
praenubilus: Will it be better to put in this header file or in the README file?

letmaik: Probably the header file.

tests/database/odbc-w-enclave/odbc_proxy.c
letmaik: What does it do?

tests/database/odbc-w-enclave/odbc_proxy.c
letmaik: I don't understand, how is hashing related to the URL? What happens if you replace the URL?

tests/database/odbc-w-enclave/README.md
Pengpeng-Microsoft: I would recommend a postfix to CEK/CMK/table by combination of configuration and random number. Beyond that, you can also do a lock record based checking mechanism to be more cautious.

tests/database/odbc-w-enclave/odbc_proxy.c
praenubilus: Here is the [CREATE COLUMN MASTER KEY (Transact-SQL)](https://docs.microsoft.com/en-us/sql/t-sql/statements/create-column-master-key-transact-sql?view=sql-server-ver15), to generate a column master key, a `[,ENCLAVE_COMPUTATIONS (SIGNATURE = signature)]` is also required. Usually that signature hash is generated by SQL Server client like(SSMS, Data Studio). The signature is used to verify the authentic of the Key Path URL(dummy URL in our case). You have to manually compute the signature hash in your application if you want to generate pragmatically.

tests/database/odbc-w-enclave/odbc_proxy.c
praenubilus: It includes all SQL operations(create/drop table/key, query(insert, query) data,  connect to database) through ODBC

tests/database/odbc-w-enclave/odbc_proxy.c
letmaik: OK, could you add some comment to the code that explains this?

tests/database/odbc-w-enclave/odbc_proxy.c
letmaik: Not sure, my only bad suggestion would be odbc_helper. I'll let others make the call.

tests/database/odbc-w-enclave/cksp.c
jxyang: This is a hack suggested by SQL team because AEv2 only recognizes built-in key vaults, not the custom ones. In fact, if we register with AEv2 a custom key vault, it doesn't even forward the key decryption request to the driver. Yes this could break in the future, and we should document  in more details.

tests/database/odbc-w-enclave/odbc_proxy.c
jxyang: I think  odbc_helper is a proper name.

tests/database/odbc-w-enclave/Dockerfile
```diff
@@ -0,0 +1,38 @@
+# builder for compile
+FROM alpine:3.10 AS builder
```
jxyang: Suggest renaming the test case to `odbc_with_cksp`

tests/database/odbc-w-enclave/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+
+USER root
+
+RUN apk update
+RUN apk add --no-cache bash shadow unixodbc-dev build-base
+
+ADD *.c /app/
+ADD *.h /app/
+RUN gcc -g -fshort-wchar -fPIC -o /app/cksp.so -shared /app/cksp.c
+RUN gcc -g -o /app/odbc_app -fshort-wchar /app/odbc_app.c /app/odbc_proxy.c -lodbc -ldl
```
jxyang: This is what I suggested to Huseyin: using one docker file for both building the app and creating the sgx-lkl disk image. Nice!

tests/database/odbc-w-enclave/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+COPY --from=builder /app/odbc_app .
+
+RUN mkdir -p /tmp/msodbcinstall && cd /tmp/msodbcinstall && \
+curl -O https://download.microsoft.com/download/e/4/e/e4e67866-dffd-428c-aac7-8d28ddafb39b/msodbcsql17_17.5.2.2-1_amd64.apk 
+RUN cd /tmp/msodbcinstall && sudo apk add --allow-untrusted $(ls)
+RUN rm -rf /tmp/msodbcinstall
+
+WORKDIR /
+# Start from a Bash prompt
+CMD ["/bin/bash"]
```
jxyang: Probably don't need the cmd. sgxlkl_run_oe takes a command line option for this.

tests/database/odbc-w-enclave/Makefile
```diff
@@ -0,0 +1,40 @@
+include ../../common.mk
+
+PROG=odbc_app
+IMAGE_SIZE=50M
+
+EXECUTION_TIMEOUT=120
+SGXLKL_VERBOSE=0 
+SGXLKL_ENV=SGXLKL_VERBOSE=1 SGXLKL_KERNEL_VERBOSE=1 SGXLKL_GETTIME_VDSO=0 SGXLKL_TAP=sgxlkl_tap0 SGXLKL_HOST_IMPORT_ENV="DB_SERVER_NAME,DB_NAME,DB_PASSWORD,DB_USERID,MAA_SERVICE_ADD"
```
jxyang: Break up the line. There is a redundant SGXLKL_VERBOSE. We don't need SGXLKL_GETTIME_VDSO=0  because the clock implementation is not depending on vdso anymore.

tests/database/odbc-w-enclave/README.md
```diff
@@ -0,0 +1,39 @@
+# Database Test Case (ODBC Attestation)
```
jxyang: The name is too generic. Also ODBC Attestation is ambiguous, e.g., who attests who?

tests/database/odbc-w-enclave/README.md
jxyang: When the test is running, it doesn't know which CI config it's running in. Probably appending a random number to the table name is good enough.

jxyang: The first step is connecting to the DB.

tests/database/odbc-w-enclave/README.md
```diff
@@ -0,0 +1,39 @@
...
+## Description
+
+This is the test case for database connection and query from sgxlkl application to a remote SQLServer database with always encrypted V2. After connection with attestation established, the test will do following queries and operations:
+
+1. before start, drop the test data table, CEK(Column Encrypted Key), CMK(Column Master Key)
+2. Create CMK (Column Master Key)
+3. create CEK (Column Encryption Key)
+4. Create Data table
+5. Insert Data into the table
+6. Query data
```
jxyang: We should drop the table as the final step, as other pipeline jobs don't know the exact table name anymore with the random suffix.

tests/database/odbc-w-enclave/README.md
```diff
@@ -0,0 +1,39 @@
...
+ 
+## Components and Tools Used
+
+- Application programming language: **_C_**. 
+- Database Server: **_Azure SQL Database_**.
```
jxyang: We should emphasize this is SQL AE

tests/database/odbc-w-enclave/README.md
```diff
@@ -0,0 +1,39 @@
...
+- Database driver: **_ODBC_**.
+- Docker builder: **_alpine:3.10_**
```
jxyang: Alpine version is not important and most likely changing soon.

tests/database/odbc-w-enclave/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Description
+
+This is the test case for database connection and query from sgxlkl application to a remote SQLServer database with always encrypted V2. After connection with attestation established, the test will do following queries and operations:
+
+1. before start, drop the test data table, CEK(Column Encrypted Key), CMK(Column Master Key)
+2. Create CMK (Column Master Key)
+3. create CEK (Column Encryption Key)
+4. Create Data table
+5. Insert Data into the table
```
jxyang: There should be detailed description for this step on how SQL enclave attests to MAA, and after verification, how is key decryption requests flow from SQL server to ODBC, and then get handled by CKSP.

tests/database/odbc-w-enclave/README.md
```diff
@@ -0,0 +1,39 @@
+# Database Test Case (ODBC Attestation)
+
+## Description
+
+This is the test case for database connection and query from sgxlkl application to a remote SQLServer database with always encrypted V2. After connection with attestation established, the test will do following queries and operations:
+
+1. before start, drop the test data table, CEK(Column Encrypted Key), CMK(Column Master Key)
+2. Create CMK (Column Master Key)
```
jxyang: Pls. describe CMK and CEK.

tests/database/odbc-w-enclave/cksp.c
jxyang: Please remove confidential data from the file.

tests/database/odbc-w-enclave/odbc_proxy.c
jxyang: Again this is a hack to work around the SQL AE's failure to recognize custom key vaults. The key path has to be a real akv path (even though it's a placeholder only and never been accessed for key operations). The signature has to be a hash value computed from the key path (and other settings). Please explain this in the comments. @praenubilus

tests/database/odbc-w-enclave/odbc_proxy.c
```diff
@@ -0,0 +1,388 @@
...
+    unsigned short ECEKlen;
+    char ecekStr[2 * ECEKlen + 1];
+    CEKEYSTORECONTEXT* ctx = {0};
+    CEKEYSTOREPROVIDER2* pKsp = NULL;
+    SQLRETURN rc;
+
+    // Load n init custom key store provider dynamic link library
+    if (!(pKsp = loadKspByName(DLLPATH, KSPNAME)))
+    {
+        fprintf(stderr, "Failed to load CKSP\n");
```
jxyang: Failure cases should take care of closing `hProvLib`.

tests/database/odbc-w-enclave/odbc_proxy.c
```diff
@@ -0,0 +1,388 @@
...
+    SQLRETURN rc;
+    char data[DEFAULT_VCHAR_LEN];
+
+    rc = SQLExecDirect(stmt, "SELECT * FROM " TABLE_NAME, SQL_NTS);
+    OK_CHECK(checkRC(rc, "Query table", stmt, SQL_HANDLE_STMT));
+    rc = SQLBindCol(stmt, 1, SQL_C_CHAR, data, DEFAULT_VCHAR_LEN, 0);
+
+    if (SQL_SUCCESS == (rc = SQLFetch(stmt)))
+    {
+        printf("\nQuery from encrypted column successfully. Decrepted data: %s\n", data);
```
jxyang: We should check the decrypted string is `encrypted-data`

jxyang (changes requested):

jxyang:
Overall I expect more comments in the code.

tests/database/odbc-w-enclave/README.md
praenubilus: Using a suffix(timestamp+random number) for all database objects. A registry table has also been created in the database, for every table/cmk/cek creation, a record will be registered in registry. At the end of the test, table, cmk and cek will be dropped. For each successful drop operation, related registry record will be deleted.

praenubilus:
@jxyang @Pengpeng-Microsoft @letmaik all comments and change requests have been addressed.

tests/database/odbc-with-cksp/Makefile
```diff
@@ -0,0 +1,53 @@
...
+ENCLAVE_CONFIG=--enclave-config=$(ENCLAVE_CONFIG_FILE)
+SGXLKL_HW_PARAMS=$(ENCLAVE_CONFIG) --hw-debug
+SGXLKL_SW_PARAMS=$(ENCLAVE_CONFIG) --sw-debug
+
+SGXLKL_ROOTFS=sgx-lkl-rootfs.img
+
+.DELETE_ON_ERROR:
+.PHONY: all clean
+
+$(SGXLKL_ROOTFS): clean $(PROG_SRC)
```
letmaik: clean is a phony target that will always be run here. If you need to clean up, then do this inside the build rule.

tests/database/odbc-with-cksp/Makefile
```diff
@@ -0,0 +1,53 @@
...
+run-hw-gdb: ${SGXLKL_ROOTFS}
+	  $(SGXLKL_ENV) $(SGXLKL_GDB) --args $(SGXLKL_STARTER) $(SGXLKL_HW_PARAMS) $(SGXLKL_ROOTFS)
+
+run-sw-gdb: ${SGXLKL_ROOTFS}
+	  $(SGXLKL_ENV) $(SGXLKL_GDB) --args $(SGXLKL_STARTER) $(SGXLKL_SW_PARAMS) $(SGXLKL_ROOTFS)
+
+clean:
+	rm -f $(PROG) *.so *.o *.img*
+
+tt:
```
letmaik: ?

tests/database/odbc-with-cksp/README.md
```diff
@@ -0,0 +1,51 @@
+# Azure SQL ODBC w/ Database Attestation and Custom Key Store Provider Test Case
```
letmaik: Please don't use "w/" in the README.

letmaik: Can a native English speaker go over this and suggest fixes re grammar/typos?

tests/database/odbc-with-cksp/enclave-config.json
```diff
@@ -0,0 +1,14 @@
+{
+    "args": [
+        "odbc_app",
+        "classic",
+        ""
```
letmaik: ?

tests/database/odbc-with-cksp/odbc_helper.c
```diff
@@ -0,0 +1,550 @@
...
+static void postKspError(CEKEYSTORECONTEXT* ctx, const wchar_t* msg, ...)
+{
+    if (msg > (wchar_t*)65535)
+        wprintf(L"Provider emitted message: %s\n", msg);
+    else
+        wprintf(L"Provider emitted message ID %d\n", msg);
+}
+
+// Find the custom Key Store Provider by name
+static CEKEYSTOREPROVIDER2* loadKspByName(char* libPath, wchar_t* kspName)
```
letmaik: Use a consistent function naming convention.

tests/database/odbc-with-cksp/odbc_helper.h
```diff
@@ -0,0 +1,11 @@
+#ifndef ODBC_PROXY_H
```
letmaik: Outdated

tests/database/odbc-with-cksp/Makefile
praenubilus: I forget to delete the recipe for local host debugging

tests/database/odbc-with-cksp/enclave-config.json
praenubilus: there are two args for running the application, the 2nd one by default is set to empty value

tests/database/odbc-with-cksp/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Description
+
+This is the test case for database connection and query from sgxlkl application to a remote SQLServer database with always encrypted. After connection between app and database with attestation established, the test will do following queries and operations:
```
jxyang:
```suggestion
This is the test case for establishing database connection and sending queries from a sgxlkl application to a remote SQL Server database with always encrypted feature. After the connection between the app and the database is established with attestation, the test will do the following queries and operations:
```

jxyang: @praenubilus Please fix the rest of this document.

praenubilus:
all comments addressed.

praenubilus:
@jxyang @Pengpeng-Microsoft @letmaik All comments and required changes have been addressed. Tested and passed in nightly pipeline.

Pengpeng-Microsoft (approved):

tests/database/odbc-with-cksp/enclave-config.json
```diff
@@ -0,0 +1,14 @@
+{
+    "args": [
+        "odbc_app",
```
letmaik, line 3: @prp Any reason why this works at all? Shouldn't only `./odbc_app` and `/odbc_app` work here?

tests/database/odbc-with-cksp/enclave-config.json
```diff
@@ -0,0 +1,14 @@
...
+        "classic",
+        "default"
+    ],
+    "host_import_env": [
+        "MAA_URL",
```
letmaik, line 8: Where is MAA_URL used?

tests/database/odbc-with-cksp/enclave-config.json
praenubilus: because I have set `WORKDIR` in Dockerfile?

tests/database/odbc-with-cksp/enclave-config.json
praenubilus: In `odbc_app.c` line:47. It is used as part of the SQL Server Always Encrypted connection string

tests/database/odbc-with-cksp/enclave-config.json
praenubilus: I followed this sample([https://github.com/lsds/sgx-lkl/blob/oe_port/samples/basic/helloworld/enclave_config.json](https://github.com/lsds/sgx-lkl/blob/oe_port/samples/basic/helloworld/enclave_config.json)) to set the config file. As you can see, it is not necessary to put `./` or `absolute path` before the binary

letmaik (approved):
Thanks for all the cleanup and work on this PR, I think it's in a good state now.

tests/database/odbc-with-cksp/enclave-config.json
prp: The binary is accessed with a regular `open` call, so the path will be interpreted relative to the current working directory.

jxyang (approved):

Issue #597, opened 2020-07-06
Replace enclave mmap with LKL mmap implementation


prp:
Currently SGX-LKL has a separate mmap implementation, and does not use LKL's `mmap()` for heap allocations. A drawback of this design is that LKL is unaware of userspace allocations and maintains its own fragmented memory space.

Instead, we should remove the enclave `mmap()` implementation and have all allocations be handled by LKL.

Currently, replacing the enclave mmap allocator with the LKL one results in a segfault in the musl libc dynamic loader.

PR #675, opened 2020-07-24
LTP: Fixed fallocate04 and fallocate05 testcases

vamsikrishna935:
Tests were failing with Kernel panic (out of memory) in loop device.
loop device filesystem is having a limit of 32MB
So modified the tests using root filesystem.


davidchisnall:
I think this patch is fine, but please can we track the tests that are failing because we don't have enough kernel memory centrally?  These should all be revisited after #597 is fixed.

davidchisnall (approved):

tests/ltp/patches/ltp_fallocate_fallocate04_fix.patch
```diff
@@ -0,0 +1,44 @@
++ Tesst Failure reason:
```
hukoyu: Can you change the patch file name to fallocate04.patch
You can run 
git mv ltp_fallocate_fallocate04_fix.patch  fallocate04.patch

tests/ltp/patches/ltp_fallocate_fallocate05_fix.patch
```diff
@@ -0,0 +1,66 @@
++ Test failure reason:
```
hukoyu: Can you change the patch file name to fallocate05.patch
You can run 
git mv ltp_fallocate_fallocate05_fix.patch  fallocate05.patch

hukoyu (changes requested):
@vamsikrishna935 please change patch file names as suggested. Afte rthat we can merge this PR.

hukoyu (approved):
LGTM

Issue #598, opened 2020-07-06
Add ELF loading support to LKL


prp:
We want LKL to be able to load ELF executables and libraries using the `binfmt_elf_fpdic` format. This format supports the loading of regular ELF binaries on a non-MMU architecture. 

This means that LKL must be compiled with `CONFIG_ELFCORE` and `CONFIG_ BINFMT_ELF_FDPIC` support. For this, we must include missing architecture-specific header files for ELF support under `arch/lkl`.

prp:
I presume that this was closed by accident?

PR #676, opened 2020-07-24
LTP: fix test case sigrelse01 issue

shaikshavali1:
This test case creates a child process using "FORK_OR_VFORK".
Fork behavior is not supported in sgx-lkl.
Hence, the below modifications are performed.
	1. Removed the code related to child process creation.
	2. Moved child and parent-related code/functionality to a single function "verify_sigrelse"
	3. Signal with ID 34 is skipped from testing. This signal ID is used for internal signals in the thread library of sgx-lkl-musl.
	4. Removed pipe handling code which is used for interprocess communication between parent and child.
	5. Removed the code related to the timer, we already have 60sec top-level timer for the test.


tests/ltp/patches/fix_sigrelse_sigrelse01.patch
```diff
@@ -0,0 +1,673 @@
...
+-static void getout(void);
+ int choose_sig(int sig);
+ 
+ #define TRUE  1
+@@ -141,29 +135,12 @@ int choose_sig(int sig);
+ #ifndef DEBUG
+ #define DEBUG 0
+ #endif
+-
+-#define CHILD_EXIT(VAL) ((VAL >> 8) & 0377)	/* exit value of child process */
```
davidchisnall: I think this is now dead code, please leave it in: eventually we will want to revert these and it makes it much easier to understand the diffs if they only include functional changes.

tests/ltp/patches/fix_sigrelse_sigrelse01.patch
```diff
@@ -0,0 +1,673 @@
...
+-#define SIG_CAUGHT 8
+-#define WRITE_BROK 16
+-#define HANDLE_ERR 32
+-
+ int TST_TOTAL = 1;		/* number of test items */
+ 
+ char *TCID = "sigrelse01";	/* test case identifier */
+-static char mesg[MAXMESG];	/* message buffer for tst_res */
+ static int pid;			/* process id of child */
+-static int pipe_fd[2];		/* file descriptors for pipe parent read */
```
davidchisnall: We ought to be able to use pipes between threads, why was this change made?

tests/ltp/patches/fix_sigrelse_sigrelse01.patch
```diff
@@ -0,0 +1,673 @@
...
++}
++
+ /***********************************************************************
+  *   M A I N
+  ***********************************************************************/
+@@ -192,9 +216,6 @@ int main(int argc, char **argv)
+ 	 * parse standard options
+ 	 */
+ 	tst_parse_opts(argc, argv, NULL, NULL);
+-#ifdef UCLINUX
```
davidchisnall: Why is this removed?  We should be using the UCLINUX code paths because we are a uCLinux implementation.

davidchisnall (changes requested):
This is difficult to review with extraneous changes in the diff, please remove them.

If we are not using the uCLinux changes, please document why and track it (is it because exec is the problem)?

tests/ltp/patches/fix_sigrelse_sigrelse01.patch
```diff
@@ -0,0 +1,673 @@
+This test case creates a child process using "FORK_OR_VFORK".
```
hukoyu: Please change patch file name to sigrelse01.patch 
You can run command
git mv fix_sigrelse_sigrelse01.patch sigrelse01.patch

hukoyu (changes requested):
@shaikshavali1  please change the patch file name as requested. Please also address David's feedback.

tests/ltp/patches/fix_sigrelse_sigrelse01.patch
shaikshavali1: @hukoyu, I renamed the file to sigrelse01.patch.

shaikshavali1:
@hukoyu, I addressed your review comment.
@davidchisnall, This patch is not using threads. Hence, the pipe related code is removed. uClinux code is removed because exec is the problem and I will update the same in the commit message.

davidchisnall:
Thank you.  Please track this patch as one that can be reverted when #598 is fixed.

tests/ltp/patches/fix_sigrelse_sigrelse01.patch
shaikshavali1: This patch file will be removed/reverted once #598 is fixed.

tests/ltp/patches/fix_sigrelse_sigrelse01.patch
shaikshavali1: This patch is not using threads. Hence, the pipe related code is removed.

tests/ltp/patches/fix_sigrelse_sigrelse01.patch
shaikshavali1: uClinux code is removed because exec is the problem and I will update the same in the patch description.

shaikshavali1: I updated the patch description.

hukoyu:
@davidchisnall  if looks good can you approve so that I can continue to process this change. Thanks

hukoyu:
We are making test changes in the https://github.com/lsds/ltp/sgx-lkl now.
This PR resubmitted here: https://github.com/lsds/ltp/pull/2

PR #678, opened 2020-07-26
LTP: fixed fallocate01 and fallocate03 testcases

vamsikrishna935:
Modified the tests to use root filesystem.

### DIFF
--- a/tests/ltp/patches/fallocate01.patch
+++ b/tests/ltp/patches/fallocate01.patch
@@ -1,0 +1,46 @@
++ Failure reason:
++ fallocate03    1  TCONF  :  fallocate03.c:251: fallocate system call is not implemented
++ fallocate03    2  TCONF  :  fallocate03.c:251: Remaining cases not appropriate for configuration
+
++ Patch Description: Modified the tests to use root filesystem. 
+diff --git a/testcases/kernel/syscalls/fallocate/fallocate01.c b/testcases/kernel/syscalls/fallocate/fallocate01.c
+index 02aa33100..803f4fbc9 100644
+--- a/testcases/kernel/syscalls/fallocate/fallocate01.c
++++ b/testcases/kernel/syscalls/fallocate/fallocate01.c
+@@ -116,6 +116,7 @@ int fd_mode1, fd_mode2;
+ int TST_TOTAL = 2;
+ loff_t block_size;
+ int buf_size;
++char *tempdir = "tempdir";
+
+ /******************************************************************************
+  * Performs all one time clean up for this test on successful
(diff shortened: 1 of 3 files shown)


tests/ltp/patches/fallocate01.patch
```diff
@@ -0,0 +1,46 @@
++ Failure reason:
++ fallocate03    1  TCONF  :  fallocate03.c:251: fallocate system call is not implemented
++ fallocate03    2  TCONF  :  fallocate03.c:251: Remaining cases not appropriate for configuration
```
davidchisnall, line 3: Why is there a patch of fallocate01 talking about fallocate03 failures?

tests/ltp/patches/fallocate03.patch
```diff
@@ -0,0 +1,40 @@
...
+
++char *tempdir = "tempdir";
+ char *TCID = "fallocate03";
+ char fname[255];
+ int fd;
+@@ -147,8 +148,8 @@ void cleanup(void)
+        if (close(fd) == -1)
+                tst_resm(TWARN | TERRNO, "close(%s) failed", fname);
+
+-       tst_rmdir();
```
davidchisnall, line 22: Why is this changed?  We seem to be patching every file that uses `tst_rmdir`, rather than modifying what `tst_rmdir` and friends do in a single place.

davidchisnall (changes requested):
We seem to be patching a lot of tests in identical ways, why are we not centrally fixing the helpers that they are all using?

As before, please track that these centrailly somewhere (and, ideally, on the related issue) so that we can revert the changes when we have fixed the problem that required the test to be patched.  This is not virtual-memory related, so we should eventually be able to make it work.

davidchisnall:
Please re-raise this against the lsds/ltp repo.  Please also address the comments about patching individual tests rather than generic infrastructure.  A lot of tests seem to use the same helpers to use a different filesystem, we should be able to just fix those helpers rather than modifying every single test that relates to them.

vamsikrishna935:
> 
> 
> Please re-raise this against the lsds/ltp repo. Please also address the comments about patching individual tests rather than generic infrastructure. A lot of tests seem to use the same helpers to use a different filesystem, we should be able to just fix those helpers rather than modifying every single test that relates to them.

@davidchisnall  In very few tests, we have changed the temporary directory creation, changing this method will effect existing tests. Could you please let us if we can create another helper method

davidchisnall:
Does changing the generic code cause test failures?  If not, then do that.  If so, add a helper that does what we need for these changes.

Issue #634, opened 2020-07-15
LTP: Fix the patch file names with one pattern


hukoyu:
LTP patch filenames under tests/ltp/patches doesn't have a unique pattern. Fix them with one file name pattern with test name.

hukoyu:
PR https://github.com/lsds/sgx-lkl/pull/679 submitted to fix this issue

PR #679, opened 2020-07-27
LTP: Fix patch filenames with a pattern

hukoyu:
This addresses issue https://github.com/lsds/sgx-lkl/issues/634
We didn't have a standart file name pattern for LTP test patch file names. 

For a syscall test let's say abc, the patch file name will be abc.patch
Fixed all existing patch file names under tests/ltp/patches according to this pattern.
Also there were a few patch files contained multiple patches, split it to separate patch files for each test.

Will enforce this patch file name standard for new PRs.


letmaik (approved):
That's great, thanks a lot for doing this! :)

You also made some changes to the patches, is that on purpose?

letmaik:
Nevermind, I should read the PR description properly.... Monday morning!
All good from my side.

davidchisnall:
Thanks for doing this.  Can you describe somewhere how you are tracking the related patches so that we can revert them easily when we extend the set of features that SGX-LKL supports (e.g. vfork + exec, larger kernel memory so the loopback mounts work, and so on)?

SeanTAllen:
This is awesome.

SeanTAllen (approved):

davidchisnall:
@hukoyu / @prp, would it be easier for tracking if we replaced these patches entirely with a submodule that's a fork of the LTP tests?  We would then have a separate issue tracker for the LTP tests, which would give us a place to track all of them, and would make it easier to send PRs upstream for patches where we're doing things that are general cleanup and would be useful.

SeanTAllen:
At least for me, a submodule would make debugging problems with failing tests easier. Not enough to say "that's a reason to make the switch", but it would ease a bit of pain in the workflow of figuring out failing tests.

prp:
I agree that having a forked submodule would be easier to maintain than patches.

davidchisnall:
@prp, please can you create a fork?  @hukoyu, let's have a discussion about the processes for this.

prp:
> @prp, please can you create a fork? @hukoyu, let's have a discussion about the processes for this.

@davidchisnall done

hukoyu:
> > @prp, please can you create a fork? @hukoyu, let's have a discussion about the processes for this.
> 
> @davidchisnall done

@davidchisnall , @prp we are using a specific tag/branch of LTP: 
https://github.com/lsds/sgx-lkl/blob/oe_port/tests/ltp/buildenv.sh#L8
https://github.com/lsds/sgx-lkl/blob/oe_port/tests/ltp/buildenv.sh#L44

We are applying patches to specific branch to make sure upstream changes doesn't break our changes. Should we create our forked branch from this specific tag? 

I agree submodule approach will be better. Let's plan this and also applying patches to submodule. This might be a good opportunity to revise our patches as well.

hukoyu:
@davidchisnall yes we need a tracking system about enabling LTP tests and what to do when a specific issue fixed or a feature implemented. We can create a REDAME.md file in tests/ltp/ is that good?

hukoyu:
@davidchisnall @prp @letmaik  @paulcallen @jxyang  @Pengpeng-Microsoft  @SeanTAllen 
Submitted PR https://github.com/lsds/sgx-lkl/pull/682 for using LTP fork instead of patch mechanism.

PR #688, opened 2020-07-29
Fix infinite loop in JSON parser triggered by comments. Fixes #684.

### LINKED ISSUE #684: json comment will hang program execution
Comment is not officially supported in JSON. If comments has been put in the `enclave-config.json` file, application run through sgx-lkl-run-oe will hang at early stage, enclave init and other kernel information will not be printed out. 
sample problematic `enclave-config.json` file:
```json
{
    "args": [
        "app" // comments
    ]
}
```
The expected result should be exit w/ error code.

wintersteiger:
The JSON parser would hang when it sees a comment (`//`, i.e. invalid characters) inside an array, where it would keep adding `_strtol`'d numbers over 0 characters (i.e. `start == end`).

### DIFF
--- a/src/libjson/json.c
+++ b/src/libjson/json.c
@@ -1076,7 +1076,7 @@ static json_result_t _get_number(
         un->real = _strtod(start, &end, parser->options.allow_whitespace);
     }
 
-    if (!end || end != parser->ptr)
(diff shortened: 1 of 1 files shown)


mikbras (approved):
Thank you for doing this.

PR #689, opened 2020-07-29
Set memory as PROT_NONE when unmapping it

SeanTAllen:
As a defensive, debug only mechanism, we set memory to PROT_NONE when
munmapping it. This might, in debug-mode, surface otherwise hard to find
sgx-lkl bugs.


SeanTAllen:
This is currently branched off seantallen/no-mmap-fallback, #655 should be reviewed and merged before this is opened for review.

src/enclave/enclave_mem.c
```diff
@@ -456,6 +453,9 @@ int enclave_munmap(void* addr, size_t length)
     used_pages -= occupied_pages;
 
     bitmap_clear(mmap_bitmap, index_top, pages);
+#if DEBUG
```
SeanTAllen, line 456: These 3 lines are the only part of this PR that isn't part of #655.

SeanTAllen:
@davidchisnall @prp with the fallback removed, doing the PROT_NONE still causes a lot of failures on shutdown. Is this something we want to move forward with doing? If yes, then I'll create an issue for this which can be triaged and prioritized accordingly.

See https://dev.azure.com/sgx-lkl/sgx-lkl/_build/results?buildId=1548&view=results for CI results.

PR #692, opened 2020-07-30
Label all newly opened issues as "needs-triage"

SeanTAllen:
This commit adds a GitHub action to label all newly opened issues with the "needs-triage" label.
This is the simplest way forward with automating this. This uses a 3rd party GitHub action that is pinned
to specific SHA hash.

A larger project would be to replace this with a custom written GitHub action that runs on a schedule using
the GitHub actions cron facilities. That action would iterate through all issues in the repo that don't have a `p*` priority 
label and add the `needs-triage` label. 

This approach which is much simpler and doesn't require any real investment of time.

For this to work, if GitHub actions (specifically running 3rd party ones) aren't enabled for the lsds/sgx-lkl repo,
they will need to be turned on.

### DIFF
--- a/.github/label-issues.yml
+++ b/.github/label-issues.yml
@@ -1,4 +1,3 @@
-
 name: automate issue labels
 
 on:
--- a/.github/label-issues.yml
+++ b/.github/label-issues.yml
@@ -1,0 +1,15 @@
+
+name: automat issue labels
+
+on:
+  issues:
+    types: [opened]
+
+jobs:
+  automate-issue-labels:
+    runs-on: ubuntu-latest
+    steps:
+      - name: needs-triage labeling
+        uses: andymckay/labeler@467347716a3bdbca7f277cb6cd5fa9c5205c5412
+        with:
+          add-labels: "needs-triage"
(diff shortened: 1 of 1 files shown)


letmaik (approved):

davidchisnall (commented):
Looks good to me. Can we also set up a triage project and ass everything there, with columns for ‘proposed p0’, and so on so that we can move things around while we’re discussing their priority and then assign the labels and remove the, from the triage board once they’re done?

davidchisnall:
I have created a triage board for us to use. It looks as if this action can add things to the board.

SeanTAllen:
@davidchisnall the particular action in use here can change labels when a project card is moved. it can't add to a board.

If using a project board, we'd want an action that adds new issues to the project, then using github project automation, we could turn on the "to do" preset column which allows you to select to add all new issues and prs to the "to do" column. that would replace the triage column.

davidchisnall:
Ah, I misread the readme on the action.  I've tweaked the project's automation thingy so that new issues all appear in the needs triage column.

jxyang (approved):

PR #699, opened 2020-07-30
Test framework some house cleanup

hukoyu:
Some test framework cleanup work:

1) Moving disabled_tests.tx to other/ folder which is better place for it
2) Making nightly_build boolean which makes more sense

cc @letmaik


letmaik (approved):
Thanks for doing this, I really appreciate cleanup work! 👍

davidchisnall (approved):

PR #703, opened 2020-07-31
Remove dead enclave futex code

SeanTAllen:
enclave_futex is only used from posix-host.c. None of the code removed
in this PR is live.

Without the fallthroughs needed by dead code, I decided that an if/else
block would be better than a switch so I made some changes to structure
in enclave_futex after the changes.


src/sched/futex.c
```diff
@@ -243,7 +237,7 @@ static int futex_wake(int* uaddr, unsigned int num, uint32_t bitset)
 
     SLIST_FOREACH_SAFE(fq, &futex_queues, entries, tmp)
     {
-        if (fq->futex_key == futex_key && fq->futex_bitset & bitset && w < num)
+        if (fq->futex_key == futex_key && w < num)
```
davidchisnall, line 240: This line makes me sad because it makes me try to remember C operator precedence rules.  Not essential for this PR, but some brackets would be nice.

src/sched/futex.c
```diff
@@ -274,62 +268,45 @@ int enclave_futex(
...
-            rc = futex_wake(uaddr, val, bitset);
-            break;
-        default:
-            FUTEX_SGXLKL_VERBOSE("futex invalid op: %d\n", op);
-            rc = -ENOSYS;
+        rc = futex_wake(uaddr, val);
+    }
+    else
+    {
+        FUTEX_SGXLKL_VERBOSE("futex invalid op: %d\n", op);
```
davidchisnall, line 298: Let's just `sgxlkl_fail` here.  We tried to do a futex op that we shouldn't be doing.  Bad us!  That avoids some errors from things failing to check the return value of this.

davidchisnall (approved):
A couple of small tweaks would be nice.

Not necessary for this PR, but it would be nice to just expose `enclave_futex_wait`, `enclave_futex_wake` and `enclave_futex_timedwait` as the interfaces here, rather than the little conditional.  The generated code should be simpler as a result and the calls easier to understand.

src/sched/futex.c
SeanTAllen: I'll do another PR to make that change.

SeanTAllen:
@davidchisnall That is a good idea. I don't want to do as part of this PR, but a general clean-up PR I can do after this is merged.

vtikoo (approved):
LGTM

PR #706, opened 2020-08-02
Make LTP a submodule

hukoyu:
While creating image for LTP tests, we were clonning LTP code from  lsds/ltp.
We are making LTP a submodule and it is copied into image instead of cloning from repo.

Also excluding ltp from code check since it was causing check-ci to fail and we exluded them.

### DIFF
--- a/scripts/format-py-code
+++ b/scripts/format-py-code
@@ -51,4 +51,4 @@ if [[ ${verbose} -eq 1 ]]; then
     args+=("--verbose")
 fi
 
-"${cmd[@]}" "${args[@]}" "$this_dir/.."
+"${cmd[@]}" "${args[@]}" "--exclude" 'ltp/|lkl/|openenclave/|host-musl/|sgx-lkl-musl/|build/|build_musl/|third_party/' "$this_dir/.."
--- a/scripts/format-py-code
+++ b/scripts/format-py-code
@@ -51,4 +51,4 @@ if [[ ${verbose} -eq 1 ]]; then
     args+=("--verbose")
 fi
 
-"${cmd[@]}" "${args[@]}" "--exclude" 'ltp/|lkl/|openenclave/|host-musl/|sgx-lkl-musl/|build/|build_musl/|third_party/' "$this_dir/.."
+"${cmd[@]}" "${args[@]}" "$this_dir/.."
--- a/tests/ltp/Makefile
+++ b/tests/ltp/Makefile
@@ -29,6 +29,7 @@ $(ROOT_FS): $(ALPINE_TAR) buildenv.sh
 	$(ESCALATE_CMD) mount -t ext4 -o loop "$@" $(MOUNTPOINT)
 	$(ESCALATE_CMD) tar -C $(MOUNTPOINT) -xvf $(ALPINE_TAR)
 	$(ESCALATE_CMD) cp /etc/resolv.conf $(MOUNTPOINT)/etc/resolv.conf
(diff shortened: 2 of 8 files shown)


hukoyu:
Instead of fixing coder format issues for LTP, I excluded ltp folder for code check as recommended by @letmaik  and @davidchisnall 

cc @SeanTAllen @letmaik @paulcallen @davidchisnall @prp

davidchisnall (approved):

scripts/format-py-code
```diff
@@ -51,4 +51,4 @@ if [[ ${verbose} -eq 1 ]]; then
     args+=("--verbose")
 fi
 
-"${cmd[@]}" "${args[@]}" "$this_dir/.."
+"${cmd[@]}" "${args[@]}" "--exclude" 'ltp/|lkl/|openenclave/|host-musl/|sgx-lkl-musl/|build/|build_musl/|third_party/' "$this_dir/.."
```
letmaik: This is not the right place. Please look at https://github.com/lsds/sgx-lkl/blob/oe_port/pyproject.toml

paulcallen (approved):

scripts/format-py-code
paulcallen: I think excluding the .git directory in the LTP submodule would make it significantly quicker

hukoyu: @paulcallen I am already excluding ltp folder and which means it will skip any subfolder as well.

hukoyu:
> Please check the CI times before doing such changes. This increases checkout time by 2 min for every test job. It would be better to change this to a sequence of rm -rf (like in the build job), checkout without submodules, manual submodule checkout of LTP.

@letmaik you are absolutely right. I didn't realize it would take so much time to checkout all submodules. I reverted my change that checksout all submodules. I am checking out LTP submodule only.

scripts/format-py-code
hukoyu: @letmaik I moved ltp exclusion to pyproject.toml file. I wasn't aware of such a file. Done.

tests/ltp/Makefile
```diff
@@ -29,6 +29,10 @@ $(ROOT_FS): $(ALPINE_TAR) buildenv.sh
 	$(ESCALATE_CMD) mount -t ext4 -o loop "$@" $(MOUNTPOINT)
 	$(ESCALATE_CMD) tar -C $(MOUNTPOINT) -xvf $(ALPINE_TAR)
 	$(ESCALATE_CMD) cp /etc/resolv.conf $(MOUNTPOINT)/etc/resolv.conf
+	git submodule init $(SGXLKL_ROOT)/ltp
+	git submodule update --remote $(SGXLKL_ROOT)/ltp
+	$(ESCALATE_CMD) mkdir $(MOUNTPOINT)/ltp
+	$(ESCALATE_CMD) cp -R $(SGXLKL_ROOT)/ltp/* $(MOUNTPOINT)/ltp/
```
paulcallen, line 35: why isnt this just a symbolic link?

hukoyu: @paulcallen here we are copying into image through mounted folder. I am not sure if image will work with symlink and also we umount after copying ltp/ folder into image. LTP tests run from image directly.

hukoyu: We pass a fresh image file for running each LTP test and using symlink might not work for this. We will keep as is and it is already super fast. If we think we need such optimization we can submit another PR with necessary code changes with symlink usage and good testing to make sure nothing wil be broken. Merging as is.

PR #707, opened 2020-08-03
Fix typo or thing previously missed in changes

KenGordon:
Trivial fix to readme as a test of workflow.


wintersteiger (approved):
That's right, thanks!

paulcallen (approved):

PR #708, opened 2020-08-03
Remove enclave_futex and narrow the enclave provided futex API

SeanTAllen:
The prior version of enclave_futex was left over from when we had a
"complete" futex implementation in the enclave. At one point, it was
used as the futex implementation for everything flowing through musl.

That was changed, however, enclave_futex lived on and continued to
accept an operation to perform (e.g. FUTEX_WAKE). Not every operation
from the futex API was guaranteed to work.

To make matters worse, if the caller didn't check the result of
enclave_futex, it is possible they would miss that nothing was done
for unimplemented futex operations.

This commit removes enclave_futex and replaces it with 3 targeted
API calls enclave_futex_wake, enclave_futex_wait, and enclave_futex_timedwait
between the 3, we have complete coverage of all futex functionality currently
used to support internal enclave functionality.


davidchisnall (approved):

src/lkl/posix-host.c
```diff
@@ -34,14 +34,16 @@
 
 #define NSEC_PER_SEC 1000000000L
 
-// The function used to implement the futex system call on top of lthreads
-int enclave_futex(
+// The functions used to implement the futex system call on top of lthreads
```
vtikoo: Does this comment still hold?

SeanTAllen: It does at the moment. I updated to reflect the refactoring change here.

SeanTAllen: I can see how that would be misleading.

SeanTAllen: Removed

paulcallen (approved):
Looks good to me

Issue #179, opened 2020-05-05
Move mmap-related syscall bypasses into the kernel


davidchisnall:
SGX-LKL provides its own implementation of the mmap family of APIs and musl calls these directly instead of via the system call layer.  This can't be done until we've removed all malloc / mmap dependencies from code that runs before LKL starts.

The relevant system calls are:

- [ ]  `mmap`
- [ ]  `munmap`
- [ ]  `mremap`
- [ ]  `msync`
- [ ]  `mprotect`

Currently, `mprotect` calls the host directly from userspace.  This call should be moved into the kernel.

davidchisnall:
@SeanTAllen, please close this if it's done.

SeanTAllen:
I don't believe this is done until #604 is done.

SeanTAllen:
Closed by #710.

PR #710, opened 2020-08-03
Add memory system call replacements

SeanTAllen:
These system calls are used for heap setup and so we can't yet remove
the musl wrappers, but this provides the kernel support that we'll need
when we do.

### DIFF
--- a/src/enclave/enclave_mem.c
+++ b/src/enclave/enclave_mem.c
@@ -14,9 +14,11 @@
 
 #include "enclave/bitops.h"
 #include "enclave/lthread.h"
+
 #include "enclave/enclave_mem.h"
 #include "enclave/enclave_util.h"
 #include "enclave/lthread_int.h"
+#include "enclave/sgxlkl_t.h"
 #include "enclave/ticketlock.h"
 
 static struct ticketlock mmaplock;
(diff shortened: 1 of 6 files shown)


SeanTAllen:
This is the code from #604 rebased and updated to work with the changes since I put it down after taking over debugging from @davidchisnall.

SeanTAllen:
@davidchisnall this replaces #604. can you give this a decently thorough review to make sure when I brought the rest of the changes over from #604 that didn't make it into other mmap related PRs from the last week that I didn't mess anything up? I think I got it all but I'm not 100% positive. CI is passing so that is good.

SeanTAllen:
@vtikoo I asked for you to review this as you signed off on the original #604.

SeanTAllen:
The branch `davidchisnall/mmap` should be deleted once this is merged. I didn't want to delete it before this was merged in case it was needed for comparison.

https://github.com/lsds/sgx-lkl/pull/604

SeanTAllen:
@davidchisnall isn't there some issue that gets closed by this? I can't find it if there is.

davidchisnall:
#179 ?

SeanTAllen:
@davidchisnall yes, #179. When this is merged, I will review #179 and check the appropriate boxes.

src/enclave/enclave_mem.c
```diff
@@ -343,6 +237,7 @@ void* enclave_mmap(
     if (((intptr_t)ret) >= 0)
     {
         int found_only_fresh_pages = 0;
+        int mprotect_ret;
```
vtikoo, line 240: Is this placed just to satisfy the function signature for `sgxlkl_host_syscall_mprotect`?

SeanTAllen: Yes. You need to provide pointer to something that can be used for the return value. prior to David's work here, the return value was ignored. So it is with this.

vtikoo (approved):
lgtm!

davidchisnall (approved):

PR #711, opened 2020-08-03
Change posix-hosts to have consistent indentation style

### LINKED ISSUE #665: src/lkl/posix-host.c contains a mix of tabs and spaces for indentation
It should be switched to all spaces.

SeanTAllen:
Somewhere along the line, someone did a format to spaces from tabs.
Probably using clang format. This particular file because it could
be upstreamed is supposed to be tabs, not spaces.

Closes #665

### DIFF
--- a/src/lkl/posix-host.c
+++ b/src/lkl/posix-host.c
@@ -657,13 +657,13 @@ static size_t kernel_mem_size;
  */
 static void *host_malloc(size_t size)
 {
-	// If we're allocating over 1MB, we're probably allocating the kernel heap.
-	// Pull this out of the enclave mmap area: there isn't enough space in the
-	// OE heap for it.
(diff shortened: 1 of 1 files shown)


davidchisnall:
I think it's probably fine to do consistent formatting, rather than special-case this file.  We've rewritten pretty much all of it, so I doubt we will ever want to diff with upstream.

SeanTAllen:
@prp @davidchisnall in that case I will switch this entirely to spaces at it was mixed.

SeanTAllen:
@davidchisnall @prp updated to be spaces as consistent formatting. I left the PR comment the same as before the switch (so that the conversation here will make sense in the future) but updated the commit comment.

davidchisnall (approved):

paulcallen (approved):

PR #717, opened 2020-08-04
Update sgx-lkl-musl

SeanTAllen:
Update sgx-lkl-musl to use a version that uses mmap rather than
enclave_mmap in expand_heap. This moves sgx-lkl-musl one more custom
patch away from being different from musl.


SeanTAllen:
Corresponding sgx-lkl-musl PR that would also need sign-off: https://github.com/lsds/sgx-lkl-musl/pull/34

PR #722, opened 2020-08-05
Fix memory management for cloned threads.

davidchisnall:
Most of this fix is in the LKL changes but now that LKL is calling join
on the resulting threads, we need to not detach them.

### DIFF
--- a/src/lkl/posix-host.c
+++ b/src/lkl/posix-host.c
@@ -343,8 +343,6 @@ static lkl_thread_t thread_create_host(void* pc, void* sp, void* tls, struct lkl
     // mechanism begin life in the kernel and so need to be associated with the
     // kernel task that created them.
     lthread_setspecific_remote(thread, task_key->key, task_value);
-    // Detach the thread.  Any join operations will be handled by LKL.
-    lthread_detach2(thread);
     // Mark the thread as runnable.  This must be done *after* the
     // `lthread_setspecific_remote` call, to ensure that the thread does not
     // run while we are modifying its TLS.
--- a/tests/basic/clone/Dockerfile
+++ b/tests/basic/clone/Dockerfile
@@ -4,7 +4,9 @@ RUN apk add --no-cache gcc musl-dev
 
 ADD *.c /
 RUN gcc -fPIE -pie -o clone clone.c -g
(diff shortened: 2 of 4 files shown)


davidchisnall:
Note: This points to the pre-merge LKL branch, I will update it to point to the merged version once the merge is done.

prp:
It's worth adding @vtikoo's clone-in-a-loop test here and see the CI output.

davidchisnall:
I am basically in meetings until the end of the day and then off until Monday, but I'm happy for someone else to add things to this to the branch.

prp:
@vtikoo could you add your clone-in-a-loop test as a second test under `tests/basic/clone` here? Thanks.

SeanTAllen (approved):

SeanTAllen:
@prp @vtikoo can we do as a different clone test rather than replacing the existing?

prp:
Yes, that's what I mean. We can have the one Makefile under `tests/basic/clone` run the two different tests in sequence. We have done the same for other test directories.

vtikoo:
@prp could you provide a reference to such a test setup?

prp:
> @prp could you provide a reference to such a test setup?

@vtikoo you can find an example [here](https://github.com/lsds/sgx-lkl/blob/oe_port/tests/basic/exit/Makefile). Actually, this is not particularly clean, and it would be great if you could use a more scalable template for multiple tests. (In this example, tests aren't separate Makefile targets, and the naming isn't applied consistently.)

davidchisnall:
The current clone test will need removing at some point because it assumes it's safe to call libc functions in threads that have not done any libc setup (including `syscall`, which tries to write to `errno` on return, which causes breakage with Vikas' branch), so I don't mind just replacing it with the other test and then merging the relayering work, which then makes all of the pthread tests into clone tests.

vtikoo:
The core issue with clone test is that it doesn't set up its TLS/TP as `pthread_create` would. This causes the errno and strerror failures.
However there seems to be value to keep it around from a debugging perspective, as its without any libc interactions.

vtikoo (approved):

prp (approved):
LGTM

PR #727, opened 2020-08-05
Ken/issue297

### LINKED ISSUE #297: [Tests] lkl_access_ok() should return failure on test invalid access, not fatal exit.

KenGordon:
Fix for issue 297. This was caused by the code that checks for bad memory addresses exiting with a fatal error rather than just logging as appropriate and telling the caller that the memory is not valid.

One line fix as suggested by David.

The complication was turning on and properly reverting all the tests in ltp which would trigger the bug.  The commits were not entirely clear of other changes.
[Test] system calls with NULL causes abort #169 is a duplicate of 297. Some cases were labelled with TODOs for that one.
See the LTP PR to match is https://github.com/lsds/ltp/pull/51 against branch ken/issue297ltp


hukoyu (approved):
LGTM

hukoyu:
@KenGordon  @SeanTAllen  a question here. In this PR we are bumping LTP submodule commit. But these files are part of lsds/ltp PR https://github.com/lsds/ltp/pull/51 which is not merged yet. So how this works? Should we merge https://github.com/lsds/ltp/pull/51 first and then update the ltp submodule here and then merge this PR? Or it doesn't matter https://github.com/lsds/ltp/pull/51 is merged or not and this ltp submodule bump is good enough?

SeanTAllen:
@hukoyu this will need to be rebased manually first. generally the workflow is to merged the corresponding submodule commit and then merge the sgx-lkl commit.

if no squashing was involved for the submodule commit this works fine. if the submodule commit involves squashing then the PR commit first needs to be updated to the squashed commit for the submodule.

davidchisnall:
@SeanTAllen: Minor correction, GitHub never does fast-forward merges (even if you say rebase) because it adds PR metadata to the final version of the commit.  This means that the commit hash will always be different.  The two workflows that work are:

 - Merge the submodule commit, then update the PR to include the new hash and merge that.
 - Merge this pointing at its current git hash then make sure that those changes are in the submodule head so that the next submodule update will pull them in as long as it's based on the current main branch.

Neither is particularly great.

davidchisnall (approved):

davidchisnall:
@KenGordon, please fix the conflicts and then do the merge dance outlined above.

PR #731, opened 2020-08-06
Fixes #730. Call stacks dumps corrupt mem

### LINKED ISSUE #730: sgxlkl_print_backtrace is calling wrong deallocator for freeing symbols
`sgxlkl_print_backtrace` is calling `oe_free` to deallocate symbols. This is wrong. It should be calling `oe_backtrace_symbols_free`. This function does not call oe_free, rather `oe_allocator_free`, but we should use `oe_backtrace_symbols_free` anyway incase implementation changes.
This sometimes causes ltreads to not get dumped when a panic in sgx-lkl happens in the test pipeline so we do not get all the information we need to determine the failure location.
Fix is trivial!
Its important because it makes it difficult to track down issues occasionally

paulcallen:
stack dumps were freeing memory incorrectly for symbols. Now call
correct function.
This caused certain failures to not display the correct lthread dumps.

### DIFF
--- a/src/enclave/enclave_util.c
+++ b/src/enclave/enclave_util.c
@@ -82,6 +82,7 @@ void* oe_calloc_or_die(size_t nmemb, size_t size, const char* fail_msg, ...)
  * other lthreads.
  */
 extern int oe_backtrace_impl(void** start_frame, void** buffer, int size);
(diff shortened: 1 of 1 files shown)


prp (approved):
LGTM

vtikoo (approved):

PR #735, opened 2020-08-06
Don't overwrite LTP tests when doing make clean

SeanTAllen:
Prior to this change when doing `make clean` in the `ltp` directory,
it would resync the ltp git submodule. This is highly problematic
for any workflow where you would need to work with a different commit
of the ltp submodule or if you want to make changes in tree to test.

For example, if debugging a problem with an ltp test.

You modify the ltp test.
Run make clean and go to run the given test.

Prior to this change, all your modifications to the ltp test would be
overwritten.

As of this commit, the ltp submodule will only be sync'd if it hasn't
already been pulled down. This should keep existing workflows operating
while also making submodule checkout for ltp work the same as those for
sgx-lkl-musl etc.

### DIFF
--- a/tests/ltp/Makefile
+++ b/tests/ltp/Makefile
@@ -16,21 +16,24 @@ LTP_TEST_EXEC_TIMEOUT=3600
 LTP_TEST_SCRIPT="./run_ltp_test.sh"
 ESCALATE_CMD=sudo
 
+LTP_SOURCE_DIR=$(SGXLKL_ROOT)/ltp
+
 .DELETE_ON_ERROR:
 .PHONY: all clean
+
+$(LTP_SOURCE_DIR)/.git:
+	git submodule update --progress --init $(LTP_SOURCE_DIR)
 
 $(ALPINE_TAR):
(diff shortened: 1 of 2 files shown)


hukoyu (approved):
LGTM. Just for the record LTP tests in pipeline in folders ltp-batch1 and ltp-batch2 uses batch.mk
So ltp pipeline tests don't use Makefile. This is kept only for manual testing and any change we make to batch.mk, we also do to Makefile, trying to keep them in sync. But this change is needed as @SeanTAllen  said after ltp submodule change.

SeanTAllen:
@hukoyu that's a good point. I should update batch.mk as well otherwise there will be very different results depending on the directory you run make from. I'm going to update this.

hukoyu:
> @hukoyu that's a good point. I should update batch.mk as well otherwise there will be very different results depending on the directory you run make from. I'm going to update this.

@SeanTAllen with current test framework, make is supposed to be run in the folder that has Makefile for non-ltp tests (All folders under tests/ except /tests/ltp). 

We have three modes can be passed to test_runner.sh : core, ltp1 and ltp2.
For ltp1 there will be only one Makefile in list: tests/ltp/ltp-batch1/Makefile
For ltp2 there will be only one Makefile in list: tests/ltp/ltp-batch2/Makefile
For core all Makekfiles under tests/ recursively except tests/ltp. 

To sum up, current assumption is to run make in the folder that has Makefile.

hukoyu (approved):
LGTM

davidchisnall (approved):

PR #741, opened 2020-08-07
LTP: Added a support to mount file system in tests

shaikshavali1:
Problem:
This test case causing oom killer to be invoked and causing
kernel panic. This is because the test case invokes test framework
function "tst_acquire_device" which creates a 256MB of .img file
and binds with a loop device. This is not supported because the
default size of LKL memory is set to 32M. This is kept low due to
the limited size of EPC (Enclave page cache) and to avoid page
cache being swapped out.
Also, the test case invokes "tst_mkfs" framework function which
inturn invokes mkfs utility command with the help of system() syscall.
It is recommended to use root file system, but, this test case
needs a read-only filesystem to test one of the sub-test case.
we don't have a read-only file system mounted in sgx-lkl.

To support the a formated ext4 filesystem image created in master
.img. Test case will pick this image and use it in test.
Test case changes link: https://github.com/lsds/ltp/pull/49

### DIFF
--- a/tests/ltp/Makefile
+++ b/tests/ltp/Makefile
@@ -36,6 +36,9 @@ $(ROOT_FS): $(ALPINE_TAR) buildenv.sh
 	$(ESCALATE_CMD) install buildenv.sh $(MOUNTPOINT)/usr/sbin
 	$(ESCALATE_CMD) chroot $(MOUNTPOINT) /sbin/apk update
 	$(ESCALATE_CMD) chroot $(MOUNTPOINT) /sbin/apk add bash
+	$(ESCALATE_CMD) mkdir $(MOUNTPOINT)/ltp_tst_mnt_fs
+	$(ESCALATE_CMD) dd if=/dev/zero of=$(MOUNTPOINT)/ltp_tst_mnt_fs/tstfs_ext4.img count=256 bs=1M
+	$(ESCALATE_CMD) mkfs -t ext4 $(MOUNTPOINT)/ltp_tst_mnt_fs/tstfs_ext4.img
 
 
 	$(ESCALATE_CMD) chroot $(MOUNTPOINT) /bin/bash /usr/sbin/buildenv.sh 'build' '/ltp/testcases/kernel/syscalls'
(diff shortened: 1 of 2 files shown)


hukoyu (changes requested):
@shaikshavali1  pipeline uses tests/ltp/batch.mk for running LTP tests that included in Makefiles in ltp-batch folders. Please make the same changes in batch.mk as well.

tests/ltp/Makefile
```diff
@@ -36,6 +36,9 @@ $(ROOT_FS): $(ALPINE_TAR) buildenv.sh
 	$(ESCALATE_CMD) install buildenv.sh $(MOUNTPOINT)/usr/sbin
 	$(ESCALATE_CMD) chroot $(MOUNTPOINT) /sbin/apk update
 	$(ESCALATE_CMD) chroot $(MOUNTPOINT) /sbin/apk add bash
+	$(ESCALATE_CMD) mkdir $(MOUNTPOINT)/ltp_tst_mnt_fs
```
letmaik, line 39: This has to be freshly created for each test case. @hukoyu Is that right? I think that's the plan, but I lost track a bit.

hukoyu: @letmaik we create an image and then we use a fresh copy of this image for each LTP test. We don't create a new image for each test but we keep the original image and copy it as fresh image for each test.

letmaik: Ah alright, and this image would contain the disk which is created here.

shaikshavali1:
> @shaikshavali1 pipeline uses tests/ltp/batch.mk for running LTP tests that included in Makefiles in ltp-batch folders. Please make the same changes in batch.mk as well.

@hukoyu, updated the patch.

hukoyu (approved):
Pipeline and all tests passed. Test duration didn't increase. No regression due to this change that will enable mount file system in tests. It looks good. @letmaik  can you also review and approve if looks good.

letmaik (approved):
@hukoyu The duplication of code in batch.mk has to be removed, people constantly stumble over it. Could you open an issue for this?

tests/ltp/Makefile
hukoyu: @SeanTAllen  already submitted https://github.com/lsds/sgx-lkl/pull/746 to combine make files for ltp.

Issue #732, opened 2020-08-06
Segfault and PANIC inside LKL Enclave


Arslan8:
Hi,
While doing genration based fuzzing on existing SGX programs, we found that SGX-LKL encalve does not do proper checking on the following fields:

- #0  0x00007fe0005c93d2 in lkl_virtio_console_add (console=0x0) 
- args->shm->enc_dev_config
- args->shm->timer_dev_mem
- args->shm->virtio_blk_dev_mem
- args->shm->virtio_blk_dev_names
- args->shm->env 
- args->shm->virtio_swiotlb

The fuzzer works on the principle that arguments to enclave are coming from untrusted runtime and should be checked accordingly.

bodzhang:
This issue has security implication. Proposed to assign P1.

douglasmaciver:
@Arslan8 Kudos to you and those involved. This is good work. Please keep it up.

PR #743, opened 2020-08-07
Add shared memory checks 

wintersteiger:
This adds a number of checks that ensure pointers into shared memory from the host don't point into enclave memory.

Also adds a check to ensure the enclave is initialized only once.


wintersteiger:
Also fixes #732.

src/lkl/virtio_netdev.c
```diff
@@ -70,6 +70,16 @@ static void lkl_deliver_irq(uint64_t dev_id)
...
 {
     int ret = -1;
+
+    if (!netdev)
+        return -1;
+
+    if (!oe_is_outside_enclave(netdev, sizeof(struct virtio_dev)))
+        oe_abort();
+
+    // TODO: Ensure netdev->queue/ops/config_data/base are outside enclave on
```
wintersteiger: @prp I assume we will have to check queue/ops/config_data/base wherever they are accessed. If that's not so, please let me know so I don't go down unnecessary rabbit holes :)

src/enclave/enclave_oe.c
```diff
@@ -253,30 +254,55 @@ static void _read_eeid_config()
...
+    /* Temporary, volatile copy to make sure checks aren't reordered */
+    volatile sgxlkl_shared_memory_t* enc = oe_calloc_or_die(
+        1,
+        sizeof(sgxlkl_shared_memory_t),
+        "Could not allocate memory for shared memory\n");
 
     if (cfg->io.network)
+    {
         enc->virtio_net_dev_mem = host->virtio_net_dev_mem;
+        CHECK_OUTSIDE(enc->virtio_net_dev_mem, sizeof(struct virtio_dev));
```
bodzhang: `virtio_dev` structure conains many pointers to sub structures, which also contain pointers to sub structures. Are those pointers suppoed to point to buffers outside of enclave? How to check and how to make sure a malicious host can't modify the pointers after the check?

src/enclave/enclave_oe.c
```diff
@@ -303,20 +333,33 @@ static void _copy_shared_memory(const sgxlkl_shared_memory_t* host)
         }
     }
 
-    if (host->env)
+    /* Copy the host's environment variables to enclave memory */
+    char* const* henv = host->env;
```
bodzhang, line 337: Is `henv` supposed to point to a buffer outside of the enclave?

src/enclave/enclave_oe.c
```diff
@@ -303,20 +333,33 @@ static void _copy_shared_memory(const sgxlkl_shared_memory_t* host)
...
+    if (henv)
     {
         size_t henvc = 0;
-        while (host->env[henvc++] != 0)
-            ;
+        while (henv[henvc] != 0)
```
bodzhang: `henv` buffer is already accessed before CHECK_OUTSIDE(). If malicious host sets henv pointer to point to secret buffer inside the enlcave, side channel info about which byte within the secret buffer is 0 already leaked.

src/enclave/enclave_oe.c
wintersteiger: Even a friendly host will necessarily have to change the contents - that's what the shared memory is for. We will have to add checks on every memory access to any of those pointers in the lkl submodule.

src/enclave/enclave_oe.c
wintersteiger: That's correct and expected. Note that the `henv` pointer can't change anymore, but we don't know the size of the memory this points to; we have to search for the first `NULL`. The purpose of this is to import host-controlled information, i.e. the host is welcome to change the contents at any point. All the elements (environment variable settings) are checked again further below.

bodzhang: The potential side channel attack is not to change the `henv` pointer after `CHECK_OUTSIDE` line. The attacker's goal is to set `henv` to a secret buffer inside the enclave, then time the NULL search loop to see which byte within the secret buffer is 0.

wintersteiger: Thanks, you're right, I see what you mean now! I guess that could be closed by adding a check at every iteration of the loop, but I think I'd rather pass the length of the buffer (`henvc`) in as a separate parameter so we can check the location of the whole buffer up front.

wintersteiger: (Fixed)

src/enclave/enclave_event_channel.c
```diff
@@ -155,6 +155,10 @@ void initialize_enclave_event_channel(
     uint8_t* dev_id = NULL;
     _evt_channel_num = evt_channel_num;
 
+    if (!oe_is_outside_enclave(
+            enc_dev_config, sizeof(enc_dev_config_t) * _evt_channel_num))
+        oe_abort();
```
davidchisnall: `sgxlkl_fail` with an informative message?

src/enclave/enclave_event_channel.c
```diff
@@ -168,6 +172,17 @@ void initialize_enclave_event_channel(
     _enc_dev_config = enc_dev_config;
     for (int i = 0; i < evt_channel_num; i++)
     {
+        if (!oe_is_outside_enclave(
```
davidchisnall: These all look like they're vulnerable to TOCTOU issues.  From line 158, we've established that `enc_dev_config` is outside the enclave, now we're checking the fields while the host can be concurrently modifying them.  Shouldn't we copy this structure into a buffer in the enclave and then check and use the fields?

src/enclave/enclave_oe.c
```diff
@@ -253,38 +251,63 @@ static void _read_eeid_config()
     sgxlkl_enclave_state.config = cfg;
 }
 
+#define CHECK_INSIDE(X, S)                                  \
```
davidchisnall: Maybe move these macros into `enclave_util.h`?  These don't appear to include anything from the surrounding environment and so should be `static inline` functions, not macros.

Please also add a message argument and use `sgxlkl_fail` to print a useful error (or, better, stringify the arguments and use those in the error message - in which case it would have to remain a macro).

src/enclave/enclave_oe.c
davidchisnall: We should check the LKL virtio drivers do this.  This is the same use as CVMs, so if they don't then we have a fairly simple attack via VirtIO on Linux on SEV and similar technologies (e.g.  Google's recent offering).  We may need to add something at the bus layer for hooks that the drivers use to check that memory is the correct kind.  All of the descriptors that we use should be SWIOTLB-owned memory.

src/enclave/enclave_timer.c
```diff
@@ -28,7 +28,12 @@ _Atomic(uint64_t) internal_counter = 0;
  */
 uint64_t enclave_nanos()
 {
-    uint64_t e = sgxlkl_enclave_state.shared_memory.timer_dev_mem->nanos;
+    struct timer_dev* t = sgxlkl_enclave_state.shared_memory.timer_dev_mem;
+
+    if (!oe_is_outside_enclave(t, sizeof(struct timer_dev)))
```
davidchisnall: If the macros are moved into `enclave_util.h` they can be used here.

src/enclave/enclave_oe.c
```diff
@@ -303,20 +330,32 @@ static void _copy_shared_memory(const sgxlkl_shared_memory_t* host)
...
-            henvc + 1,
+            henvc,
             sizeof(char*),
             "Could not allocate memory for host import environment variable\n");
         for (size_t i = 0; i < henvc; i++)
-            tmp[i] = oe_strdup(host->env[i]);
-        tmp[henvc] = NULL;
+        {
+            const char* env_i = henv[i];
+            size_t n = strlen(env_i) + 1;
```
bodzhang: `strlen()` operates on `env_i` buffer already, before the check of whether the buffer is outside of enclave. String buffer is hard to manage in enclave interfaces, exactly because the data marshalling code inside the enclave has to operate on the buffer to calcuate the length. If the string buffer can be defined with a max length, the marshalling code can check the the max-length string buffer is outside of enclave before calling `strlen()`.

src/enclave/enclave_oe.c
bodzhang: Relevant dicussion at https://github.com/lsds/sgx-lkl/issues/745

src/enclave/enclave_oe.c
douglasmaciver: 295 `sizeof(enc_dev_config_t) * enc->evt_channel_num` may overflow.

douglasmaciver: Have you considered where the lfences should be placed?

src/enclave/enclave_event_channel.c
wintersteiger: Yes, that's right. According to @prp we can deep-copy all of `enc_dev_config`, so these checks won't be necessary, which is what I'll do now.

src/enclave/enclave_oe.c
```diff
@@ -311,38 +305,117 @@ static void _read_eeid_config()
...
+        {
+            const enc_dev_config_t* host_conf_i = &host->enc_dev_config[i];
+            enc_dev_config_t* enc_conf_i = &enc->enc_dev_config[i];
+
+            enc_conf_i->dev_id = host_conf_i->dev_id;
+
+            enc_evt_channel_t* host_chn = host_conf_i->enc_evt_chn;
+            if (host_chn)
+            {
+                enc_conf_i->enc_evt_chn = host_chn;
```
wintersteiger: @prp when I try to deep-copy `enc_evt_chn` (see commented block), the devices don't work anymore (execution hangs). So, the host must be changing that pointer (plus possibly the contents of the `enc_evt_channel_t`). Is this expected or does this indicate that something somewhere must be going wrong?

davidchisnall: The host shouldn't be changing the pointer to the event channels, but event channels are shared 64-bit values between enclave and guest and both will need to read / write the values.

wintersteiger: When I allocate a new `enc_evt_chn_t` and copy the contents from the host:

    enc_conf_i->enc_evt_chn = oe_calloc_or_die(...);
    *enc_conf_i->enc_evt_chn = *host_chn;

Then the devices don't work anymore, so I suspect the host does indeed change the pointer at some point. I'll dig a bit deeper.

davidchisnall: I think this is copying one too deep.  The `enc_evt_channel_t*` is a pointer to a shared 64-bit value.  If you copy the initial value, you no longer have an event channel, you just have an event.

src/enclave/enclave_event_channel.c
wintersteiger: Done

src/enclave/enclave_timer.c
wintersteiger: Done

src/enclave/enclave_oe.c
wintersteiger: We don't have max sizes for all strings, e.g. this particular one pointed at here is the content of a host-controlled environment variable. I think I can pass all of the variable in one string + size and tokenize afterwards.

The overflow is fine, in that case it would reserve too little memory and segfault. I'm not putting any lfences anywhere, that will have to be a separate effort.

src/enclave/enclave_oe.c
```diff
@@ -311,38 +305,87 @@ static void _read_eeid_config()
...
+            sizeof(enc_dev_config_t),
+            "Could not allocate memory for event channel config\n");
+
+        for (size_t i = 0; i < num_channels; i++)
+        {
+            const enc_dev_config_t* host_conf_i = &host->enc_dev_config[i];
+            enc_dev_config_t* enc_conf_i = &enc->enc_dev_config[i];
+            enc_conf_i->dev_id = host_conf_i->dev_id;
+            enc_conf_i->enc_evt_chn = host_conf_i->enc_evt_chn;
+            sgxlkl_ensure_outside(
```
bodzhang, line 349: Simply checking enc_evt_chn is outside of the enclave at this point is not enough, as the malicious host can change the host memory content after the check hear passes. Probably should analyze the data structure and keep a "shadow" structure inside the enclave and copy out enclave update to the host-read-only components from the "shadow " structure to the strcuture outside of the enclave.

wintersteiger: Yes, that's correct and expected; this check here is really redundant. The content of this data structure is shared memory that both host and enclave write to (all of the contents) and the important checks are in `enclave_event_channel.c`.

src/enclave/enclave_oe.c
```diff
@@ -351,8 +394,12 @@ static void _copy_shared_memory(const sgxlkl_shared_memory_t* host)
         for (size_t i = 0; i < enc->num_virtio_blk_dev; i++)
         {
             enc->virtio_blk_dev_mem[i] = host->virtio_blk_dev_mem[i];
```
bodzhang, line 371: As discussed in [issue 745](https://github.com/lsds/sgx-lkl/issues/745), a "shawdow" virtio_dev structure inside the enclave is needed.

wintersteiger: Yes, that will not be part of this PR. Someone with detailed knowledge of the virtio interface will have to work that out.

src/enclave/enclave_oe.c
wintersteiger: I turned them into functions so we can easily set breakpoints in them.

src/enclave/enclave_oe.c
wintersteiger: Yes, it doesn't do that anymore, but I had to add a bunch of checks in the event channel code.

src/enclave/enclave_event_channel.c
wintersteiger: (That turned out to be wrong, checks have moved around a bit.)

src/enclave/enclave_timer.c
```diff
@@ -28,7 +28,11 @@ _Atomic(uint64_t) internal_counter = 0;
  */
 uint64_t enclave_nanos()
 {
-    uint64_t e = sgxlkl_enclave_state.shared_memory.timer_dev_mem->nanos;
+    struct timer_dev* t = sgxlkl_enclave_state.shared_memory.timer_dev_mem;
+
+    sgxlkl_ensure_outside(t, sizeof(struct timer_dev));
```
bodzhang: Should do the memory check when intialize sgxlkl_enclave_state.shared_memory, instead of here, to avoid unncessary performace waste.

davidchisnall: Yes.

src/lkl/setup.c
```diff
@@ -1355,13 +1353,14 @@ static void init_enclave_clock()
 
     SGXLKL_VERBOSE("Setting enclave realtime clock\n");
 
-    if (oe_is_within_enclave(shm->timer_dev_mem, sizeof(struct timer_dev)))
+    struct timer_dev* t = shm->timer_dev_mem;
+
+    if (!oe_is_outside_enclave(t, sizeof(struct timer_dev)))
```
bodzhang: The check can be done when initalizing sgxlkl_enclave_state.shared_memory. Any advantage to do it here instead?

wintersteiger: I wasn't sure so went with the conservative solution. I think you're right though, it's safe to check that during initialization.

src/enclave/enclave_oe.c
```diff
@@ -311,38 +285,87 @@ static void _read_eeid_config()
...
+    if (ptr != NULL && !oe_is_within_enclave(ptr, sz))
+        sgxlkl_fail("Unexpected: buffer is not in enclave memory.\n");
+}
+
+void sgxlkl_ensure_outside(const void* ptr, size_t sz)
+{
+    if (ptr != NULL && !oe_is_outside_enclave(ptr, sz))
+        sgxlkl_fail("Unexpected: buffer is in enclave memory.\n");
+}
+
```
douglasmaciver: Regarding the placement of `oe_lfence()`: One option is to place it in `sgxlkl_ensure_outside()` which would ensure its use for all outside memory checks.  The other option is to place `oe_lfence()` at select locations that aggregate multiple mem checks. The former is preferred for completeness and simplicity, while the latter may improve performance slightly.
For recent discussion of the need for manually inserted lfences, please see:
https://github.com/openenclave/openenclave/issues/3401
and the discussion:
https://github.com/openenclave/openenclave/pull/3418

src/enclave/enclave_oe.c
```diff
@@ -311,38 +285,87 @@ static void _read_eeid_config()
...
     const sgxlkl_enclave_config_t* cfg = sgxlkl_enclave_state.config;
 
-    /* Deep copy where necessary */
+    /* The host shared memory struct has been copied by OE (but not its
+     * contents). */
+    sgxlkl_ensure_inside(host, sizeof(sgxlkl_shared_memory_t));
 
-    sgxlkl_shared_memory_t* enc = &sgxlkl_enclave_state.shared_memory;
-    memset(enc, 0, sizeof(sgxlkl_shared_memory_t));
+    /* Temporary, volatile copy to make sure checks aren't reordered */
```
bodzhang, line 307: `host` points to a `sgxlkl_shared_memory_t` structure copied into the enclave memory by OE. It's safe to access the structure inside the enclave directly, without worring about TOCTOU. Once all a check passes, the field from the `sgxlkl_shared_memory_t` structure pointed by `host` can be copied to sgxlkl_enclave_state.shared_memory directly, without using the volatile copy. If a shadow copy of a buffer pointed to by a pointer inside the `sgxlkl_shared_memory_t` structure is used, the correspnding pointer in sgxlkl_enclave_state.shared_memory can be handled differently.

wintersteiger: OE only copies the outermost layer, it doesn't deep-copy. If something goes wrong here, we may end up with a partially initialized data structure, which may or may not be a security problem. I prefer to be conservative here and keep the copy. Since this runs once, it's also not a performance concern.

src/enclave/enclave_oe.c
```diff
@@ -351,8 +374,12 @@ static void _copy_shared_memory(const sgxlkl_shared_memory_t* host)
         for (size_t i = 0; i < enc->num_virtio_blk_dev; i++)
         {
             enc->virtio_blk_dev_mem[i] = host->virtio_blk_dev_mem[i];
+            sgxlkl_ensure_outside(
+                enc->virtio_blk_dev_mem[i], sizeof(struct virtio_dev));
+
             const char* name = host->virtio_blk_dev_names[i];
             size_t name_len = oe_strlen(name) + 1;
```
bodzhang: If the name string is manipulated by the malicios host (for exmaple, no NULL terminator), `strlen()` might read memory from both untrusted memory and enclave memory, exposing side channel info of enclave memory (whether a byte is NULL or not). One possible solution is to require the host side provide the string length as input.

wintersteiger: Yes, that's on my to-do list.

src/enclave/enclave_oe.c
```diff
@@ -361,42 +388,112 @@ static void _copy_shared_memory(const sgxlkl_shared_memory_t* host)
...
+    if (cfg->num_host_import_env && cfg->host_import_env && henv && henv_len)
     {
-        size_t henvc = 0;
-        while (host->env[henvc++] != 0)
-            ;
+        sgxlkl_ensure_outside(henv, sizeof(char) * henv_len);
+
+        size_t num_vars = 0;
+        for (size_t i = 0; i < henv_len; i++)
+            if (henv[i] == '\0')
```
bodzhang: What's the definition of `host->env`?

src/include/shared/vio_event_channel.h
```diff
@@ -18,7 +18,6 @@ typedef struct enc_dev_config
 {
     uint8_t dev_id;
     enc_evt_channel_t* enc_evt_chn;
```
bodzhang, line 20: Pointer chain here is unnecessarily complex. uint64_t enclave_evt_channel, uint64_t host_evt_channel, uint32_t qidx all shoud be in the share memory. Defining enc_dec_config as below should simplify the memory check in enclave.
`````
{
    uint8_t dev_id;
    evt_t *enc_evt_channel;
    evt_t *host_evt_channel;
    uint32_t *qidx_p
}
```

src/enclave/enclave_oe.c
wintersteiger: I'm not sure I understand your question. `host->henv` is a long string of size `host->env_length`, which contains `\0` characters between environment variable definitions.

bodzhang: Got it.

The code here counts the number of \0 terminators, but later code might encounter TOCTOU situation, where the malicious host code modifies the buffer. Safer to copy the whole buffer into the enclave, if the paser logic later is complex.

wintersteiger: Sure, that's easy enough to add.

src/enclave/enclave_oe.c
```diff
@@ -361,42 +388,112 @@ static void _copy_shared_memory(const sgxlkl_shared_memory_t* host)
...
+        size_t begin = 0;
+        sgxlkl_enclave_state.num_env_imported = 0;
+        for (size_t i = 0; i < henv_len; i++)
+        {
+            if (henv[i] == '\0')
+            {
+                const char* henv_i = henv + begin;
+                size_t henv_i_len = i - begin;
+
+                for (size_t j = 0; j < cfg->num_host_import_env; j++)
```
bodzhang, line 445: Hard to understand the code logic here. Adding comments will help.

src/enclave/enclave_oe.c
wintersteiger: I added oe_lfence()s to the SGX-LKL memory boundary checks for good measure. We can remove or adjust them once https://github.com/openenclave/openenclave/pull/3418 comes to fruition.

src/include/shared/vio_event_channel.h
wintersteiger: This brings a lot of complications in other parts of the code (e.g. during setup on the host side and `host_evt_channel`  and `enclave_evt_channel` would have to be a `evt_t**`) and I'm not sure those parts are even correct. @prp @davidchisnall: is it intended and correct that `host_dev_event_channel_init` ([here](https://github.com/lsds/sgx-lkl/blob/fcd488160888ef111508cbf1865c3d59dcbdd83f/src/main-oe/sgxlkl_evt_chn_cfg.c#L44)) allocates `evt_channel_num` many `enc_evt_channel_t`s for each event channel with number `evt_channel_num`? Are these essentially (n^2)/2 communication links?

wintersteiger:
I think this one's ready to be merged. @SeanTAllen is working on the virtio refactoring, but that will be a separate PR.

PR #723, opened 2020-08-05
chmod rootfs files when mounting.

SeanTAllen:
This fixes #236.

The problem was that we were setting "/" to be 700 for perms which doesn't allow any user other than root to access "/tmp" and other directories.

After some review and conversation with @prp, I believe that in general the existing code in lkl_prepare_rootfs was there as  "just in case" sanity check. It creates directories if they are missing and provides some permissions (which differ from normal Linux installs). However, those directories almost always existed.

This change will now chmod those directories (including "/") if they exit. Additionally it updates the perms to be set to be 755 to match a standard Linux environment.

PR #744, opened 2020-08-07
Undo LTP patches related to #224

### LINKED ISSUE #224: [Tests] nobody user does not have permissions to open /etc/passwd
Identified issue while fixing setfsuid03 testcase. 
In setfsuid03 test case at line no 55 : https://github.com/linux-test-project/ltp/blob/master/testcases/kernel/syscalls/setfsuid/setfsuid03.c#L55
getpwuid(uid) function is being called with nobody user.

From here it goes in the below mentioned pattern and reaches __getpw_a function at src/passwd/getpw_a.c:24 and tries to open /etc/passwd with **nobody** user. Here it fails with Permission denied error (EACCESS: 13).

```
**Code flow:**
getpwuid (uid=1) at src/passwd/getpwent.c:28
28              __getpw_a(0, uid, &pw, &line, &size, &res);
(gdb) s
__getpw_a (name=0x0, uid=1, pw=0x7fff00ad12c0 <pw>, buf=0x7fff00b1aa50 <line>, size=0x7fff00b1aa58 <size>,
    res=0x7fff40b76d98) at src/passwd/getpw_a.c:24

int __getpw_a(const char *name, uid_t uid, struct passwd *pw, char **buf, size_t *size, struct passwd **res)
21      {
22              FILE *f;
23              int cs;
24              int rv = 0;
25
26              *res = 0;
27
28              pthread_setcancelstate(PTHREAD_CANCEL_DISABLE, &cs);
		f = fopen("/etc/passwd", "rbe");
31              if (!f) {
32                      rv = errno;
```

SeanTAllen:
Issue #224 was fixed by #723.

Closes #224


wintersteiger (approved):
LGTM! For this type of change, it would be really nice to see one revert-commit that makes it obvious that exactly the same changes have been undone. Not sure that's easy to do though.

hukoyu (approved):
LGTM

PR #749, opened 2020-08-07
Generate TLS credentials for user space

jxyang:
Generate TLS certificate and private key and expose them to user space through tmpfs. We need this for establishing attested TLS channel with external apps. Currently a few scenario tests also depend on this feature.

### DIFF
--- a/src/enclave/enclave_cert.c
+++ b/src/enclave/enclave_cert.c
@@ -1,0 +1,237 @@
+#include "enclave/enclave_cert.h"
+#include <mbedtls/x509.h>
+#include <mbedtls/x509_crt.h>
+#include <openenclave/enclave.h>
+#include "openenclave/corelibc/oemalloc.h"
+#include "openenclave/corelibc/oestring.h"
+
+static oe_result_t _generate_key_pair(
+    uint8_t** public_key_out,
+    size_t* public_key_size_out,
+    uint8_t** private_key_out,
+    size_t* private_key_size_out)
+{
+    oe_result_t result = OE_FAILURE;
+    oe_result_t ret;
+    oe_asymmetric_key_params_t params;
+    char user_data[] = "__USER_DATA__";
+    size_t user_data_size = sizeof(user_data) - 1;
+    uint8_t* public_key = NULL;
+    size_t public_key_size = 0;
+    uint8_t* private_key = NULL;
+    size_t private_key_size = 0;
+
+    *public_key_out = NULL;
+    *public_key_size_out = 0;
+    *private_key_out = NULL;
+    *private_key_size_out = 0;
+
+    memset(&params, 0, sizeof(params));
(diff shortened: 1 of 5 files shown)


src/enclave/enclave_cert.c
```diff
@@ -0,0 +1,247 @@
...
+    *public_key_out = public_key;
+    *public_key_size_out = public_key_size;
+    public_key = NULL;
+
+    result = OE_OK;
+
+done:
+
+    if (private_key)
+        oe_free(private_key);
```
paulcallen, line 76: should use oe_free_key instead

src/enclave/enclave_cert.c
```diff
@@ -0,0 +1,247 @@
...
+
+    if (public_key)
+        oe_free(public_key);
```
paulcallen, line 79: same

paulcallen (changes requested):
Minor change requested, other than that looks good to me. Christoph needs to review too.

jxyang:
This PR failed to generate certificate due to https://github.com/openenclave/openenclave/issues/3378. We need to fix OE before merging this PR.

src/enclave/enclave_init.c
```diff
@@ -107,6 +117,26 @@ static int startmain(void* args)
...
 
+    /* Generate TLS certificate and private key and save them to files.
+     * Once we have kernel/user separation, pass the cert and private
+     * key to user space through the stack, and move
+     * sgxlkl_write_tls_credentials to user space.
+     */
+
+    if (!sgxlkl_in_sw_debug_mode())
+    {
+        enclave_generate_tls_credentials(
```
wintersteiger, line 128: I think there's a layering problem here; we cant call `oe_*` at this point in time anymore. If I'm not mistaken, the consensus was that we would generate this kind of data earlier and then pass it to the application via auxv. This would also mean that applications could pick it up directly from memory instead of a file. The current implementation would also fail on a read-only root file system. (CC @davidchisnall)

src/enclave/enclave_cert_user.c
```diff
@@ -0,0 +1,42 @@
+#include "enclave/enclave_cert_user.h"
+#include <stdio.h>
+#include <stdlib.h>
+
+// This function and its caller should be moved to
```
wintersteiger, line 5: There's no need to wait, we can do this now, can't we?

wintersteiger (changes requested):
Technically fine, but likely a layering problem.

src/enclave/enclave_init.c
davidchisnall: Yes, this routine is a part of the nominal userspace and will be refactored soon to make that distinction more explicit.  We may not call OE APIs from here.

More generally, given that we've had so many meetings discussing the design and the mechanism for passing this via the ELF aux vector, I'd like to understand why a different approach is being pursued here.

Issue #748, opened 2020-08-07
pthread_join test hangs occasionally


prp:
A CI run of `basic-pthread_join-(debug)-(run-hw)-(8-ethreads)` led to the following hanging behaviour in the middle of the test:
```
Execution timeout: 60
Will run tests for run-hw
/opt/sgx-lkl-debug/bin/sgx-lkl-disk create --size=5M --docker=./Dockerfile sgx-lkl-rootfs.img
Using detected host DNS server in /etc/resolv.conf: 168.63.129.16
Creating sgx-lkl-rootfs.img from Dockerfile ./Dockerfile...
Building Docker image...
Creating disk image file from Docker container...
Succesfully created sgx-lkl-rootfs.img.
SGXLKL_VERBOSE=1 SGXLKL_KERNEL_VERBOSE=1 /opt/sgx-lkl-debug/bin/sgx-lkl-run-oe --hw-debug sgx-lkl-rootfs.img pthread_join-test
[   SGX-LKL  ] nproc=4 ETHREADS=8 CMDLINE="mem=32M"
[   SGX-LKL  ] HW TLS support: econf->fsgsbase=1
[   SGX-LKL  ] Registering disk 0 (path='sgx-lkl-rootfs.img', mnt='/', [RW])
[   SGX-LKL  ] No tap device specified, networking will not be available.
[   SGX-LKL  ] get_signed_libsgxlkl_path... result=/opt/sgx-lkl-debug/bin/../lib/libsgxlkl.so.signed
[   SGX-LKL  ] oe_create_enclave...
[...]

Issue #171, opened 2020-05-04
Need to support debugging/trouble-shooting in NON-RELEASE/RELEASE builds


prp:
We need to decide how to support debugging (using gdb) and trouble-shooting (stdout/stderr etc.) of applications running using the NON-RELEASE/RELEASE builds of SGX-LKL (see https://github.com/lsds/sgx-lkl/issues/144).

One aspect is debugging/trouble-shooting of SGX-LKL itself; the other is providing debugging/trouble-shooting support for the application run by SGX-LKL.

App (in-enclave) trouble-shooting:
- Secure stdout/stderr capture: #170
- Print stack traces on in-enclave kernel exceptions: #321
- TODO: Print stack traces on in-enclave userspace exceptions

App (in-enclave) debugging:
- Only possible for debug/non-release builds of SGX-LKL (debug enclaves)
  - Currently broken for non-release builds: #573
- TODO: make sure core dumps are readable for host and enclave part
- TODO: documentation on using `sgx-lkl-gdb` for different app types
  - decide if this should work for debug CCs as well or just non-CC usage

SGX-LKL host-side trouble-shooting:
- Print stack traces on host-side segfaults etc.: #205

SGX-LKL host-side debugging:
- Regular gdb is enough, `sgx-lkl-gdb` useful to debug both host and enclave
  - decide if this should work for debug CCs as well or just non-CC usage
- TODO: make sure core dumps for the host-side are readable for RELEASE builds
- ensure all SGX-LKL builds (also RELEASE) have debug symbols for the host-side: https://github.com/lsds/sgx-lkl/issues/205

SGX-LKL in-enclave trouble-shooting:
- Print stack traces on in-enclave segfaults etc.: #321
- enable verbose SGX-LKL in-enclave logging (also RELEASE) via app config options, used in combination with stdout/stderr capturing: https://github.com/lsds/sgx-lkl/issues/319
- Extend & improve SGX-LKL logging: #215

SGX-LKL in-enclave debugging:
- Similar to app debugging
- documentation on using `sgx-lkl-gdb` to debug SGX-LKL code #330 -- DONE

cc: @paulcallen

Issue #144, opened 2020-05-03
Add support for RELEASE builds of SGX-LKL


prp:
Build types for SGX-LKL
=================

We want to support the following three mutually-exclusive build types for SGX-LKL:

1. `make DEBUG=true` **# DEBUG build**

This creates a debug build of SGX-LKL with full debug and tracing functionality, debug symbols and `-O0`.

This version should be used for development of SGX-LKL. It is not secure because it permits execution without attestation and secrets (e.g., root encryption keys) can be passed directly to the launcher via environment variables.

2. `make` **# NON-RELEASE build**

This builds a version of SGX-LKL with some (low-overhead) debug and tracing functionality, no debug symbols and `-O3`. This is the default build when no make target is specified.

This version should be used for performance benchmarking and testing. It is not secure because it permits execution without attestation and secrets (e.g., root encryption keys) can be passed directly to the launcher via environment variables.

3. `make RELEASE=true` **# RELEASE build**

This builds a release version of SGX-LKL without debug functionality, no debug symbols and `-O3`. 

This version should be used for production deployments because it is the only version that is secure. It excludes any code from the enclave that compromises security, i.e., it: 
- removes support for execution of SGX-LKL in software mode;
- requires a valid app_config to be specified, as all app config environment variables are ignored;
- only permits execution of a non-debug OE hardware enclave;
- enforces attestation, so it becomes impossible to run an enclave application locally (without access to an attestation and key release service) for testing;
- deactivates output/input from/to the LKL console, which would otherwise constitute data leakage from the enclave.

Packaging of build types
=================

The different builds of SGX-LKL could be installed in difference directory prefixes on the file system, e.g.:
* `/opt/sgx-lkl-debug/`
* `/opt/sgx-lkl-non-release/`
* `/opt/sgx-lkl-release/`

We could have a symlink `sgx-lkl` that points to one of them (e.g. the RELEASE build to have security by default?). When launching `sgx-lkl-run-oe`, it outputs clearly which version it is.

Dependencies
===========

- The implementation of a RELEASE build depends on attestation support.
- The changes to the build system should probably happen after the overall build system has been updated.

cc: @letmaik @wintersteiger @davidchisnall @paulcallen @jxyang

letmaik:
I don't think it's a good idea to have different launcher names. This will complicate any kind of automation. Why not simply use a different installation prefix and keep everything inside the same?

prp:
[...]

PR #751, opened 2020-08-08
Dump stack traces from all lthreads on SIGTERM

prp:
This PR makes the lthread scheduler display stack traces for all lthreads after the launcher has received a SIGTERM. This means that CI tests that time out will now display stack traces because the timeout command sends a SIGTERM.

It works by installing a SIGTERM signal handler in the launcher and using a new ecall to output the stack traces. 

This is only supported in DEBUG builds.

In addition the PR also:
- fixes a merge conflict related to the outputting of mount tables;
- fixes the SGXLKL_TRACE_SIGNAL environment variable.

This is to help get more information when issue https://github.com/lsds/sgx-lkl/issues/748 occurs in the CI, as I find it hard to reproduce it locally.

### DIFF
--- a/src/lkl/setup.c
+++ b/src/lkl/setup.c
@@ -1220,6 +1220,10 @@ static void* lkl_termination_thread(void* args)
             lkl_strerror(ret));
     }
 
+#ifdef DEBUG
+    display_mount_table();
+#endif
+
     // Unmount mounts
     long res;
     for (int i = sgxlkl_enclave_state.num_disk_state - 1; i > 0; --i)
@@ -1239,10 +1243,6 @@ static void* lkl_termination_thread(void* args)
                 "Could not unmount disk %d, %s\n", i, lkl_strerror(res));
         }
 
-#ifdef DEBUG
-    display_mount_table();
-#endif
-
         if (!cfg->root.readonly)
         {
             // Not really necessary for mounts in /mnt since /mnt is
@@ -1261,10 +1261,6 @@ static void* lkl_termination_thread(void* args)
         }
     }
 
-#ifdef DEBUG
-    display_mount_table();
-#endif
-
     /* Unmount root.
      * We are calling umount with the MNT_DETACH flag for the root
      * file system, otherwise the call fails to unmount the file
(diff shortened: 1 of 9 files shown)


src/lkl/setup.c
```diff
@@ -1243,6 +1239,10 @@ static void* lkl_termination_thread(void* args)
                 "Could not unmount disk %d, %s\n", i, lkl_strerror(res));
         }
 
+#ifdef DEBUG
+    display_mount_table();
```
davidchisnall: Why is this moved?  Don't we want the mount table from before we unmount everything?

src/main-oe/sgxlkl_run_oe.c
```diff
@@ -331,14 +335,20 @@ static void help_enclave_config()
...
-/* Control the event channel notification between host & guest */
-
-/* Signal handler to resume paused evt chn to resume guest */
+#ifdef DEBUG
+// Signal handler for thread stack dumps and virtio event channel tests
 static void sgxlkl_loader_signal_handler(int signo)
 {
     switch (signo)
     {
+        case SIGTERM:
```
davidchisnall, line 344: Can we do this for SIGUSR1 as well?  It would be convenient to have a way of doing this form outside without stopping the program.

src/sgxlkl.edl
```diff
@@ -31,6 +31,11 @@ enclave {
         // Enclave call for initializing ethreads to enter enclave
         public void sgxlkl_ethread_init(void);
 
+        // Enclave call to dump stack traces for all lthreads (DEBUG only)
+        // TODO: This should only be included for a DEBUG build, but EDL doesn't seem to
```
davidchisnall, line 35: It does have include.  We could have a separate top-level EDL file for debug and non-debug builds and include the non-debug one in the debug build.

davidchisnall (approved):

src/lkl/setup.c
prp: Before the wrong merge, the idea was to output the mount table twice, before the unmount and after, to check that things have been correctly unmounted. The broken merge meant that we are outputting the mount table twice before the unmount, which  outputs the same information twice.

As an aside, I am not convinced that the unmouting always reliably flushes the file system state back to disk. I may open an issue on this.

src/main-oe/sgxlkl_run_oe.c
prp: Ok, that's an easy change to make (especially since the stack traces always come from the scheduler state and do not interrupt thread execution).

src/sgxlkl.edl
prp: Ah, ok. I suggest that this should be a separate PR though, as it's related to #144 and #171.

davidchisnall: No objections.

src/lkl/setup.c
prp: Ok, I didn't move it to the correct location, but I fixed this now.

I've also opened this issue: https://github.com/lsds/sgx-lkl/issues/754

Issue #370, opened 2020-06-03
lkl_virtio_netdev_add ignores errors


letmaik:
- Return code of `lkl_virtio_dev_setup()` is not checked.
- Return code of `dev_register()` is checked but only leads to `sgxlkl_info(..)` on error.

davidchisnall:
@prp, I thought you fixed this one?

prp:
@davidchisnall no, cleaning up the error handling throughout the virtio code is still outstanding. I only did the minimum amount of work to fix https://github.com/lsds/sgx-lkl/issues/491.

PR #753, opened 2020-08-10
Check lkl_virtio_dev_setup() return value

Francis-Liu:
Part of fix to #370

### DIFF
--- a/src/lkl/setup.c
+++ b/src/lkl/setup.c
@@ -861,7 +861,8 @@ void lkl_mount_disks(
     if (!root)
         sgxlkl_fail("No root disk provided. Aborting...\n");
 
-    lkl_add_disks(root, mounts, num_mounts);
+    if (lkl_add_disks(root, mounts, num_mounts) != 0)
+        sgxlkl_fail("Add root disk failed. Aborting...\n");
 
     lkl_mount_root_disk(root, 0);
 
--- a/src/lkl/virtio_blkdev.c
+++ b/src/lkl/virtio_blkdev.c
@@ -31,7 +31,7 @@ static void lkl_deliver_irq(uint8_t dev_id)
 /*
  * Function to perform virtio device setup
  */
-void lkl_add_disks(
+int lkl_add_disks(
     const sgxlkl_enclave_root_config_t* root,
     const sgxlkl_enclave_mount_config_t* mounts,
     size_t num_mounts)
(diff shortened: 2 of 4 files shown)


src/lkl/setup.c
```diff
@@ -861,7 +861,8 @@ void lkl_mount_disks(
     if (!root)
         sgxlkl_fail("No root disk provided. Aborting...\n");
 
-    lkl_add_disks(root, mounts, num_mounts);
+    if (lkl_add_disks(root, mounts, num_mounts) != 0)
```
davidchisnall, line 864: `!= 0` is redundant here.  I normally prefer explicit checks here, but we're using this return value as a boolean and so C boolean semantics (0 is false, non-zero is true) are fine.  This test makes me wonder what the other return values are.

src/lkl/virtio_blkdev.c
```diff
@@ -42,14 +42,17 @@ void lkl_add_disks(
         sgxlkl_enclave_state.shared_memory.virtio_blk_dev_mem
             [sgxlkl_enclave_state.disk_state[0].host_disk_index];
     int mmio_size = VIRTIO_MMIO_CONFIG + root_dev->config_len;
-    lkl_virtio_dev_setup(root_dev, mmio_size, lkl_deliver_irq);
+    if (lkl_virtio_dev_setup(root_dev, mmio_size, lkl_deliver_irq) != 0)
```
davidchisnall, line 45: As above.

src/lkl/virtio_blkdev.c
```diff
@@ -42,14 +42,17 @@ void lkl_add_disks(
...
+        return -1;
 
     for (size_t i = 0; i < num_mounts; ++i)
     {
         struct virtio_dev* dev =
             sgxlkl_enclave_state.shared_memory.virtio_blk_dev_mem
                 [sgxlkl_enclave_state.disk_state[i + 1].host_disk_index];
         int mmio_size = VIRTIO_MMIO_CONFIG + dev->config_len;
-        lkl_virtio_dev_setup(dev, mmio_size, lkl_deliver_irq);
+        if (lkl_virtio_dev_setup(dev, mmio_size, lkl_deliver_irq) != 0)
```
davidchisnall, line 54: And here.

davidchisnall (approved):
Minor comments, happy for this to be merged once those are addressed.

paulcallen (approved):

src/lkl/setup.c
Francis-Liu: Sorry @davidchisnall I replied your comment using VScode GitHub plugin, but apparently the comment didn't go through to github.

I saw in sgxlkl return code 0 is used as SUCCESS. And -1, 1, or -SOME_ERROR_CODE are used on failure. Which gave me an impression that it follows Unix program's convention (return 0 on success). I think if we use non-zero as success, then we need to check against specific error code. What do you think?

This PR was merged, if you insist we use non-zero as success, then I can make the change in my next PR.

PR #755, opened 2020-08-10
Ignore fd when doing MAP_ANONYMOUS for mmap

SeanTAllen:
It's considered best practice for portable applications to provide -1
as the fd to mmap when doing MAP_ANONYMOUS. However, for Linux compatibility,
we want to ignore the fd entirely.Ignore fd when doing MAP_ANONYMOUS for mmap


SeanTAllen:
This came up from https://github.com/lsds/ltp/pull/58#issuecomment-671242240

davidchisnall (approved):

bodzhang (approved):

PR #757, opened 2020-08-10
Handle errors in console_enqueue

### LINKED ISSUE #374: console_enqueue() has incorrect error handling
https://github.com/lsds/sgx-lkl/blob/6016fbc9eabda2799ac914c8278e37d2e2f385b2/src/host_interface/virtio_console.c#L172-L199

- No status code is returned.
- `virtio_req_complete()` is called with status code that may be a negative error code, but the function accepts unsigned integers only which is the length of data processed. Should `virtio_req_complete()` be called with length 0 in case of errors? Or should a `readv` failure always lead to a fatal exit?

SeanTAllen:
Prior to this commit, console_enqueue would ignore errors and not
report them via the return value. After this commit, console_enqueue
handles errors the same as sgx-lkl's other virtio devices.

Closes #374

### DIFF
--- a/src/host_interface/virtio_console.c
+++ b/src/host_interface/virtio_console.c
@@ -175,7 +175,7 @@ static int console_enqueue(
     int q,
     struct virtio_req* req)
 {
-    int ret = 0;
+    ssize_t ret = 0;
     struct virtio_console_dev* vcd = get_console_dev_instance();
     struct iovec* iov = req->buf;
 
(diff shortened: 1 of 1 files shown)


SeanTAllen:
See #374 for fuller discussion on how to handle this and more background.

https://github.com/lsds/sgx-lkl/issues/374

davidchisnall (approved):

src/host_interface/virtio_console.c
```diff
@@ -185,20 +185,19 @@ static int console_enqueue(
...
+        {
+            ret = writev(vcd->out_console_fd, iov, req->buf_count);
+        } while (ret == -1 && errno == EINTR);
     }
 
-    virtio_req_complete(req, ret);
-    // TODO https://github.com/lsds/sgx-lkl/issues/374
-    return 0;
+    ssize_t bytes = ret <= 0 ? 0 : ret;
+    virtio_req_complete(req, bytes);
```
prp, line 198: We are calling this (and returning success), even if `readv` or `writev` didn't process all of the bytes. Is that correct?

prp (commented):
Thanks. I wonder if we handle the case of partial reads/writes correctly.

src/host_interface/virtio_console.c
SeanTAllen: Correct as in, is that what we are doing? Yes
Correct as in, the right thing to do? I'm not sure. From my fairly limited understanding from reading the descriptions in headers, I think it is fine.

paulcallen (approved):

vtikoo (approved):

PR #758, opened 2020-08-10
Add lthread_yield_and_sleep() call and remove unused RB tree

prp:
This PR adds an `lthread_yield_and_sleep()` call to be used by the event channel mechanism and removes the unused sleeping RB tree.


davidchisnall:
I don’t think this is correct, now that I’ve reread the logic. This thread sleeps here and is woken by the event channel being signalled, near the bottom of the file. With this change, the thread will keep polling.

prp:
> I don’t think this is correct, now that I’ve reread the logic. This thread sleeps here and is woken by the event channel being signalled, near the bottom of the file. With this change, the thread will keep polling.

@davidchisnall you're right. Actually looking at the code, are we doing anything with the sleeping RB tree? Can this go?

davidchisnall (approved):

vtikoo (approved):
Looks good to me.

Also I don't think the sleeping rbtree was being used for anything.

vtikoo:
Can there be a race condition between the yield callback setting the sleeping bit, and lthread_wakeup to signal the thread? Causing lost notifications or maybe even running `lthread_resume` in two different ethreads for the same lthread.

Wondering if its better to acquire `_lthread_lock` at both places - while yielding in [vio_wait_for_host_event()](https://github.com/lsds/sgx-lkl/blob/oe_port/src/enclave/enclave_event_channel.c#L55)(unlock in the yield callback), and `lthread_wakeup`?

prp:
> Can there be a race condition between the yield callback setting the sleeping bit, and lthread_wakeup to signal the thread? Causing lost notifications or maybe even running `lthread_resume` in two different ethreads for the same lthread.
> 
> Wondering if its better to acquire `_lthread_lock` at both places - while yielding in [vio_wait_for_host_event()](https://github.com/lsds/sgx-lkl/blob/oe_port/src/enclave/enclave_event_channel.c#L55)(unlock in the yield callback), and `lthread_wakeup`?

Hmm, that's a good point. The state is atomic, so I think that the issue is rather that a concurrent wakeup may be missed (if we only deliver it once)? @davidchisnall @SeanTAllen what do you think?

SeanTAllen:
I don't think I understand the race condition. I'm not particularly familiar with the yield code so that would be why. I defer to you on this one.

PR #759, opened 2020-08-10
Free enclave state when last ethread exits.

vtikoo:
### Summary
`sgxlkl_enclave_state state` is being cleaned up during `lkl_termination_thread`. This is too early for the lthread scheduler which relies on `state->config.mode` for context switches. According to @prp this is a cause of other failures as well.

This PR moves `sgxlkl_free_enclave_state` to run later, when the last ethread exits.

### DIFF
--- a/src/enclave/enclave_oe.c
+++ b/src/enclave/enclave_oe.c
@@ -1,3 +1,4 @@
+#include <stdatomic.h>
 #include "pthread_impl.h"
 
 #include <openenclave/bits/eeid.h>
@@ -12,6 +13,7 @@
 #include "shared/env.h"
 #include "shared/timer_dev.h"
 
+int ethreads_minus_1 = -1;
 sgxlkl_enclave_state_t sgxlkl_enclave_state = {0};
 
 bool sgxlkl_in_sw_debug_mode()
@@ -216,6 +218,14 @@ static void _sgxlkl_enclave_show_attribute(const void* sgxlkl_enclave_base)
 }
 #endif
 
+static inline void dec_ethread_counter() {
+    if (a_fetch_add(&ethreads_minus_1, -1) == 0) {
+        // last ethread to exit
+        ethreads_minus_1 = 0;
+        sgxlkl_free_enclave_state();
+    }
+}
+
 void sgxlkl_ethread_init(void)
 {
     void* tls_page;
@@ -234,9 +244,11 @@ void sgxlkl_ethread_init(void)
     }
 
     /* Initialization completed, now run the scheduler */
+    a_inc(&ethreads_minus_1);
     __init_tls();
     _lthread_sched_init(sgxlkl_enclave_state.config->stacksize);
     lthread_run();
+    dec_ethread_counter();
 
     return;
 }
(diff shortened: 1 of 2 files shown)


src/enclave/enclave_oe.c
```diff
@@ -216,6 +218,14 @@ static void _sgxlkl_enclave_show_attribute(const void* sgxlkl_enclave_base)
 }
 #endif
 
+static inline void dec_ethread_counter() {
+    if (a_fetch_add(&ethreads_minus_1, -1) == 0) {
```
davidchisnall, line 222: In new code, please use C11 atomic instead of the musl functions.  Since we want sequential consistency, just declare the variable with the _Atomic qualifier as _Atomic(int).

src/enclave/enclave_oe.c
```diff
@@ -12,6 +13,7 @@
 #include "shared/env.h"
 #include "shared/timer_dev.h"
 
+int ethreads_minus_1 = -1;
```
davidchisnall, line 16: Please add a comment explaining why this need the minus one bit.

src/enclave/enclave_oe.c
```diff
@@ -379,7 +391,10 @@ int sgxlkl_enclave_init(const sgxlkl_shared_memory_t* shared_memory)
     SGXLKL_VERBOSE("calling _dlstart_c()\n");
     _dlstart_c((size_t)sgxlkl_enclave_base);
```
davidchisnall, line 393: It feels as if this should be earlier. Since this is the first ethread, why not skip it entirely and initialise the global assuming the initial ethread exists (if it doesn’t, nothing happens).

src/enclave/enclave_oe.c
```diff
@@ -234,9 +244,11 @@ void sgxlkl_ethread_init(void)
     }
 
     /* Initialization completed, now run the scheduler */
+    a_inc(&ethreads_minus_1);
```
davidchisnall, line 247: I think there’s a race here. What happens if we call this just after another ethread has exited and deleted the enclave state? This won’t happen in normal operation, but an attacker can trigger it.

I think we want to add a second atomic counter on the number of exited threads and increment it before decrementing this counter on thread exit. At this point, we want to check if the exited threads counter is >0 and, if so, exit.

davidchisnall (changes requested):

wintersteiger (approved):
That's right, we definitely need to run this later. I think there is in deed a data race where @davidchisnall suspects one and his suggested fix should work. 

On the general subject of cleanup: since we don't clean up the thread state, we might as well not clean up the config and drop `sgxlkl_free_enclave_state` entirely.

davidchisnall:
> On the general subject of cleanup: since we don't clean up the thread state, we might as well not clean up the config and drop `sgxlkl_free_enclave_state` entirely.

I agree.  There is no real benefit in trying to do this: the enclave will be destroyed on exit and all state will be cleaned up.  The main thing I think we need is a guard on entry to the ecalls to make sure the host can't try to launch new ethreads when we are shutting down.  We probably also need a guard here to make sure that we only ever get one initial thread:

https://github.com/lsds/sgx-lkl/blob/c4885aa7127099c5f95d9a6c86c655b0d93fadb2/src/enclave/enclave_oe.c#L339

vtikoo:
Closing this PR, as there is consensus on not freeing enclave state. I 'll create a new one with the `sgxlkl_free_enclave_state()` removed and additional asserts.

PR #762, opened 2020-08-11
Fix some atomicity in VirtIO control status accesses.

davidchisnall:
None of these updates were atomic, which may have caused problems.
Most notably, there was a potential race in the interrupt-status
flag: setting the flag was guarded by a ticket lock (which was still
held when the thread slept, which should never happen because ticket
locks are not sleepable locks) but clearing it was not guarded.


prp (approved):
LGTM.

paulcallen (approved):

PR #763, opened 2020-08-11
Use correct file descriptor number for STDERR.

### LINKED ISSUE #221: Mismatching in-enclave stdout/stderr file numbers
https://github.com/lsds/sgx-lkl/blob/ce88f974da24683bdf2903bf3c165724954a3dcc/src/enclave/enclave_util.c#L8-L17

However, in OE fd 1 is stdout:

https://github.com/openenclave/openenclave/blob/d5a7a8393a50e0613791187e4bc205c6b213ad0b/include/openenclave/internal/syscall/unistd.h#L16-L18
```
#define OE_STDIN_FILENO 0
#define OE_STDOUT_FILENO 1
#define OE_STDERR_FILENO 2
```

I noticed this because I wanted to change the non-error output to stdout in enclave_util.c:
https://github.com/lsds/sgx-lkl/blob/ce88f974da24683bdf2903bf3c165724954a3dcc/src/enclave/enclave_util.c#L35-L40

SeanTAllen:
Closes #221

### DIFF
--- a/src/enclave/enclave_util.c
+++ b/src/enclave/enclave_util.c
@@ -9,7 +9,7 @@
 #include "openenclave/internal/backtrace.h"
 #endif
 
-#define OE_STDERR_FILENO 1
+#define OE_STDERR_FILENO 2
 
 void sgxlkl_fail(const char* msg, ...)
 {


SeanTAllen:
There are a couple alternatives here.

Given that we were already sending to STDOUT, rename to indicate more correctly that all messages are going to STDOUT and not "suddenly" start sending all the output to STDERR instead of STDOUT.

Combine make sgxlkl_fail go to STDERR and the rest go to STDOUT.

I did the simplest thing here to close the issue, but I'm not sure it's "the right thing".

Whatever we choose to do should be easy enough to do so I went ahead and opened this PR for one of the options.

davidchisnall (approved):
I think all informative messages going to stderr is the right thing to do.

prp (approved):
LGTM.

SeanTAllen:
@davidchisnall @prp can one of you approve?

prp (approved):
LGTM.

davidchisnall (approved):
This hasn't changed since I last approved it...

jxyang (approved):

PR #764, opened 2020-08-11
LTP: Pass ext4 block device to enclave

shaikshavali1:
This test case causing oom killer to be invoked and causing
kernel panic. This is because the test case invokes test framework
function "tst_acquire_device" which creates a 256MB of .img file
and binds with a loop device. This is not supported because the
default size of LKL memory is set to 32M. This is kept low due to
the limited size of EPC (Enclave page cache) and to avoid page
cache being swapped out.
Also, the test case invokes "tst_mkfs" framework function which
inturn invokes mkfs utility command with the help of system() syscall.
It is recommended to use root file system, but, this test case
needs a read-only filesystem to test one of the sub-test cases.
we don't have a read-only file system mounted in sgx-lkl.

To support the test cases which need a block device a formatted
ext4 filesystem image created and passed
to enclave as extended disk. The test case passes the details about this
new block disk image in the application configuration file. The test framework
will pick this image details in the environment variable and use it in test.
Test case changes link: lsds/ltp#49, lsds/ltp#59, lsds/ltp#65,
lsds/ltp#66.

### DIFF
--- a/tests/ltp/batch.mk
+++ b/tests/ltp/batch.mk
@@ -25,6 +25,10 @@ ESCALATE_CMD=sudo
 
 LTP_SOURCE_DIR=$(SGXLKL_ROOT)/ltp
 
+# file system image to be mount in ltp tests
+LTP_TEST_MNT_IMG="ltp_tst_mntfs.img"
+LTP_TEST_MNT_IMG_SIZE=256
+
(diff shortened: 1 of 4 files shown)


SeanTAllen (approved):

hukoyu (approved):
LGTM

Issue #766, opened 2020-08-11
LTP fchown05 failed with OE_OUT_OF_THREADS error


hukoyu:
This looks like a regression. It doesn't fail everytime but recording this failure.
This LTP test (fchown05) is not pacthed, original test case enabled from the beginning.

```
----------output-start-------------
[[  SGX-LKL ]] lkl_start_init(): calling lkl_mount_virtual()
[[  SGX-LKL ]] lkl_termination_thread(): enter
[[  SGX-LKL ]] lkl_termination_thread(): Performed LKL syscall to get host task allocated (pid=36)
[[  SGX-LKL ]] lkl_start_init(): calling init_random()
[[  SGX-LKL ]] init_random(): Adding entropy to entropy pool
[    0.192578] random: crng init done
[[  SGX-LKL ]] init_enclave_clock(): Setting enclave realtime clock
[[  SGX-LKL ]] wg0 has public key Fft6thRsLMfkUnWF+7cO3pVygjkFRiulFw4pF90C8TI=
[    0.194755] virtio_blk virtio1: [vda] 3072000 512-byte logical blocks (1.57 GB/1.46 GiB)
[    0.195312] virtio_blk virtio2: [vdb] 524288 512-byte logical blocks (268 MB/256 MiB)
[[  SGX-LKL ]] lkl_mount_disk(): lkl_mount_disk(dev="/dev/vda", mnt="/mnt/vda", ro=0)
[    0.196983] EXT4-fs (vda): mounted filesystem with ordered data mode. Opts: 
[[  SGX-LKL ]] lkl_mount_disk(): lkl_mount_disk(dev="/dev/vdb", mnt="/data", ro=0)
[    0.204713] EXT4-fs (vdb): mounted filesystem with ordered data mode. Opts: 
[[  SGX-LKL ]] lkl_mount_disks(): Set working directory: /
[[  SGX-LKL ]] libc_start_main_stage2(): Calling app main: /ltp/testcases/kernel/syscalls/fchown/fchown05
fchown05    1  [1;32mTPASS[0m  :  fchown() succeeds to Change Owner/Group ids of testfile
fchown05    2  [1;32mTPASS[0m  :  fchown() succeeds to Change Owner id only of testfile
fchown05    3  [1;32mTPASS[0m  :  fchown() succeeds to Change Owner id only of testfile
fchown05    4  [1;32mTPASS[0m  :  fchown() succeeds to Change Group id only of testfile
[...]

PR #770, opened 2020-08-11
Fix shutdown bugs

prp:
This PR fixes two shutdown bugs:
- it fixes an off-by-one error that prevented the unmounting of additional file systems (see https://github.com/lsds/sgx-lkl/issues/766);
- it prevents the premature deallocation of the enclave configuration in the shutdown sequence. Instead, it lets the memory be reclaimed when we terminate the process (see https://github.com/lsds/sgx-lkl/pull/759).

### DIFF
--- a/src/lkl/setup.c
+++ b/src/lkl/setup.c
@@ -1176,6 +1176,9 @@ static void* lkl_termination_thread(void* args)
         "Performed LKL syscall to get host task allocated (pid=%li)\n", pid);
     SGXLKL_ASSERT(pid);
 
+    lthread_set_funcname(lthread_self(), "sgx-lkl-terminate");
+    lthread_detach();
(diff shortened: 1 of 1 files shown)


davidchisnall (approved):

vtikoo (approved):

src/lkl/setup.c
```diff
@@ -1227,7 +1230,7 @@ static void* lkl_termination_thread(void* args)
 
     // Unmount mounts
     long res;
-    for (int i = sgxlkl_enclave_state.num_disk_state - 1; i > 0; --i)
+    for (int i = cfg->num_mounts - 1; i >= 0; i--)
```
SeanTAllen, line 1233: o man.

PR #775, opened 2020-08-13
Extend kill delay after timeout to 60 secs

prp:
This PR extends the kill delay after a timeout to 60 secs, so there is
enough time to output a thread dump.


KenGordon (approved):

SeanTAllen (approved):

hukoyu (approved):
LGTM

PR #785, opened 2020-08-13
Perform sync syscall to persist file system contents

### LINKED ISSUE #713: Not able to write to files in the root disk image
I modified helloworld sample so that in addition to reading from "helloworld.txt", it also writes to the file. Note the enclave_config.json file indicates this is a writable disk image:

{
  "args": [
    "app/helloworld"
  ],
  "root": {
    "readonly": false
  }
}
However, after I run the sample, and mount the disk image, file "/app/helloworld.txt" is not updated.

This seems to be blocking code coverage measurement as we can't generate the coverage data files on the disk image.

### LINKED ISSUE #754: File systems unmounted using MNT_DETACH may become inconsistent
We currently unmount the root file system using the `MNT_DETACH` option, otherwise we sometimes observe hanging behaviour (despite the timeout): https://github.com/lsds/sgx-lkl/blob/a1fb63f406557cb02123682efaa48d78018a3421/src/lkl/setup.c#L1276

We should check that this always cleanly unmounts the file systems and does not lead to data loss. It also not clear that the code correctly unmounts the entire file system hierarchy.

prp:
This commit adds a sync syscall to the shutdown sequence, which ensures
that the content of the buffer cache is flushed to disk. It also adds a
test that verifies that a file is persisted across SGX-LKL runs.

Fixes https://github.com/lsds/sgx-lkl/issues/713 and #754.

### DIFF
--- a/src/lkl/setup.c
+++ b/src/lkl/setup.c
@@ -1215,12 +1215,22 @@ static void* lkl_termination_thread(void* args)
     }
 
     // Switch back to root so we can unmount all filesystems
-    SGXLKL_VERBOSE("calling lkl_sys_chdir(/)\n");
+    SGXLKL_VERBOSE("calling lkl_sys_chdir(\"/\")\n");
     int ret = lkl_sys_chdir("/");
     if (ret != 0)
     {
         sgxlkl_warn(
             "lkl_sys_chdir(\"/\") failed: ret=%i error=\"%s\"\n",
+            ret,
+            lkl_strerror(ret));
+    }
+
+    SGXLKL_VERBOSE("calling lkl_sys_sync()\n");
+    ret = lkl_sys_sync();
+    if (ret != 0)
+    {
+        sgxlkl_warn(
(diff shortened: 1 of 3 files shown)


src/lkl/setup.c
```diff
@@ -1215,7 +1215,7 @@ static void* lkl_termination_thread(void* args)
     }
 
     // Switch back to root so we can unmount all filesystems
-    SGXLKL_VERBOSE("calling lkl_sys_chdir(/)\n");
+    SGXLKL_VERBOSE("calling lkl_sys_chdir(\"/\")\n");
```
davidchisnall, line 1218: Why this change?

davidchisnall (approved):

src/lkl/setup.c
prp: Consistency with other output.

jxyang (approved):

hukoyu (approved):
LGTM

Issue #740, opened 2020-08-06
SIGSEGV: running basic/clone looping with 2 ethreads and OE debug allocator


paulcallen:
I enabled OE debug allocator  (-DUSE_DEBUG_MALLOC=ON) for building OE
Running on 4-core ACC VM with 2 ETHREADS
Running test tests/basic/clone with command `sudo make DEBUG=1 run-hw-gdb-clone-loop`
Crash is here:
```
Thread 6 "ENCLAVE" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffef172700 (LWP 13152)]
0x00007fe0005b80ba in _lthread_free (lt=0x7fe000b15ef0) at sched/lthread.c:361
361             void (*f)(void*) = lt->cancelbuf->__f;
(gdb) bt
#0  0x00007fe0005b80ba in _lthread_free (lt=0x7fe000b15ef0) at sched/lthread.c:361
#1  0x00007fe0005b866c in _lthread_resume (lt=0x7fe000b15ef0) at sched/lthread.c:530
#2  0x00007fe0005b7d1b in lthread_run () at sched/lthread.c:248
#3  0x00007fe0005a7d11 in __libc_init_enclave (argc=1, argv=0x7fe000b04840) at enclave/enclave_init.c:169
#4  0x00007fe0005a93e5 in __sgx_init_enclave () at enclave/enclave_oe.c:159
#5  0x00007fe0005a9c2d in sgxlkl_enclave_init (shared_memory=0x7fe000b03160) at enclave/enclave_oe.c:382
#6  0x00007fe0005d32e9 in ecall_sgxlkl_enclave_init (input_buffer=0x7fe000b03150 "", input_buffer_size=112, output_buffer=0x7fe000b031c0 "", output_buffer_size=16, output_bytes_written=0x7fe040f03c68) at en
#7  0x00007fe0004f7fdc in oe_handle_call_enclave_function (arg_in=140737204657448) at /home/azureuser/sgx-lkl/openenclave/enclave/core/sgx/calls.c:289
#8  0x00007fe0004fa06e in _handle_ecall (td=0x7fe040f0a000, func=2, arg_in=140737204657448, output_arg1=0x7fe040f03fc8, output_arg2=0x7fe040f03fc0) at /home/azureuser/sgx-lkl/openenclave/enclave/core/sgx/ca
[...]

PR #788, opened 2020-08-14
Make shutdown sequence more robust under cooperative scheduling

### LINKED ISSUE #733: Segfault during SGX-LKL shutdown
I observed the following failure during a CI run of `kernel-syscalls-mmap-mmap11-(nonrelease)-(run-hw)-(8-ethreads)`:
```
[   SGX-LKL  ] Warning: SGXLKL_TRACE_SIGNAL ignored in non-debug mode.
[   SGX-LKL  ] nproc=4 ETHREADS=8 CMDLINE="mem=512m"
[   SGX-LKL  ] HW TLS support: econf->fsgsbase=1
[   SGX-LKL  ] Registering disk 0 (path='sgxlkl-miniroot-fs.img', mnt='/', [RW])
[   SGX-LKL  ] get_signed_libsgxlkl_path... result=/opt/sgx-lkl-nonrelease/bin/../lib/libsgxlkl.so.signed
[   SGX-LKL  ] oe_create_enclave...
[   SGX-LKL  ] Enclave config: {"args":["/ltp/testcases/kernel/syscalls/mmap/mmap11"],"clock_res":[{"resolution":"0000000000000001"},{"resolution":"0000000000000001"},{"resolution":"0000000000000000"},{"resolution":"0000000000000000"},{"resolution":"0000000000000001"},{"resolution":"00000000003d0900"},{"resolution":"00000000003d0900"},{"resolution":"0000000000000001"}],"cwd":"/","env":[],"esleep":"16000","espins":"500","ethreads":"8","exit_status":"full","format_version":"1","fsgsbase":true,"host_import_env":[],"hostname":"lkl","hostnet":false,"image_sizes":{"num_heap_pages":"262144","num_stack_pages":"1024"},"io":{"block":true,"console":true,"network":true},"kernel_cmd":"mem=512m","kernel_verbose":true,"max_user_threads":"256","mmap_files":"shared","mode":"hw_debug","mounts":[],"net_gw4":"10.0.1.254","net_ip4":"10.0.1.1","net_mask4":"24","oe_heap_pagecount":"8192","root":{"key":"","key_id":null,"overlay":false,"readonly":false,"roothash":null,"roothash_offset":"0"},"stacksize":"524288","swiotlb":true,"sysctl":null,"tap_mtu":"0","verbose":true,"wg":{"ip":"10.0.2.1","key":null,"listen_port":"56002","peers":[]}}
result=0 (OE_OK)
[    0.000000] Linux version 5.3.0+ (AzDevOps@cisca24ba0002V4) (gcc version 7.5.0 (Ubuntu 7.5.0-3ubuntu1~18.04)) #1 Thu Aug 6 09:09:59 UTC 2020
[    0.000000] memblock address range: 0x7f00207c2000 - 0x7f00407c2000
[    0.000000] Built 1 zonelists, mobility grouping on.  Total pages: 129280
[    0.000000] Kernel command line: mem=512m console=hvc0 virtio_mmio.device=256@0x1000000:1
[    0.000000] Dentry cache hash table entries: 65536 (order: 7, 524288 bytes, linear)
[    0.000000] Inode-cache hash table entries: 32768 (order: 6, 262144 bytes, linear)
[    0.000000] mem auto-init: stack:off, heap alloc:off, heap free:off
[    0.000000] Memory available: 516256k/524288k RAM
[    0.000000] SLUB: HWalign=32, Order=0-3, MinObjects=0, CPUs=1, Nodes=1
[    0.000000] NR_IRQS: 4096
[    0.000000] lkl: irqs initialized
[    0.000000] clocksource: lkl: mask: 0xffffffffffffffff max_cycles: 0x1cd42e4dffb, max_idle_ns: 881590591483 ns
[    0.000000] lkl: time and timers initialized (irq2)
[    0.000559] printk: console [hvc0] enabled
[    0.000559] pid_max: default: 4096 minimum: 301
[    0.000560] Mount-cache hash table entries: 1024 (order: 1, 8192 bytes, linear)
[    0.000560] Mountpoint-cache hash table entries: 1024 (order: 1, 8192 bytes, linear)
[    0.002735] devtmpfs: initialized
[    0.018031] random: get_random_bytes called from 0x00007f000057a455 with crng_init=0
[    0.018032] clocksource: jiffies: mask: 0xffffffff max_cycles: 0xffffffff, max_idle_ns: 19112604462750000 ns
[    0.018033] futex hash table entries: 16 (order: -4, 384 bytes, linear)
[    0.018584] xor: automatically using best checksumming function   8regs     
[    0.018586] NET: Registered protocol family 16
[    0.021872] clocksource: Switched to clocksource lkl
[    0.022983] NET: Registered protocol family 2
[    0.022985] tcp_listen_portaddr_hash hash table entries: 256 (order: 0, 4096 bytes, linear)
[    0.022985] TCP established hash table entries: 4096 (order: 3, 32768 bytes, linear)
[    0.022986] TCP bind hash table entries: 4096 (order: 3, 32768 bytes, linear)
[    0.022986] TCP: Hash tables configured (established 4096 bind 4096)
[    0.023539] UDP hash table entries: 256 (order: 1, 8192 bytes, linear)
[    0.023539] UDP-Lite hash table entries: 256 (order: 1, 8192 bytes, linear)
[    0.023540] NET: Registered protocol family 1
[    0.023541] virtio-mmio: Registering device virtio-mmio.0 at 0x1000000-0x10000ff, IRQ 1.
[    0.024095] workingset: timestamp_bits=62 max_order=18 bucket_order=0
[    0.025208] NET: Registered protocol family 38
[    0.025208] io scheduler mq-deadline registered
[    0.025759] io scheduler kyber registered
[    0.113409] software IO TLB: mapped [mem 0x7f10322c1000-0x7f10362c1000] (64MB)
[    0.116643] loop: module loaded
[    0.117200] device-mapper: ioctl: 4.40.0-ioctl (2019-01-18) initialised: dm-devel@redhat.com
[    0.126234] wireguard: WireGuard 0.0.20191219 loaded. See www.wireguard.com for information.
[    0.126235] wireguard: Copyright (C) 2015-2019 Jason A. Donenfeld <Jason@zx2c4.com>. All Rights Reserved.
[    0.126789] NET: Registered protocol family 10
[    0.127333] Segment Routing with IPv6
[    0.127886] sit: IPv6, IPv4 and MPLS over IPv4 tunneling driver
[    0.128946] This architecture does not have kernel memory protection.
[    0.129500] Run /init as init process
[    0.131722] random: crng init done
[[  SGX-LKL ]] wg0 has public key W5/XFYlRAoayhmNTBF1ikKfZZapwu2VaDBW1TXPfkB0=
[    0.133394] virtio_blk virtio1: [vda] 3072000 512-byte logical blocks (1.57 GB/1.46 GiB)
[    0.135517] EXT4-fs (vda): mounted filesystem with ordered data mode. Opts: 
mmap11      1  [1;32mTPASS[0m  :  test completed.
[    0.147612] reboot: Restarting system
[   SGX-LKL  ] ethread (5: 0) ethread (6: 0) ethread (3: 0) ethread (2: 0) init (0: 0 exit=0) [[  SGX-LKL ]] FAIL: ethread (4: 0) Exception SIGSEGV (page fault) received before LKL initialisation/after LKL shutdown (lt->tid=-1 [(?)] code=5 addr=0x7f00000ef327 opcode=0x3883 ret=0)
2020-08-06T09:36:05.000000Z [(H)ERROR] tid(0x7f10277fe700) | :OE_ENCLAVE_ABORTING [/agent/_work/1/s/openenclave/host/calls.c:oe_call_enclave_function_by_table_id:91]
ethread (7: 19) [   SGX-LKL  ] FAIL: sgxlkl_ethread_init() failed (id=7 result=19 (OE_ENCLAVE_ABORTING))
TEST_FAILED EXIT CODE: 1
```
This happens to one of the ethreads, after 6 of the 8 ethreads have already exited the enclave. Note that the signal handler cannot retrieve the tid anymore.

I suspect that the test here is irrelevant, and this is rather a rare race condition/memory corruption during shutdown. It would be good to observe this for a DEBUG build and get a stack trace. It seems different from the other issues that we have seen.

prp:
This PR changes the shutdown sequence to make it more robust under cooperative scheduling. The main idea is to use the idle host task for the shutdown and only allow a single ethread in the enclave during shutdown. In addition, it introduces a separate lthread to run the application's main function and makes this lthread exit in a regular fashion.

The PR also fixes a memory corruption issue that potentially affected shutdown, and ensures that userspace threads cannot make SGX-LKL shutdown hang.

The problem, illustrated by a new exit spin test in this PR, is that an application that has a thread with a busy compute loop can make SGX-LKL shutdown hang. There is no clean way under cooperative scheduling to make the associated ethread exit the enclave. In the current shutdown design, we need all ethreads to exit the enclave, which is brittle.

In this design, a single ethread is designated the "terminating ethread" and executes the shutdown code after the other ethreads have left the enclave. Only the terminating ethread needs to exit the enclave successfully. The launcher process will then quit, with ethreads potentially still executing in the enclave.

Fixes #733 #795 #796 #797 #802 

Known limitations
------------------

1. The clone CI test sometimes fails:
```
[[  SGX-LKL ]] [7f0042b46f78] [  44] app_main_thread(): Invoking dynamic loader (stage 3)
Clone syscall number: 56
lkl_syscall: 0
fn: 0x7f003daddce7 
ctid: 0x7f003de69dd8 
Stack: 0x7f003dadb000-0x7f003dadd000 
Clone returned 43, ctid: 43 ptid: 43
New thread created.
Arg: 0x42.
Stack: 0x7f003dadcfe7.
After futex call, ctid is 0
Other thread should have terminated by now.
Thread stacks: 0x7f003dadb000, 0x7f003dce20a0, 0x7f003dce40a0
Thread 0 started
Thread 1 started
Created two threads: 44, 45, waking them now

If this test hangs after waking up one thread, check you have at least 2 ethreads
Thread 0 started
Thread 1 started
Created two threads: 44, 45, waking them now
Thread This test is checking that LKL is able to wake up two cloned threads and leaving them running in parallel

1 woke up
tarted
Thread 0 woke up
Thread 0 finished
Waiting for for Thread 1 to finish
Segmentation fault
[[  SGX-LKL ]] [7f0042b46f78] [  41] lkl_terminate(): terminating LKL (exit_status=139)
```
Even when the clone test passes, I see garbage characters being printed (scroll to the right below to see them):
```
[[  SGX-LKL ]] [7f504171ef78] [  44] app_main_thread(): Invoking dynamic loader (stage 3)
Clone syscall number: 56
lkl_syscall: 0
fn: 0x7f503daddce7
ctid: 0x7f503de69dd8
Stack: 0x7f503dadb000-0x7f503dadd000
Clone returned 43, ctid: 43 ptid: 43
New thread created.
Arg: 0x42.
Stack: 0x7f503dadcfe7.
After futex call, ctid is 0
Other thread should have terminated by now.
Thread stacks: 0x7f503dadb000, 0x7f503dce20a0, 0x7f503dce40a0
Thread 0 started                                                                                                                                                                                           Thread 0 woke up                                                                                                                                                                                           >P                                                                                                                                                                                                          Pd                                                                                                                                                                                                        Created two threads: 4                                                                                                                                                                                     If this test hangs after waking up one thread, check you have at least 2 ethreads
This test is checking that LKL is able to wake up two cloned threads and leaving them running in parallel

Thread 1 woke up
Waiting for for Thread 0 to finish
Waiting for for Th
TEST_PASSED
```
The same corruption happens in current runs of the clone test (without this PR), so I believe that the above is not a new regression introduced by this PR (cc: @davidchisnall @vtikoo @SeanTAllen). This may also be related to #740.

2. The enclave cannot be shut down with an OE API call if there are still ethreads executing. The `oe_terminate_enclave()` call results in a segfault in such cases, which may be an OE bug: 
```
[[  SGX-LKL ]] [7f1041b26f78] [  42] lkl_termination_thread(): done
[   SGX-LKL  ] ethread exited (ethread_id=3)
[   SGX-LKL  ] oe_terminate_enclave... 2020-08-15T19:11:30.000000Z [(H)VERBOSE] tid(0x7f28bed85300) | /home/v-pepiet/devel/sgx-lkl/build_musl/libsgxlkl.so.signed 0x0 OE_ECALL: DESTRUCTOR
2020-08-15T19:11:30.000000Z [(H)VERBOSE] tid(0x7f28bed85300) | _do_eenter(tcs=0x7f1041b21000 aep=0x4008a120 codeIn=1, funcIn=0 argIn=0)
 [/home/v-pepiet/devel/sgx-lkl/openenclave/host/sgx/calls.c:_do_eenter:178]
2020-08-15T19:11:30.000000Z [(H)VERBOSE] tid(0x7f28bed85300) | /home/v-pepiet/devel/sgx-lkl/build_musl/libsgxlkl.so.signed 0x0 EDL_OCALL: CALL_HOST_FUNCTION
2020-08-15T19:11:30.000000Z [(E)INFO] tid(0x7f28bed85300) | libsgxlkl.so.signed:attesters is NULL [/home/v-pepiet/devel/sgx-lkl/openenclave/enclave/sgx/attester.c:oe_attester_shutdown:512]
2020-08-15T19:11:30.000000Z [(H)VERBOSE] tid(0x7f28bed85300) | /home/v-pepiet/devel/sgx-lkl/build_musl/libsgxlkl.so.signed 0x0 EDL_OCALL: CALL_HOST_FUNCTION
2020-08-15T19:11:30.000000Z [(E)INFO] tid(0x7f28bed85300) | libsgxlkl.so.signed:verifiers is NULL [/home/v-pepiet/devel/sgx-lkl/openenclave/common/sgx/verifier.c:oe_verifier_shutdown:847]
2020-08-15T19:11:30.000000Z [(H)VERBOSE] tid(0x7f28aeffd700) | /home/v-pepiet/devel/sgx-lkl/build_musl/libsgxlkl.so.signed 0x0 EDL_OCALL: CALL_HOST_FUNCTION
2020-08-15T19:11:30.000000Z [(H)VERBOSE] tid(0x7f28b4f56700) | /home/v-pepiet/devel/sgx-lkl/build_musl/libsgxlkl.so.signed 0x0 EDL_OCALL: CALL_HOST_FUNCTION
2020-08-15T19:11:30.000000Z [(H)ERROR] tid(0x7f28adffb700) | tcs=0x41f29000
 [/home/v-pepiet/devel/sgx-lkl/openenclave/host/sgx/enclavemanager.c:oe_query_enclave_instance:204]
Makefile:56: recipe for target 'run-spin-hw' failed
make: *** [run-spin-hw] Aborted (core dumped)
```
This PR avoids calling `oe_terminate_enclave()` if not all ethreads have exited, which shouldn't be a problem as I presume that the enclave will be terminated correctly on process termination.

3. There are rare cases when the idle thread is seen to be doing too much work by the kernel, and there is a failure. A better solution would be to have a separate kernel host task to execute the termination code, but this would require larger changes.


src/lkl/setup.c
```diff
@@ -1285,16 +1285,20 @@ static void* lkl_termination_thread(void* args)
...
-    res = lkl_sys_halt();
-    if (res < 0)
-    {
-        sgxlkl_fail("LKL halt, %s\n", lkl_strerror(res));
-    }
+    /**
+     * If kernel threads are stuck, this may block indefinitely under
+     * cooperative scheduling.
+     */
+    // SGXLKL_VERBOSE("calling lkl_sys_halt()\n");
```
SeanTAllen: was this commented out code left behind by mistake?

prp: I would like to have a discussion here if it is the right thing to remove this (at least for now). It may even make sense to leave in the comment, so we don't forget to think of a way to put it back in.

src/main-oe/sgxlkl_run_oe.c
```diff
@@ -2032,24 +2037,26 @@ int main(int argc, char* argv[], char* envp[])
...
+    long exit_status = (long)enclave_return_status;
+    pthread_mutex_unlock(&first_ethread_exited_mtx);
+
+    /**
+     * This fails with an error when we try to terminate the enclave while
+     * ethreads are still executing (?). We need to figure out a way to shutdown
+     * the enclave under coperative scheduling, i.e., when an ethread may be
+     * stuck executing an application busy-loop.
+     */
+    // if (oe_enclave)
```
SeanTAllen: What about this commented out code?

prp: Same point as above. We should have a discussion if it is right to remove this. I also don't understand why the shutdown doesn't work anymore.

src/lkl/setup.c
SeanTAllen: I'd rather see an issue or issues opened with a link back to the removal commit.

prp: Sure, I can do this once we've decided if this shutdown design makes more sense than the current one.

src/sched/lthread.c
```diff
@@ -240,6 +240,17 @@ void lthread_run(void)
...
                 _lthread_resume(lt);
+
+                // Break out of scheduler loop when previous thread triggered
+                // termination
+                if (_lthread_should_stop)
+                {
+                    SGXLKL_TRACE_THREAD(
+                        "[%4d] lthread_run(): quitting\n", lt ? lt->tid : -1);
+                    // We only need this ethread to leave the enclave
+                    _lthread_should_stop = false;
```
davidchisnall: I don't understand what's happening here.  One or more ethreads will reach this path, because this is a read then write, not a CAS.  Why do we want a nondeterministic number of ethreads to hit this path?

src/enclave/enclave_oe.c
```diff
@@ -295,7 +295,7 @@ void sgxlkl_ethread_init(void)
     _lthread_sched_init(sgxlkl_enclave_state.config->stacksize);
     lthread_run();
 
-    return;
+    return sgxlkl_enclave_state.exit_status;
```
davidchisnall: Please add a comment explaining this.  I guess it's so that one ethread exits while another can be stuck in a non-yielding loop, and the first one to exit provides the return value?

src/sched/lthread.c
prp: Yes, this should be CAS. The idea is that we will only have a single ethread exit the enclave.

src/enclave/enclave_oe.c
prp: Yes, we don't know which of our ethreads (either the first `einit` ethread or any of the subsequent ethreads) will be exiting, so we need to propagate the exit code on both code paths.

src/lkl/setup.c
```diff
@@ -1288,16 +1304,20 @@ static void* lkl_termination_thread(void* args)
...
-    }
 
-    /* Notify host about the guest shutdown */
+    SGXLKL_VERBOSE("calling sgxlkl_host_shutdown_notification()\n");
     sgxlkl_host_shutdown_notification();
 
-    /* Set termination flag to notify lthread scheduler to bail out. */
-    lthread_notify_completion();
+    SGXLKL_VERBOSE("calling vio_terminate()\n");
+    vio_terminate();
```
davidchisnall, line 1296: Should this go before the `lkl_sys_halt` call?  Or will VirtIO interrupts be silently ignored if they happen after the kernel has exited?  I vaguely remember some shutdown-gate code for this, but it would be good to have a comment explaining this assumption.

src/lkl/setup.c
```diff
@@ -1266,6 +1272,16 @@ static void* lkl_termination_thread(void* args)
         }
     }
 
+    SGXLKL_VERBOSE("calling lkl_sys_sync()\n");
+    ret = lkl_sys_sync();
```
davidchisnall: Have we done something here to prevent any threads that we failed to kill from doing system calls?  We can still end up with the disk in an undefined state if not:

 1. Sync call flushes everything
 2. User thread does some disk writes
 3. A few of them propagate to the VirtIO device
 4. We kill the kernel.

With Vikas' work, we now differentiate between userspace and kernel threads in lthreads.  Maybe we should set a flag so that our lthread scheduler refuses to run any userspace threads at this point?

src/main-oe/sgxlkl_run_oe.c
```diff
@@ -2032,25 +2050,35 @@ int main(int argc, char* argv[], char* envp[])
...
-    sgxlkl_host_verbose_raw("\n");
-
-    if (oe_enclave)
+    // Wait for the terminating ethread to exit the enclave
+    pthread_mutex_lock(&terminating_ethread_exited_mtx);
+    ret = pthread_cond_wait(
+        &terminating_ethread_exited_cv, &terminating_ethread_exited_mtx);
+    if (ret != 0)
+        sgxlkl_host_fail("Failed to wait for enclave to finish: ret=%i\n", ret);
+    long exit_status = (long)enclave_return_status;
```
davidchisnall: Why an explicit cast from `long` to `long`?

src/main-oe/sgxlkl_run_oe.c
```diff
@@ -1304,18 +1314,29 @@ void parse_cpu_affinity_params(char* config, int** cores, size_t* cores_len)
...
+    // Check if this is the terminating ethread that returns the exit
+    // status
+    if (exit_status != INT_MAX)
     {
-        sgxlkl_host_verbose("");
+        // Signal that the terminating ethread has exited
+        pthread_mutex_lock(&terminating_ethread_exited_mtx);
+        enclave_return_status = exit_status;
+        int ret = pthread_cond_signal(&terminating_ethread_exited_cv);
+        if (ret != 0)
```
davidchisnall: Do you expect this to fail (why are we checking for errors on the signal, but not the mutex lock / unlock)?

src/main-oe/sgxlkl_run_oe.c
```diff
@@ -88,6 +88,16 @@ typedef struct ethread_args
     oe_enclave_t* oe_enclave;
 } ethread_args_t;
 
+// CV to wait for the terminating ethreads, which shuts down the enclave
+static pthread_cond_t terminating_ethread_exited_cv;
+static pthread_mutex_t terminating_ethread_exited_mtx;
+
+// Exit status returned from the enclave
+static long enclave_return_status = 0;
```
davidchisnall: Initialise to `LONG_MAX` so we can differentiate a real return value?

src/main-oe/sgxlkl_run_oe.c
```diff
@@ -2032,25 +2050,35 @@ int main(int argc, char* argv[], char* envp[])
...
-        {
-            exit_status = (long)return_value;
-        }
-    }
-    sgxlkl_host_verbose_raw("\n");
-
-    if (oe_enclave)
+    // Wait for the terminating ethread to exit the enclave
+    pthread_mutex_lock(&terminating_ethread_exited_mtx);
+    ret = pthread_cond_wait(
```
davidchisnall: For very short-lived enclaves or some pathological scheduler conditions, we can exit before we sleep and then never wake.  We should check `enclave_return_status` before exiting here.

src/main-oe/sgxlkl_run_oe.c
```diff
@@ -2032,25 +2050,35 @@ int main(int argc, char* argv[], char* envp[])
...
+        &terminating_ethread_exited_cv, &terminating_ethread_exited_mtx);
+    if (ret != 0)
+        sgxlkl_host_fail("Failed to wait for enclave to finish: ret=%i\n", ret);
+    long exit_status = (long)enclave_return_status;
+    pthread_mutex_unlock(&terminating_ethread_exited_mtx);
+
+    // Only try to destroy enclave if all ethreads successfully exited
+    if (oe_enclave && exited_ethread_count == econf->ethreads)
```
davidchisnall, line 2066: Why?  Don't we want to kill the enclave as long as one thread has delivered to us a return value?  The other threads are not doing anything we care about...

src/sched/futex.c
```diff
@@ -45,6 +45,11 @@ static uint32_t to_futex_key(int* uaddr)
     return (uint32_t)uaddr;
 }
 
+void futex_init()
```
davidchisnall, line 48: This is an lthread futex call, but declared in the pthread futex header.

src/sched/lthread.c
```diff
@@ -233,13 +231,31 @@ void lthread_run(void)
...
-                a_dec(&schedqueuelen);
                 SGXLKL_TRACE_THREAD(
                     "[%4d] lthread_run(): lthread_resume (dequeue)\n",
                     lt ? lt->tid : -1);
                 _lthread_resume(lt);
+
+                // Bail out if there is a terminating scheduler, and we are not
+                // it.
+                if (_lthread_terminating_scheduler &&
+                    _lthread_terminating_scheduler != sched)
```
davidchisnall: Two reads of an `_Atomic` make me nervous.  I *think* this is okay, because `_lthread_terminating_scheduler` can't be modified in such a way that the second condition becomes true, but I wouldn't need to think that through if we did a single read.

src/sched/lthread.c
```diff
@@ -233,13 +231,31 @@ void lthread_run(void)
...
             {
+                SGXLKL_ASSERT(!(lt->attr.state & BIT(LT_ST_EXITED)));
+
                 dequeued++;
                 pauses = sleepspins;
-                a_dec(&schedqueuelen);
                 SGXLKL_TRACE_THREAD(
                     "[%4d] lthread_run(): lthread_resume (dequeue)\n",
                     lt ? lt->tid : -1);
                 _lthread_resume(lt);
```
davidchisnall, line 242: Maybe skip non-kernel threads here if `_lthread_terminating_scheduler == sched`?

src/enclave/enclave_oe.c
```diff
@@ -294,9 +294,20 @@ void sgxlkl_ethread_init(void)
...
+    // from the enclave.
+    if (terminating_ethread)
+    {
+        return sgxlkl_enclave_state.exit_status;
+    }
+    else
+    {
+        // This ethread is not the terminating ethread
+        return INT_MAX;
+    }
```
davidchisnall: This code seems to be duplicated in both callers of `lthread_run`, can we move it into `lthread_run`?

src/sched/futex.c
prp: Ah, good point. I put it there because `futex_tick()` is already defined there. I presume that should be moved as well?

src/enclave/enclave_oe.c
prp: Done.

src/main-oe/sgxlkl_run_oe.c
prp: Done.

src/main-oe/sgxlkl_run_oe.c
prp: Check removed.

src/main-oe/sgxlkl_run_oe.c
prp: Fixed.

src/main-oe/sgxlkl_run_oe.c
prp: Hmm, but even if we check `enclave_return_status` before waiting, there would still be a race condition, no?

src/main-oe/sgxlkl_run_oe.c
prp: Right, but the problem is that the OE terminate API call segfaults in these cases: https://github.com/lsds/sgx-lkl/pull/788#issue-467907462

src/sched/lthread.c
prp: Hmm, couldn't this lead to deadlock in the LKL scheduler if it wants to schedule them?

src/lkl/setup.c
prp: If we move it earlier, there may still be background I/O by the kernel, no? (I tried this, and I saw failures.)

src/sched/futex.c
prp: There are also other changes to the file, which I don't quite understand.

src/lkl/setup.c
prp: Right, but wouldn't we have to kill them in the LKL scheduler? If the lthread scheduler terminates a userspace thread while it holds the CPU lock, we would deadlock.

src/main-oe/sgxlkl_run_oe.c
davidchisnall: No, the value is only set with the mutex held, so there is an enforced ordering of either check, mismatch, sleep, set wake, or set, check, don’t sleep.

src/sched/lthread.c
prp: What code do you have in mind?

src/main-oe/sgxlkl_run_oe.c
davidchisnall: We probably need to file that as an OE bug and link to that issue here. If you can’t tear down an uncooperative enclave, that’s a problem. That said, what happens if we send a signal to all of the host pthreads that back our ethreads first? Does that not let us exit them all from the enclave and then tear it down?

src/sched/futex.c
davidchisnall: Yes, I think they should both live in lthread.h.

src/sched/lthread.c
davidchisnall: Yes, possibly. Does LKL allow syscalls to enter the kernel after shutdown has started? It probably shouldn’t.

src/lkl/setup.c
davidchisnall: Yes, I think we have three cases. Kernel threads are fine to run. Userspace threads executing in userspace are fine to stop. Userspace threads executing in the kernel are a problem. Hopefully LKL prevents any of the second category from becoming the third, so the only thing that can happen from running a userspace thread is that it exits the kernel.

src/sched/lthread.c
davidchisnall: Read the atomic variable to a local and use that.

src/main-oe/sgxlkl_run_oe.c
prp: Done.

src/sched/futex.c
prp: Moved and reverted futex.h to the host-musl version.

src/sched/lthread.c
prp: Done.

src/sched/lthread.c
prp: Currently we don't: https://github.com/lsds/lkl/compare/385f721b339fe48b188b4924c2663e1ea2cdeb13...8552da7975f1db1e178b802a28409ea2c6d12aab#diff-040a3c76c7eac15c7395c8f8a408af4e

What error should we return to the userspace thread when it's issuing a syscall? I'm worried that this would result in weird errors in the application code. It would be better to also deallocate the userspace thread. I still think that the kernel should be killing userspace threads during shutdown. I don't understand why not all threads get killed (which also leads to the problem in `lkl_sys_halt()` -- see the comment there.)

Perhaps we can try this a part of another PR, as this one is already getting too big?

prp: This may work, but let's try this in another PR.

prp: I've now added a guard that ensures that syscalls during shutdown can only come from the termination thread.

tests/ltp/ltp-batch1/ltp_disabled_tests.txt
```diff
@@ -49,7 +49,7 @@
 #/ltp/testcases/kernel/syscalls/chown/chown05
```
hukoyu: These four tests are already disabled and merged in PR https://github.com/lsds/sgx-lkl/pull/812

tests/ltp/run_ltp_test.sh
```diff
@@ -84,8 +84,8 @@ for file in "${ltp_tests[@]}"; do
       exit 1
     fi
 
-    echo "SGXLKL_CMDLINE=mem=512m SGXLKL_VERBOSE=1 SGXLKL_KERNEL_VERBOSE=1 SGXLKL_TRACE_SIGNAL=1 timeout $timeout ${SGX_LKL_RUN_CMD[*]} $file > $stdout_file 2>&1"
-    SGXLKL_CMDLINE="mem=512m" SGXLKL_VERBOSE=1 SGXLKL_KERNEL_VERBOSE=1 SGXLKL_TRACE_SIGNAL=1 timeout $timeout "${SGX_LKL_RUN_CMD[@]}" "$file" > "$stdout_file" 2>&1
+    echo "SGXLKL_VERBOSE=1 SGXLKL_KERNEL_VERBOSE=1 SGXLKL_TRACE_SIGNAL=1 timeout $timeout ${SGX_LKL_RUN_CMD[*]} $file > $stdout_file 2>&1"
```
hukoyu: This is already removed and merged in PR https://github.com/lsds/sgx-lkl/pull/812

src/sched/lthread.c
```diff
@@ -348,9 +361,11 @@ void _lthread_yield(struct lthread* lt)
 
 void _lthread_free(struct lthread* lt)
 {
-    if (lthread_self() != NULL)
+    if (lthread_self() != NULL && !(lt->attr.state & BIT(LT_ST_APP_MAIN)))
```
vtikoo, line 359: Maybe a comment here explaining why appmain thread is skipped?

src/enclave/mpmc_queue.c
```diff
@@ -38,8 +38,8 @@
 #include "enclave/enclave_util.h"
 #include "enclave/mpmc_queue.h"
 
-static void pause(void);
-static void pause()
+static void _mpmc_pause(void);
+static void _mpmc_pause()
 {
     __asm__ __volatile__("pause" : : : "memory");
```
davidchisnall:
```suggestion
#ifdef __x86__
        __builtin_ia32_pause()
#endif
```

src/enclave/enclave_init.c
```diff
@@ -89,7 +89,29 @@ static void init_wireguard()
         wgu_list_devices();
 }
 
-static int startmain(void* args)
+static int app_main_thread(void* args)
+{
+    SGXLKL_VERBOSE("enter\n");
+
+    lthread_set_funcname(lthread_self(), "app-main");
+    lthread_set_app_main();
```
davidchisnall: Can this be done in the caller?  We're mixing lthread and pthread calls here, which is going to make finishing the layering cleanup harder.  All of this function should be considered part of userspace.  It makes me slightly unhappy that this thread is created with `lthread_create` and not a clone `lkl_syscall`, but mixing lthread and pthread calls within the same function is definitely not something we should be doing.

src/include/enclave/enclave_util.h
```diff
@@ -138,10 +138,22 @@ extern int sgxlkl_trace_redirect_syscall;
         }                                                     \
     } while (0)
```
davidchisnall, line 140:
```suggestion
/**
 * Pointer to the thread context.  This is needed because of limitations in the C address space syntax.
 */
typedef struct schedctx* threadctx_ptr;
/**
 * Pointer to the per-ethread context.
 */
static const __seg_gs threadctx_ptr *ethread_schedctx = ((__seg_gs threadctx_ptr *)48);
```

src/include/enclave/enclave_util.h
```diff
@@ -138,10 +138,22 @@ extern int sgxlkl_trace_redirect_syscall;
...
 
-#define SGXLKL_VERBOSE(x, ...)                                              \
-    if (sgxlkl_enclave_state.verbose)                                       \
-    {                                                                       \
-        oe_host_printf("[[  SGX-LKL ]] %s(): " x, __func__, ##__VA_ARGS__); \
+#define SGXLKL_VERBOSE(x, ...)                                               \
+    if (sgxlkl_enclave_state.verbose)                                        \
+    {                                                                        \
+        struct schedctx* _sgxlkl_verbose_self;                               \
+        __asm__ __volatile__("mov %%gs:48,%0" : "=r"(_sgxlkl_verbose_self)); \
```
davidchisnall, line 145:
```suggestion
        struct schedctx* _sgxlkl_verbose_self = *ethread_schedctx;           \
```

All of the assembly for accessing this with hard-coded constants all over the place makes me nervous.  Let's add a global for this and clean up code to use that.  GCC and clang have good support for gs-relative addresses.

src/include/enclave/lthread.h
```diff
@@ -235,11 +240,29 @@ extern "C"
...
+    /**
+     * Makes all schedulers in all ethreads except the caller's scheduler
+     * terminate and exit the enclave. Afterwards, the enclave will be
+     * single-threaded, with only the current ethread's scheduler running.
+     */
+    void lthread_terminate_other_schedulers(void);
+
+    /**
+     * Make the current scheduler also terminate and exit the enclave after the
+     * calling lthread has returned.
```
davidchisnall:
```suggestion
     * calling lthread has next yields.
```

src/sched/lthread.c
prp: Will do. I think the reason is the same why we can't run the destructors for cloned host tasks: otherwise the destructor will deallocate the host task again. (Without this, it segfaults when deallocating the main app thread.)

src/include/enclave/enclave_util.h
vtikoo: Will this have the value of the first ethread which evaluates this or it indeed would act as ethread local variable?

davidchisnall: Accesses to this translate to `%gs:48`.

src/enclave/enclave_init.c
prp: Ok, I moved this out.

src/enclave/mpmc_queue.c
prp: Done.

src/include/enclave/enclave_util.h
prp: Ok, I couldn't get this to work, as I presume that this needs gnu99 support? For this, we would also need to change musl to be compiled with gnu99...?

src/lkl/setup.c
prp: I've added a guard to only allow the idle host task to do system calls during shutdown.

davidchisnall (approved):

davidchisnall:
@vtikoo, since you'll probably end up merging this one, please will you do the following sequence:

1. Squash and merge the LKL and Musl PRs.
2. Add a commit to this branch updating the LKL and musl submodules to be the post-merge commits.
3. Squash and merge this, cleaning up the commit messages.

Thank you!

vtikoo (approved):

PR #798, opened 2020-08-17
Implement hook to pass cpuinfo to LKL with test

AntonioND:
This PR depends on https://github.com/lsds/lkl/pull/7 being merged. It has to be rebased and the gitmodules reference has to be updated.


src/lkl/posix-host.c
```diff
@@ -756,6 +757,16 @@ static void host_free(void *ptr)
     oe_free(ptr);
 }
 
+/**
+ * Returns the number of CPU cores to be displayed in cpuinfo
+ */
+unsigned int sgxlkl_cpuinfo_get_cpu_cores(void)
```
davidchisnall: As I commented in the other PR, I think this should provide the complete `/proc/cpuinfo` string.

tests/basic/print_cpuinfo/print_cpuinfo.c
```diff
@@ -0,0 +1,58 @@
...
+            if (num_cores != 1) {
+                printf("Unexpected number of cores %u (should be 1)\n",
+                    num_cores);
+                ret = 1;
+                break;
+            }
+        }
+    }
+
+    free(line);
```
davidchisnall: Add a `sysinfo` check here as well?

tests/basic/print_cpuinfo/print_cpuinfo.c
```diff
@@ -0,0 +1,58 @@
...
+        printf("%s", line);
+
+        if (strstr(line, "cpu cores")) {
+            unsigned int num_cores = 0;
+            if (sscanf(line, "cpu cores       : %u", &num_cores) != 1) {
+                printf("Failed to sscanf cpu cores information\n");
+                ret = -1;
+                break;
+            }
+            if (num_cores != 1) {
```
davidchisnall: If we run with more than 1 ethread, this test will fail?  Please pass `SGXLKL_ETHREADS` in as an environment variable that's propagated to the program and check that the value matches.

davidchisnall (changes requested):

tests/basic/print_cpuinfo/print_cpuinfo.c
AntonioND: Done. It is passed as an argument because passing it as an environment variable was a bit tricky and I couldn't get it working easily. Maybe I just missed something.

tests/basic/print_cpuinfo/print_cpuinfo.c
AntonioND: Done.

src/lkl/posix-host.c
AntonioND: Done.

davidchisnall (approved):

vtikoo (approved):
LGTM

PR #804, opened 2020-08-18
Delete dead code.

davidchisnall:
Now that lthreads aren't implementing pthreads, we don't need code for
some of the scary corner cases of pthreads.  Remove the cancellation
code that is on the critical path for context switches and is now dead.


vtikoo (approved):

prp (approved):
LGTM.

PR #805, opened 2020-08-19
Clean up the ticket lock implementation.

davidchisnall:
This reduces the amount of undefined behaviour in the ticket lock
considerably.  I don't remember what WG14 eventually decided on atomic
operations of different sizes at the same address, I believe this is
non-UB (it's just broken on some POWER CPUs), but it is at least safe on
x86.

At the same time, add some comments explaining what the code is trying
to do so that the next person who needs to understand it takes less time
than me.

### DIFF
--- a/src/include/enclave/ticketlock.h
+++ b/src/include/enclave/ticketlock.h
@@ -94,7 +94,7 @@ static inline int ticket_trylock(struct ticketlock* t)
 {
     // Get the expected next ticket value
     uint32_t me = t->s.users;
-    // Set our exected next-ticket value.
+    // Set our expected next-ticket value.
     uint32_t menew = me + 1;
     // If the ticket lock is not held, t->s.users will equal t->s.ticket.
     // Expect that to be the case.
--- a/src/include/enclave/ticketlock.h
+++ b/src/include/enclave/ticketlock.h
@@ -1,20 +1,51 @@
 #ifndef TICKETLOCK_H
 #define TICKETLOCK_H
 
-#include "atomic.h"
+#include <stdatomic.h>
 
 #if DEBUG
 #include "lthread.h"
 #endif /* DEBUG */
 
+/**
+ * A ticket lock.  This provides a simple spinlock that guarantees that the
+ * lock will be acquired in the order that callers attempt to acquire it.
+ *
+ * This is similar to the mechanism used in some shops, where customers take a
+ * numbered ticket and wait until that number is called.  Threads that attempt
+ * to acquire the lock 'take a ticket' by atomically incrementing one counter
+ * and then wait until that ticket is the current one.  Threads releasing the
+ * lock increment a second counter representing the current ticket.
+ */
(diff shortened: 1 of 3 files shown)


SeanTAllen (approved):

prp (approved):
LGTM

src/include/enclave/ticketlock.h
```diff
@@ -1,55 +1,111 @@
...
     // t->lt = lthread_self();
 #endif /* DEBUG */
 }
 
+/**
+ * Release the ticket lock.
+ */
 static inline void ticket_unlock(struct ticketlock* t)
 {
-    a_barrier();
```
vtikoo, line 38: Not sure if this is a valid concern, but theres nothing guaranteeing if the thread attempting the unlock actually holds the lock.

src/include/enclave/ticketlock.h
```diff
@@ -1,55 +1,111 @@
...
 }
 
+/**
+ * Try to acquire a ticket lock.  Returns zero on success, non-zero on failure.
+ */
 static inline int ticket_trylock(struct ticketlock* t)
 {
+    // Get the expected next ticket value
     uint32_t me = t->s.users;
+    // Set our exected next-ticket value.
```
vtikoo:
```suggestion
    // Set our expected next-ticket value.
```

src/include/enclave/ticketlock.h
```diff
@@ -1,55 +1,111 @@
...
     uint32_t menew = me + 1;
-    uint64_t cmp = ((uint64_t)me << 32) + me;
-    uint64_t cmpnew = ((uint64_t)menew << 32) + me;
+    // If the ticket lock is not held, t->s.users will equal t->s.ticket.
```
vtikoo, line 99: This looks like its relying on zero-initialized memory for the initial state of the `ticketlock` structure. If thats true, should we call that out?

vtikoo: Also t->s.users == t->s.ticket is what it means to be in the locked state if you are using `ticket_lock`. Are these two ways of acquiring locks supposed to be used interchangeably for the same lock?

vtikoo (approved):

src/include/enclave/ticketlock.h
vtikoo: Pls ignore my second comment, I misunderstood what `atomic_fetch_add` does.

src/include/enclave/ticketlock.h
davidchisnall: This is because the lock owner is sometimes not an lthread. Some of the invocations are from the lthread scheduler directly, with no running lthread.

src/include/enclave/ticketlock.h
davidchisnall: It doesn't actually require zero init, it requires that the two fields be initialised to the same value.  Zero init is fine though.

PR #806, opened 2020-08-19
Add proper tracing options, remove getenv calls from enclave code.

### LINKED ISSUE #780: Remove getenv calls before libc startup
[lkl_start_init is using getenv](https://github.com/lsds/sgx-lkl/blob/oe_port/src/lkl/setup.c#L1378-L1410) fairly early in the initialization sequence. Which is requiring an early call to `__init_libc`. 

According to @davidchisnall  -
When we parse the environment variables from the config file, we should pull the SGX-LKL ones into globals and skip them, then we wouldn't need that extra complexity here.
From - https://github.com/lsds/sgx-lkl-musl/pull/18#discussion_r469807014

wintersteiger:
This adds proper tracing options, which removes the need for `getenv` calls in the enclave. Fixes #780.

Goes with https://github.com/lsds/sgx-lkl-musl/pull/36.

### DIFF
--- a/src/enclave/enclave_init.c
+++ b/src/enclave/enclave_init.c
@@ -12,7 +12,7 @@
 #include "enclave/lthread_int.h"
 #include "enclave/wireguard.h"
 #include "enclave/wireguard_util.h"
-#include "shared/env.h"
+#include "shared/util.h"
 
 #include "../../user/userargs.h"
 
@@ -179,7 +179,8 @@ static int app_main_thread(void* args_)
 
 static int kernel_main_thread(void* args)
 {
-    __init_libc(sgxlkl_enclave_state.elf64_stack.envp,
+    __init_libc(
+        sgxlkl_enclave_state.elf64_stack.envp,
         sgxlkl_enclave_state.elf64_stack.argv[0]);
     __libc_start_init();
     a_barrier();
@@ -252,7 +253,7 @@ int __libc_init_enclave(int argc, char** argv)
     max_lthreads = next_power_of_2(max_lthreads);
 
     newmpmcq(&__scheduler_queue, max_lthreads, 0);
-    
+
     init_ethread_tp();
 
     size_t espins = cfg->espins;
--- a/src/enclave/enclave_init.c
+++ b/src/enclave/enclave_init.c
@@ -61,7 +61,11 @@ static void wipe_disk_keys()
     {
         if (disk_state[i].key)
         {
-            memset(disk_state[i].key, 0, disk_state[i].key_len);
+            oe_memset_s(
+                disk_state[i].key,
+                disk_state[i].key_len,
+                0,
+                disk_state[i].key_len);
             oe_free(disk_state[i].key);
         }
         disk_state[i].key = NULL;
--- a/src/enclave/enclave_init.c
+++ b/src/enclave/enclave_init.c
@@ -56,6 +56,8 @@ static void get_disk_keys()
     }
 }
 
+/* Wipe copies of (currently local, in future remote) keys after we don't need
+ * them anymore, i.e. after mounting the respective file systems. */
 static void wipe_disk_keys()
 {
     sgxlkl_enclave_disk_state_t* disk_state = sgxlkl_enclave_state.disk_state;
--- a/src/enclave/enclave_init.c
+++ b/src/enclave/enclave_init.c
@@ -30,6 +30,45 @@ extern struct mpmcq __scheduler_queue;
 _Noreturn void __dls3(elf64_stack_t* conf, void* tos);
 extern void init_sysconf(long nproc_conf, long nproc_onln);
 
+static void get_disk_keys()
+{
+    /* Here or earlier: get keys from remote key provider or auxv. */
+    const sgxlkl_enclave_config_t* cfg = sgxlkl_enclave_state.config;
+    sgxlkl_enclave_disk_state_t* disk_states = sgxlkl_enclave_state.disk_state;
+    if (cfg->root.key)
+    {
+        uint8_t* key = cfg->root.key;
+        size_t len = cfg->root.key_len;
+        disk_states[0].key = oe_malloc(sizeof(uint8_t) * len);
+        memcpy(disk_states[0].key, key, len);
+    }
+    for (size_t i = 0; i < cfg->num_mounts; i++)
+    {
+        if (cfg->mounts[i].key)
+        {
+            uint8_t* key = cfg->mounts[i].key;
+            size_t len = cfg->mounts[i].key_len;
+            disk_states[i + 1].key = oe_malloc(sizeof(uint8_t) * len);
+            memcpy(disk_states[i + 1].key, key, len);
+        }
+    }
+}
+
+static void wipe_disk_keys()
+{
+    sgxlkl_enclave_disk_state_t* disk_state = sgxlkl_enclave_state.disk_state;
+    for (size_t i = 0; i < sgxlkl_enclave_state.num_disk_state; i++)
+    {
+        if (disk_state[i].key)
+        {
+            memset(disk_state[i].key, 0, disk_state[i].key_len);
+            oe_free(disk_state[i].key);
+        }
+        disk_state[i].key = NULL;
+        disk_state[i].key_len = 0;
+    }
+}
+
 static void find_and_mount_disks()
 {
     const sgxlkl_enclave_config_t* cfg = sgxlkl_enclave_state.config;
@@ -41,7 +80,7 @@ static void find_and_mount_disks()
     estate->disk_state = oe_calloc(n, sizeof(sgxlkl_enclave_disk_state_t));
     estate->num_disk_state = n;
 
-    // root disk index
+    // root disk index 0
     estate->disk_state[0].host_disk_index = 0;
 
     for (int i = 0; i < cfg->num_mounts; i++)
@@ -68,7 +107,9 @@ static void find_and_mount_disks()
                 cfg_disk->destination);
     }
 
+    get_disk_keys();
     lkl_mount_disks(&cfg->root, cfg->mounts, cfg->num_mounts, cfg->cwd);
+    wipe_disk_keys();
 }
 
 static void init_wireguard_peers()
--- a/src/enclave/enclave_init.c
+++ b/src/enclave/enclave_init.c
@@ -32,7 +32,9 @@ extern void init_sysconf(long nproc_conf, long nproc_onln);
 
 static void get_disk_keys()
 {
-    /* Here or earlier: get keys from remote key provider or auxv. */
+    /* Here or earlier: get keys from remote key provider or auxv. For now we
+     * only support plaintext keys in sgxlkl_enclave_state.config, which we copy
+     * into the right places here. */
     const sgxlkl_enclave_config_t* cfg = sgxlkl_enclave_state.config;
     sgxlkl_enclave_disk_state_t* disk_states = sgxlkl_enclave_state.disk_state;
     if (cfg->root.key)
--- a/src/enclave/enclave_oe.c
+++ b/src/enclave/enclave_oe.c
@@ -408,6 +408,10 @@ int sgxlkl_enclave_init(const sgxlkl_shared_memory_t* shared_memory)
     // point onwards
     sgxlkl_enclave_state.trace_enabled.verbose =
         sgxlkl_enclave_state.config->verbose;
+    sgxlkl_enclave_state.trace_enabled.internal_syscall =
+        sgxlkl_enclave_state.config->trace.internal_syscall;
+    sgxlkl_enclave_state.trace_enabled.lkl_syscall =
+        sgxlkl_enclave_state.config->trace.lkl_syscall;
 #endif
 
     SGXLKL_VERBOSE("enter\n");
--- a/src/enclave/enclave_oe.c
+++ b/src/enclave/enclave_oe.c
@@ -12,8 +12,8 @@
 #include "enclave/enclave_util.h"
 #include "enclave/lthread.h"
 #include "enclave/lthread_int.h"
-#include "shared/env.h"
 #include "shared/timer_dev.h"
+#include "shared/util.h"
 
 #define AUXV_ENTRIES 13
 
--- a/src/enclave/enclave_oe.c
+++ b/src/enclave/enclave_oe.c
@@ -397,7 +397,7 @@ int sgxlkl_enclave_init(const sgxlkl_shared_memory_t* shared_memory)
 #ifdef DEBUG
     /* Make sure verbosity is off before loading the config (we don't know
      * whether it's enabled yet).*/
-    sgxlkl_enclave_state.verbose = false;
+    sgxlkl_enclave_state.trace_enabled.verbose = false;
 #endif
 
     _read_eeid_config();
@@ -406,7 +406,8 @@ int sgxlkl_enclave_init(const sgxlkl_shared_memory_t* shared_memory)
 #ifdef DEBUG
     // Initialise verbosity setting, so SGXLKL_VERBOSE can be used from this
     // point onwards
-    sgxlkl_enclave_state.verbose = sgxlkl_enclave_state.config->verbose;
+    sgxlkl_enclave_state.trace_enabled.verbose =
+        sgxlkl_enclave_state.config->verbose;
 #endif
 
     SGXLKL_VERBOSE("enter\n");
--- a/src/enclave/enclave_oe.c
+++ b/src/enclave/enclave_oe.c
@@ -299,11 +299,25 @@ static void _read_eeid_config()
     const oe_eeid_t* eeid = (oe_eeid_t*)__oe_get_eeid();
     const char* config_json = (const char*)eeid->data;
 
-    sgxlkl_enclave_config_t* cfg = oe_malloc(sizeof(sgxlkl_enclave_config_t));
-    if (!cfg)
-        sgxlkl_fail("out of memory, cannot allocate enclave config.\n");
-    sgxlkl_read_enclave_config(config_json, cfg, true);
-    sgxlkl_enclave_state.config = cfg;
+    size_t num_pages = 0;
+    const sgxlkl_enclave_config_page_t* cfg_pages =
+        sgxlkl_read_enclave_config(config_json, true, &num_pages);
+
+    /* Make the config pages read-only */
+    for (size_t i = 0; i < num_pages; i++)
+    {
+        int mprotect_ret;
+        sgxlkl_host_syscall_mprotect(
+            &mprotect_ret,
+            (void*)&cfg_pages[i],
+            sizeof(sgxlkl_enclave_config_t),
+            PROT_READ);
+
+        if (mprotect_ret != 0)
+            sgxlkl_warn("Could not protect memory pages for config\n");
+    }
+
+    sgxlkl_enclave_state.config = &cfg_pages->config;
 }
 
 static void _copy_shared_memory(const sgxlkl_shared_memory_t* host)
--- a/src/enclave/enclave_signal.c
+++ b/src/enclave/enclave_signal.c
@@ -5,8 +5,8 @@
 
 #include <asm-generic/ucontext.h>
 
-#include <lkl_host.h>
 #include <lkl/setup.h>
+#include <lkl_host.h>
 #include <string.h>
 
 #include <openenclave/enclave.h>
@@ -16,7 +16,6 @@
 #include "enclave/enclave_util.h"
 #include "enclave/lthread.h"
 #include "enclave/sgxlkl_t.h"
-#include "shared/env.h"
 
 #define RDTSC_OPCODE 0x310F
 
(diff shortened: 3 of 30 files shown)


wintersteiger:
There's one relevant getenv call in musl, see https://github.com/lsds/sgx-lkl-musl/blob/a6360f883fd906b18c1878254547ad72da28d7e6/ldso/dynlink.c#L822. This is now the only tracing/debug env var that gets passed through to the enclave as an env var:
https://github.com/lsds/sgx-lkl/blob/9979a88fa90cca275dcb0822deef5e0393cea88a/src/main-oe/sgxlkl_params.c#L3
I'm not sure this is still useful (see bunch of commented code further up in the file). Can I remove it?

src/lkl/setup.c
```diff
@@ -469,15 +459,12 @@ static void* lkl_create_crypto_disk_thread(struct lkl_crypt_device* lkl_cd)
...
-    {
+    struct crypt_pbkdf_type pbkdf = {
         .type = "pbkdf2",
         .hash = "sha256",
         .iterations = MIN_LUKS2_ITERATIONS,
     };
-    struct crypt_params_luks2 params =
-    {
-        .sector_size = SECTOR_SIZE,
+    struct crypt_params_luks2 params = {.sector_size = SECTOR_SIZE,
```
davidchisnall: I wasn't going to complain this time about a PR that mixes formatting with functional changes, but in this case the formatting changes are making the formatting less correct.  Please submit this PR with only the functional changes.

src/lkl/setup.c
```diff
@@ -646,14 +633,17 @@ static void lkl_mount_disk(
...
+    const int lkl_trace_internal_syscall_bak = tcfg->internal_syscall;
 
-    if ((sgxlkl_trace_lkl_syscall || sgxlkl_trace_internal_syscall) &&
+    if ((tcfg->lkl_syscall || tcfg->internal_syscall) &&
         (disk->roothash || is_encrypted_cfg(disk)))
     {
-        sgxlkl_trace_lkl_syscall = 0;
-        sgxlkl_trace_internal_syscall = 0;
+        // Check: const-casts acceptable?
+        *((bool*)&tcfg->lkl_syscall) = 0;
```
davidchisnall: Pointer l-value casts make me really nervous.  Why don't we just propagate the flags from the config structure to the existing globals early on, rather than making changes everywhere to directly use the config structure and then needing to modify const fields?

src/lkl/setup.c
```diff
@@ -696,13 +686,14 @@ static void lkl_mount_disk(
...
-        (lkl_trace_internal_syscall_bak && !sgxlkl_trace_internal_syscall))
+    if ((lkl_trace_lkl_syscall_bak && !tcfg->lkl_syscall) ||
+        (lkl_trace_internal_syscall_bak && !tcfg->internal_syscall))
     {
         SGXLKL_VERBOSE(
             "Disk encryption/integrity enabled: Re-enabling tracing.\n");
-        sgxlkl_trace_lkl_syscall = lkl_trace_lkl_syscall_bak;
-        sgxlkl_trace_internal_syscall = lkl_trace_internal_syscall_bak;
+        // Check: const-casts acceptable?
+        *((bool*)&tcfg->lkl_syscall) = lkl_trace_lkl_syscall_bak;
```
davidchisnall: As above.

src/lkl/setup.c
```diff
@@ -1385,42 +1376,6 @@ void lkl_start_init()
...
-
-    if (getenv_bool("SGXLKL_TRACE_SIGNAL", 0))
-        sgxlkl_trace_signal = 1;
-
-    if (getenv_bool("SGXLKL_TRACE_THREAD", 0))
-        sgxlkl_trace_thread = 1;
-
-    if (getenv_bool("SGXLKL_TRACE_DISK", 0))
-        sgxlkl_trace_disk = 1;
-
```
davidchisnall, line 1383: Propagating the fields from the config structure to the globals here would dramatically reduce the size of this diff.

src/main-oe/env.c
```diff
@@ -0,0 +1,84 @@
...
+    if (r == ULONG_MAX && errno == ERANGE)
+    {
+        r = def;
+    }
+    int m = 1;
+    switch (*endptr)
+    {
+        case 'G':
+        case 'g':
+            m *= 1024;
```
davidchisnall, line 25: Please add `__attribute__((fallthrough));` if fall-through is intended (I believe it is here, but I shouldn't have to read the entire switch statement to know that this is intentional).

davidchisnall (changes requested):

src/lkl/setup.c
wintersteiger: I just reverted all those changes before committing. Weird.

src/lkl/setup.c
wintersteiger: Me too, that's why I put the check here. I want the config to be const, but at this particular point the intention is really to modify the config. I don't want to keep the static globals, all settings should be in the config, i.e. for security reviews I'd like to have all state in `sgxlkl_enclave_state`, including the const config. Is it really necessary that we disable the traces here?

src/lkl/setup.c
wintersteiger: I already wrote the diff... it would dramatically increase my workload if I had to type in again :) 

As above, I want the config to be in one place, not distributed across a bunch of globals in different files.

src/main-oe/env.c
wintersteiger: I just moved this file. Looks intentional though, so I'll add the attributes.

src/lkl/setup.c
davidchisnall: As the comment explains, we disable tracing at this point so that we don't flood the logs.  The change is conflating the configuration state of the logging with the current state of the logging.  In the old version, the configuration state of the logging was in immutable environment variables (which was not a great place for it to be), the current state of the logging was in globals.  Now, we only have one place and we are using the configuration (which I believe should be immutable) to represent the current state.

Ideally, we'd allocate the config structure on a page boundary and do an ocall to make it read-only to prevent accidental modification.

wintersteiger: Yes, my question is whether stateful logging is really necessary. If so, IMHO, the cleanest solution would be to add a flag to `sgxlkl_enclave_state` to track whether they are currently enabled. 

I agree on the page-flag, that's a good idea. It's too easy to cast the `const` away.

davidchisnall: I can imagine other scenarios where we'd want to dynamically turn logging on and off (and, longer term, we probably want to expose APIs that we can poke in the debugger to do this), so would much rather we kept the globals.  I don't know what you use to commit things, but most git GUIs make it easy to revert parts of a change.

wintersteiger: I was kidding, it's just a few lines :) I would want to keep the globals in `sgxlkl_enclave_state` anyways.

wintersteiger: Sadly, I can't find a way to change the access permissions of a page. SGX1 does not support `EMODP*` and I can't write to a pre-allocated read-only page from within the enclave. Also, `EADD`/`EEXTEND` are disabled after `EINIT`. Since the config is public knowledge, we could theoretically save it outside of the enclave memory and `mprotect` it, but that doesn't feel like the right thing to do either.

davidchisnall: We have an ocall for this.

wintersteiger: Yes, there's the mprotect ocall, but that won't work on enclave pages, only on host pages. Are you suggesting that we should save it in host memory?

davidchisnall: The mprotect host call works on enclave pages.  That is the only place it is used.

davidchisnall: (It is advisory, the host can ignore it, but we're using it for debugging and defence in depth here)

wintersteiger: My experiments tell me that at least some part of this does indeed work. However, Intel says:

> mprotect will modify the OS permissions on that page , but not the Enclave Page Cache Map (EPCM) permissions. You need EMOD instructions to modify the EPCM permissions on that page.

(https://community.intel.com/t5/Intel-Software-Guard-Extensions/Can-I-dynamically-change-an-SGX-memory-page-s-permission-bits/td-p/1130989)

wintersteiger: Oh I see what you mean, yeah, if it's allowed to skip it, but happens not to, it's still useful for debugging.

wintersteiger: Done and it seems to work!

src/enclave/enclave_init.c
```diff
@@ -12,13 +12,52 @@
...
+}
+
+static void wipe_disk_keys()
+{
+    sgxlkl_enclave_disk_state_t* disk_state = sgxlkl_enclave_state.disk_state;
+    for (size_t i = 0; i < sgxlkl_enclave_state.num_disk_state; i++)
+    {
+        if (disk_state[i].key)
+        {
+            memset(disk_state[i].key, 0, disk_state[i].key_len);
```
davidchisnall: Needs to be `memset_s` or the compiler is free to delete the call.

src/enclave/enclave_init.c
```diff
@@ -12,13 +12,52 @@
...
 #include "enclave/wireguard_util.h"
-#include "shared/env.h"
+#include "shared/util.h"
 
 extern struct mpmcq __scheduler_queue;
 
 _Noreturn void __dls3(elf64_stack_t* conf, void* tos);
 extern void init_sysconf(long nproc_conf, long nproc_onln);
 
+static void get_disk_keys()
```
davidchisnall, line 33: Please add a comment explaining what this does.  It seems to be copying keys from one global to another and I am not sure why it's doing this.  For SKR, I assume we'll be connecting to a service to release the keys from userspace and doing the mounting there, using a cert that is passed to userspace in the ELF aux args vector, so I'm not sure what this copy is for.

src/enclave/enclave_signal.c
```diff
@@ -5,8 +5,8 @@
 
 #include <asm-generic/ucontext.h>
 
-#include <lkl_host.h>
 #include <lkl/setup.h>
+#include <lkl_host.h>
```
davidchisnall, line 9: There are still a load of formatting changes in this mixed with functional changes.  This makes code review difficult because I have to read each one and verify that there are no functional changes.

src/enclave/enclave_init.c
```diff
@@ -12,13 +12,52 @@
...
+{
+    /* Here or earlier: get keys from remote key provider or auxv. */
+    const sgxlkl_enclave_config_t* cfg = sgxlkl_enclave_state.config;
+    sgxlkl_enclave_disk_state_t* disk_states = sgxlkl_enclave_state.disk_state;
+    if (cfg->root.key)
+    {
+        uint8_t* key = cfg->root.key;
+        size_t len = cfg->root.key_len;
+        disk_states[0].key = oe_malloc(sizeof(uint8_t) * len);
+        memcpy(disk_states[0].key, key, len);
```
davidchisnall, line 45: Why do we need to do this?  Isn't `cfg->root.key` immutable?  Can't we just take a pointer to it, rather than a copy?

src/enclave/enclave_init.c
```diff
@@ -12,13 +12,52 @@
...
+        {
+            uint8_t* key = cfg->mounts[i].key;
+            size_t len = cfg->mounts[i].key_len;
+            disk_states[i + 1].key = oe_malloc(sizeof(uint8_t) * len);
+            memcpy(disk_states[i + 1].key, key, len);
+        }
+    }
+}
+
+static void wipe_disk_keys()
```
davidchisnall, line 61: This looks as if it wipes (or, rather, attempts to wipe) only the copy of the keys, not the persistent immutable keys.  What is the goal of this?  Please add a doc comment explaining why this function exists and what it is for.

src/include/enclave/enclave_state.h
```diff
@@ -51,8 +53,13 @@ typedef struct sgxlkl_enclave_state
     /* Memory shared with the host */
     sgxlkl_shared_memory_t shared_memory;
 
-    /* This flag is used by the tracing macros */
-    bool verbose;
+    /* Flags to track whether tracing macros are currently enabled */
```
davidchisnall: Please expand the doc comment to explain how these relate to the default values.

src/include/shared/oe_compat.h
```diff
@@ -19,13 +20,18 @@
...
 #define strtok_r oe_strtok_r
+#define snprintf oe_snprintf
+#define strtoul oe_strtoul
+#define memalign oe_memalign
 
 #else
 
 #include <assert.h>
 #include <inttypes.h>
+#include <malloc.h>
```
davidchisnall: `malloc.h` was deprecated around 30 years ago, why are we adding it here?

src/lkl/setup.c
```diff
@@ -123,7 +114,7 @@ static void lkl_prepare_rootfs(const char* dirname, int perm)
     {
         lkl_sys_chmod(dirname, perm);
     }
-
```
davidchisnall, line 126: This is a whitespace-only change that looks as if it's incorrect - why are we adding trailing whitespace?

src/lkl/setup.c
```diff
@@ -647,14 +629,16 @@ static void lkl_mount_disk(
...
 
-    int lkl_trace_lkl_syscall_bak = sgxlkl_trace_lkl_syscall;
-    int lkl_trace_internal_syscall_bak = sgxlkl_trace_internal_syscall;
+    const sgxlkl_trace_config_t* tcfg = &sgxlkl_enclave_state.config->trace;
+
+    const int lkl_trace_lkl_syscall_bak = tcfg->lkl_syscall;
+    const int lkl_trace_internal_syscall_bak = tcfg->internal_syscall;
 
-    if ((sgxlkl_trace_lkl_syscall || sgxlkl_trace_internal_syscall) &&
+    if ((tcfg->lkl_syscall || tcfg->internal_syscall) &&
```
davidchisnall: These checks are unnecessary, as is the backup copy taken two lines above.  We can now unconditionally set the current value to `false` and then reset it to the default value from the config later.

src/lkl/setup.c
```diff
@@ -697,13 +681,15 @@ static void lkl_mount_disk(
         dev_str = dev_str_enc;
     }
 
-    if ((lkl_trace_lkl_syscall_bak && !sgxlkl_trace_lkl_syscall) ||
-        (lkl_trace_internal_syscall_bak && !sgxlkl_trace_internal_syscall))
+    if ((lkl_trace_lkl_syscall_bak && !tcfg->lkl_syscall) ||
+        (lkl_trace_internal_syscall_bak && !tcfg->internal_syscall))
```
davidchisnall: This check is also redundant.  You can unconditionally reset the `sgxlkl_enclave_state.trace_enabled.lkl_syscall` to `sgxlkl_enclave_state.config->trace.lkl_syscall` and the same for the internal variant.

src/main-oe/env.c
```diff
@@ -0,0 +1,87 @@
...
+            __attribute__((fallthrough));
```
davidchisnall, line 26: Thank you!

src/enclave/enclave_signal.c
wintersteiger: To be precise, "a load" means exactly 1 line of diff? We've gone over this many times before - please merge the format-everything PR, otherwise you will be annoyed many, many times.

src/enclave/enclave_init.c
wintersteiger: This is in preparation for remote keys, which will need to arrive via the network instead of the immutable `cfg->root.key` (and I imagine the latter will vanish).

src/enclave/enclave_init.c
wintersteiger: Same as above, preparation for remote keys, of which there will only be copies.

src/include/shared/oe_compat.h
wintersteiger: To mirror whatever is in oemalloc.h.

src/lkl/setup.c
wintersteiger: "We" are adding whatever the auto-formatter produces; please fix `.clang_format` if it's incorrect or not what you expect.

src/include/shared/oe_compat.h
davidchisnall: Those definitions should come from `stdlib.h`, `malloc.h` is a pre-C89 header that has been deprecated for almost as long as I have been alive.

src/lkl/setup.c
davidchisnall: When you submit a PR, I expect you to review and add the relevant sections in each file from the diff before submitting, so presumably you either added this without reading it or you decided that the changes were meaningful.

src/enclave/enclave_signal.c
davidchisnall: There are several other changes.  We have, as you say, gone over it many times, yet it keeps happening.

src/lkl/setup.c
wintersteiger: I decided the changes were meaningful. They fix the problem of being formatted incorrectly. I won't submit a new PR every time I fix a typo.

src/enclave/enclave_signal.c
wintersteiger: Yes, and it will keep happening if you don't enforce it. The "several other changes" amount to exactly 1 line of semantic change in this file.

src/enclave/enclave_init.c
wintersteiger: Existing comments extended.

src/include/shared/oe_compat.h
wintersteiger: It was needed for `memalign`, which isn't in `stdlib.h`, but there was only a single use of that, so I `#ifdef`'d it instead.

src/lkl/setup.c
wintersteiger: Simplified, but I kept similar checks, so that the warning message is not emitted if those traces are already disabled in the config.

src/lkl/setup.c
wintersteiger: (Same as above, avoided emitting the spurious warning.)

src/enclave/enclave_oe.c
```diff
@@ -299,11 +299,25 @@ static void _read_eeid_config()
...
-    sgxlkl_enclave_config_t* cfg = oe_malloc(sizeof(sgxlkl_enclave_config_t));
-    if (!cfg)
-        sgxlkl_fail("out of memory, cannot allocate enclave config.\n");
-    sgxlkl_read_enclave_config(config_json, cfg, true);
-    sgxlkl_enclave_state.config = cfg;
+    size_t num_pages = 0;
+    const sgxlkl_enclave_config_page_t* cfg_pages =
+        sgxlkl_read_enclave_config(config_json, true, &num_pages);
+
+    /* Make the config pages read-only */
```
prp, line 306: Is this to prevent accidental modification?

src/include/enclave/enclave_state.h
```diff
@@ -51,8 +53,15 @@ typedef struct sgxlkl_enclave_state
...
-    /* This flag is used by the tracing macros */
-    bool verbose;
+    /* Flags to track whether tracing macros are currently enabled (initialized
+     * according to the settings in `config`; the respective traces may be
+     * disabled temporarily and this flags track the current state) */
+    struct
+    {
+        bool verbose;
+        bool lkl_syscall;
+        bool internal_syscall;
```
prp, line 63: Are these the only ones that we have left or are there others?

prp (approved):
LGTM, left minor questions.

src/enclave/enclave_oe.c
wintersteiger: Yes, exactly, the config is now completely const and won't change during execution.

src/include/enclave/enclave_state.h
wintersteiger: These are the only ones left at thus time, but if the tracing infrastructure gets updated, I can imagine more will be added.

prp: Don't we also have unsupported and overridden syscalls?

wintersteiger: Not sure what you're referring to. We have other traces, but those never get en/disabled after the config is sealed, so we don't need to track their state.

prp:
> There's one relevant getenv call in musl, see https://github.com/lsds/sgx-lkl-musl/blob/a6360f883fd906b18c1878254547ad72da28d7e6/ldso/dynlink.c#L822. This is now the only tracing/debug env var that gets passed through to the enclave as an env var:
> https://github.com/lsds/sgx-lkl/blob/9979a88fa90cca275dcb0822deef5e0393cea88a/src/main-oe/sgxlkl_params.c#L3
> 
> 
> I'm not sure this is still useful (see bunch of commented code further up in the file). Can I remove it?

This is part of a mechanism that, instead of loading debug symbols inside the enclave and then passing a pointer to gdb, it allows the enclave image to be mounted outside and thus gdb can access the files with symbols directly. For large applications (e.g. TF), this results in much better performance during debugging.

wintersteiger:
Great, thanks for the explanation, then it makes sense to keep it!

PR #807, opened 2020-08-19
Initial pass at documenting SGX-LKL threading.

davidchisnall:


docs/Threading.md
```diff
@@ -0,0 +1,167 @@
+The SGX-LKL threading implementation
+====================================
+
+SGX-LKL has four kinds of threads: ethreads, lthreads, Linux tasks, and pthreads.
+This document explains how they are implemented and how they relate.
+
+Ethreads are the lowest-level threads are ethreads.
```
SeanTAllen:
```suggestion
Ethreads are the lowest-level threads.
```

docs/Threading.md
```diff
@@ -0,0 +1,167 @@
...
+These are roughly analogous to VCPUs in conventional VMs.
+In SGX, each ethread corresponds to a thread-control structure (TCS), allocated by OpenEnclave.
```
SeanTAllen:
```suggestion
In SGX, each ethread corresponds to a thread-control structure (TCS), allocated by [Open Enclave](https://openenclave.io/sdk/).
```

docs/Threading.md
```diff
@@ -0,0 +1,167 @@
...
+Each ethread runs an instance of the lthread scheduler, which pulls lthreads from a multi-producer, multi-consumer queue (MPMCQ) to run.
+These threads run until they explicitly yield.
+
+The Linux kernel (LKL) manages its own task abstraction.
+Each Linux task is associated with an lthread.
+The LKL documentation refers to 'host threads', these are lthreads in our use.
+
+The userspace libc (musl) provides a POSIX threads (pthreads) implementation that sits atop Linux's tasks.
+Each pthread is run by the lthread scheduler but may be caused to block by LKL.
+
```
SeanTAllen, line 23: considering that you refer to "lowest level" in the beginning a diagram of "levels" and where each of the 4 types exist would be very helpful.

docs/Threading.md
```diff
@@ -0,0 +1,167 @@
...
+Ethreads
+========
+
+Ethreads are created by the host environment calling one of two ecalls, either [`sgxlkl_enclave_init`](https://github.com/lsds/sgx-lkl/blob/24467b08346cd7384eb93f845dec896a1d429711/src/enclave/enclave_oe.c#L395) (for the first ethread) or [`sgxlkl_ethread_init`](https://github.com/lsds/sgx-lkl/blob/24467b08346cd7384eb93f845dec896a1d429711/src/enclave/enclave_oe.c) (for subsequent ethreads).
```
SeanTAllen:
```suggestion
Ethreads are created by the host environment calling one of two ecalls, either [`sgxlkl_enclave_init`](https://github.com/lsds/sgx-lkl/blob/24467b08346cd7384eb93f845dec896a1d429711/src/enclave/enclave_oe.c#L395) (for the first ethread) or [`sgxlkl_ethread_init`](https://github.com/lsds/sgx-lkl/blob/24467b08346cd7384eb93f845dec896a1d429711/src/enclave/enclave_oe.c#L277) (for subsequent ethreads).
```

docs/Threading.md
```diff
@@ -0,0 +1,167 @@
...
+Each of these threads calls `lthread_run` to start an instance of the lthread scheduler.
+
+The ethreads do not use the FS segment on x86-64.
+The GS segment stores a pointer to the Open Enclave state and the lthread scheduler.
+The lthread context switch routine switches the FS segment register and makes this available for higher layers in the threading stack
```
SeanTAllen: a link to the context switch routine would be helpful here.

docs/Threading.md
```diff
@@ -0,0 +1,167 @@
...
+
+LThreads
+========
+
+The lthread library provides a *cooperative* threading model.
+The code lives in the [src/sched](../src/sched) directory in the SGX-LKL repository.
```
SeanTAllen, line 38:
```suggestion
The code lives in the [src/sched](https://github.com/lsds/sgx-lkl/tree/oe_port/src/sched) directory in the SGX-LKL repository.
```

because there are absolute links elsewhere.

docs/Threading.md
```diff
@@ -0,0 +1,167 @@
+The SGX-LKL threading implementation
```
SeanTAllen, line 1:
```suggestion
# The SGX-LKL threading implementation
```

docs/Threading.md
```diff
@@ -0,0 +1,167 @@
...
+These threads run until they explicitly yield.
+
+The Linux kernel (LKL) manages its own task abstraction.
+Each Linux task is associated with an lthread.
+The LKL documentation refers to 'host threads', these are lthreads in our use.
+
+The userspace libc (musl) provides a POSIX threads (pthreads) implementation that sits atop Linux's tasks.
+Each pthread is run by the lthread scheduler but may be caused to block by LKL.
+
+Ethreads
```
SeanTAllen, line 24:
```suggestion
## Ethreads
```

docs/Threading.md
```diff
@@ -0,0 +1,167 @@
...
+========
+
+Ethreads are created by the host environment calling one of two ecalls, either [`sgxlkl_enclave_init`](https://github.com/lsds/sgx-lkl/blob/24467b08346cd7384eb93f845dec896a1d429711/src/enclave/enclave_oe.c#L395) (for the first ethread) or [`sgxlkl_ethread_init`](https://github.com/lsds/sgx-lkl/blob/24467b08346cd7384eb93f845dec896a1d429711/src/enclave/enclave_oe.c) (for subsequent ethreads).
+Each of these threads calls `lthread_run` to start an instance of the lthread scheduler.
+
+The ethreads do not use the FS segment on x86-64.
+The GS segment stores a pointer to the Open Enclave state and the lthread scheduler.
+The lthread context switch routine switches the FS segment register and makes this available for higher layers in the threading stack
+
+LThreads
```
SeanTAllen, line 34:
```suggestion
## LThreads
```

docs/Threading.md
```diff
@@ -0,0 +1,167 @@
...
+========
+
+The lthread library provides a *cooperative* threading model.
+The code lives in the [src/sched](../src/sched) directory in the SGX-LKL repository.
+
+The main scheduler loop
```
SeanTAllen:
```suggestion
### The main scheduler loop
```

docs/Threading.md
```diff
@@ -0,0 +1,167 @@
...
+
+The [`_lthread_resume`](https://github.com/lsds/sgx-lkl/blob/47a5f0e718badfa85694a9de6222af41d9bfbb84/src/sched/lthread.c#L453) function does most of the work for running a thread.
+This sets up the initial thread state if required and then calls [`_switch`](https://github.com/lsds/sgx-lkl/blob/47a5f0e718badfa85694a9de6222af41d9bfbb84/src/sched/lthread.c#L482), a short assembly routine that saves all registers that are preserved across function calls and restores the same set of registers from a buffer.
+The `_switch` routine [restores the stack pointer](https://github.com/lsds/sgx-lkl/blob/47a5f0e718badfa85694a9de6222af41d9bfbb84/src/sched/lthread.c#L147) of the thread being switched to and pushes the new instruction pointer value onto the new stack.
+
+If this thread is a new thread, its instruction pointer was [set during thread initialisation](https://github.com/lsds/sgx-lkl/blob/47a5f0e718badfa85694a9de6222af41d9bfbb84/src/sched/lthread.c#L540).
```
SeanTAllen:
```suggestion
If the thread is a new thread, its instruction pointer was [set during thread initialisation](https://github.com/lsds/sgx-lkl/blob/47a5f0e718badfa85694a9de6222af41d9bfbb84/src/sched/lthread.c#L540).
```

docs/Threading.md
```diff
@@ -0,0 +1,167 @@
...
+The code lives in the [src/sched](../src/sched) directory in the SGX-LKL repository.
+
+The main scheduler loop
+-----------------------
+
+The main loop for the scheduler is in the [`lthread_run`](https://github.com/lsds/sgx-lkl/blob/47a5f0e718badfa85694a9de6222af41d9bfbb84/src/sched/lthread.c#L214) function.
+The function [pulls the next thread from the run queue](https://github.com/lsds/sgx-lkl/blob/47a5f0e718badfa85694a9de6222af41d9bfbb84/src/sched/lthread.c#L234) and then [calls `_lthread_resume`](https://github.com/lsds/sgx-lkl/blob/47a5f0e718badfa85694a9de6222af41d9bfbb84/src/sched/lthread.c#L242).
+
+The [`_lthread_resume`](https://github.com/lsds/sgx-lkl/blob/47a5f0e718badfa85694a9de6222af41d9bfbb84/src/sched/lthread.c#L453) function does most of the work for running a thread.
+This sets up the initial thread state if required and then calls [`_switch`](https://github.com/lsds/sgx-lkl/blob/47a5f0e718badfa85694a9de6222af41d9bfbb84/src/sched/lthread.c#L482), a short assembly routine that saves all registers that are preserved across function calls and restores the same set of registers from a buffer.
```
SeanTAllen:
```suggestion
`_lthread_resume` sets up the initial thread state if required and then calls [`_switch`](https://github.com/lsds/sgx-lkl/blob/47a5f0e718badfa85694a9de6222af41d9bfbb84/src/sched/lthread.c#L482), a short assembly routine that saves all registers that are preserved across function calls and restores the same set of registers from a buffer.
```

docs/Threading.md
```diff
@@ -0,0 +1,167 @@
...
+The `_switch` routine [restores the stack pointer](https://github.com/lsds/sgx-lkl/blob/47a5f0e718badfa85694a9de6222af41d9bfbb84/src/sched/lthread.c#L147) of the thread being switched to and pushes the new instruction pointer value onto the new stack.
+
+If this thread is a new thread, its instruction pointer was [set during thread initialisation](https://github.com/lsds/sgx-lkl/blob/47a5f0e718badfa85694a9de6222af41d9bfbb84/src/sched/lthread.c#L540).
+If this thread has previously yielded, the instruction pointer is the [return address for the `_switch` call that switched away from the lthread](https://github.com/lsds/sgx-lkl/blob/47a5f0e718badfa85694a9de6222af41d9bfbb84/src/sched/lthread.c#L134).
```
SeanTAllen:
```suggestion
If the thread has previously yielded, the instruction pointer is the [return address for the `_switch` call that switched away from the lthread](https://github.com/lsds/sgx-lkl/blob/47a5f0e718badfa85694a9de6222af41d9bfbb84/src/sched/lthread.c#L134).
```

docs/Threading.md
```diff
@@ -0,0 +1,167 @@
...
+
+Note that `_switch` does *not* switch between two arbitrary lthreads.
+One of the threads involved in a `_switch` call is always the scheduler.
+The `_switch` call in `_lthread_resume` switches to another thread, the call [in `_lthread_yield_cb`](https://github.com/lsds/sgx-lkl/blob/47a5f0e718badfa85694a9de6222af41d9bfbb84/src/sched/lthread.c#L340) and [in `_lthread_yield`](https://github.com/lsds/sgx-lkl/blob/47a5f0e718badfa85694a9de6222af41d9bfbb84/src/sched/lthread.c#L346) switch back to the scheduler.
+
+
+Locking
```
SeanTAllen:
```suggestion
### Locking
```

docs/Threading.md
```diff
@@ -0,0 +1,167 @@
...
+-------
+
+There are two primitives for building locks in the lthreads implementation.
+Ticket locks are spinlocks, futexes provide a fast path using atomic operations but then sleep.
+
+### Ticked locks
```
SeanTAllen:
```suggestion
#### Ticket locks
```

docs/Threading.md
```diff
@@ -0,0 +1,167 @@
...
+
+Ticket locks are defined and documented in [src/include/enclave/ticketlock.h](src/include/enclave/ticketlock.h).
+Acquiring a ticket lock will spin in a compare and swap (CAS) loop.
```
SeanTAllen: This isn't quite what you meant.

I'm not sure of a short way to resay it.

Perhaps "Waiting to acquire"?

docs/Threading.md
```diff
@@ -0,0 +1,167 @@
...
+If an lthread yields while holding a ticket lock, there is a chance of deadlock because the next thread to try to acquire the lock will spin in a CAS loop and not yield.
+This may, in turn, prevent the thread holding the lock from being able to run.
```
SeanTAllen, line 73: This perhaps requires more explanation.

docs/Threading.md
```diff
@@ -0,0 +1,167 @@
...
+
+Ticket locks are used to implement some parts of the lthread scheduler.
+This means that there is no way for them to be made to yield in their spin loop.
+Yielding transfers control from the current lthread to the scheduler and so the concept does not exist when the lthread is running.
+
+### Lthread futexes
```
SeanTAllen:
```suggestion
#### Lthread futexes
```

docs/Threading.md
```diff
@@ -0,0 +1,167 @@
...
+
+### Ticked locks
+
+Ticket locks are defined and documented in [src/include/enclave/ticketlock.h](src/include/enclave/ticketlock.h).
+Acquiring a ticket lock will spin in a compare and swap (CAS) loop.
+If an lthread yields while holding a ticket lock, there is a chance of deadlock because the next thread to try to acquire the lock will spin in a CAS loop and not yield.
+This may, in turn, prevent the thread holding the lock from being able to run.
+
+Ticket locks are used to implement some parts of the lthread scheduler.
+This means that there is no way for them to be made to yield in their spin loop.
```
SeanTAllen, line 80: I think the last 2 sentences are confusing. I'm not sure what you are trying to convey here.

docs/Threading.md
```diff
@@ -0,0 +1,167 @@
...
+Note that `_switch` does *not* switch between two arbitrary lthreads.
+One of the threads involved in a `_switch` call is always the scheduler.
+The `_switch` call in `_lthread_resume` switches to another thread, the call [in `_lthread_yield_cb`](https://github.com/lsds/sgx-lkl/blob/47a5f0e718badfa85694a9de6222af41d9bfbb84/src/sched/lthread.c#L340) and [in `_lthread_yield`](https://github.com/lsds/sgx-lkl/blob/47a5f0e718badfa85694a9de6222af41d9bfbb84/src/sched/lthread.c#L346) switch back to the scheduler.
+
+
+Locking
+-------
+
+There are two primitives for building locks in the lthreads implementation.
+Ticket locks are spinlocks, futexes provide a fast path using atomic operations but then sleep.
```
SeanTAllen, line 65:
```suggestion
```

docs/Threading.md
```diff
@@ -0,0 +1,167 @@
...
+
+### Ticked locks
+
+Ticket locks are defined and documented in [src/include/enclave/ticketlock.h](src/include/enclave/ticketlock.h).
```
SeanTAllen:
```suggestion
Ticket locks are spinlocks.
They are defined and documented in [src/include/enclave/ticketlock.h](../src/include/enclave/ticketlock.h).
```

docs/Threading.md
```diff
@@ -0,0 +1,167 @@
...
+
+Ticket locks are used to implement some parts of the lthread scheduler.
+This means that there is no way for them to be made to yield in their spin loop.
+Yielding transfers control from the current lthread to the scheduler and so the concept does not exist when the lthread is running.
+
+### Lthread futexes
+
+The lthread futex implementation is a simplified version of the Linux `futex` system call.
+A futex is a 32-bit word that can be used to hold state used to implement arbitrary locking primitives.
+On the fast path, all updates to the futex word are done by CPU atomic operations.
```
SeanTAllen, line 89: I find the next 3 sentences confusing. I'm not sure what the value they are bringing is. They don't really seem to provide any additional details that are important  to the following 3 paragraphs which seem to be the really important information in this context.

docs/Threading.md
```diff
@@ -0,0 +1,167 @@
...
+For example, a mutex could be implemented by treating a value of 0 as unlocked and a value of 1 as locked, with a CAS changing the state from 0 to 1 to acquire the lock on the fast path.
+The futex calls handle the slow path, where one or more threads need to wait.
+
+The `enclave_futex_wait` call suspends the thread until a corresponding `enclave_futex_wake` call on the same key.
```
SeanTAllen:
```suggestion
The `enclave_futex_wait` call suspends a given thread until a corresponding `enclave_futex_wake` call for that thread.
```

There's been no mention of "key" so while correct, I think it would lead to questions.

docs/Threading.md
```diff
@@ -0,0 +1,167 @@
...
+The futex system is responsible for ensuring that wake-up events are not lost and so acquires a global lock and then checks that the expected value is still present at the memory location.
```
SeanTAllen, line 94: If someone doesn't know how futexes work, I think this is going to be really confusing.

I'm not sure how it needs to be reworked.

docs/Threading.md
```diff
@@ -0,0 +1,167 @@
...
+This lock ensures that a `enclave_futex_wake` call either happens after the `enclave_futex_wait` has sent the calling thread to sleep or happens before and prevents the thread from sleeping.
+
+The sleeping threads are stored in a linked list, threaded through the thread pointers.
```
SeanTAllen:
```suggestion
The threads suspended by `enclave_futex_wait` are stored in a linked list, threaded through the thread pointers.
```

I don't know what "threaded through the thread pointers" means here.

docs/Threading.md
```diff
@@ -0,0 +1,167 @@
...
+A `enclave_futex_wait` call can return (waking up the sleeping thread) as a result of one of two things.
```
SeanTAllen, line 98: The rest of this is confusing. But only because of a lack of info.

So far we have discussed wait purely as a thing which requires a corresponding wake.

I think it would be better to pivot here and say something like "in addition, to watching wait/wake futex call pairs, enclave_futex_wait` can also be called using a timeout. The given thread will be suspended until it is woken up after it's timeout has passed".

Threads suspended via enclave_futex_wait with a timeout are awoken by `futext_tick`.
The lthread scheduler calls `futex_tick` in between scheduling threads.

docs/Threading.md
```diff
@@ -0,0 +1,167 @@
...
+Either a `enclave_futex_wake` acquires the lock, finds a waiting thread, and schedules it, or the `futex_tick` call triggers the timeout.
+
+The lthread scheduler calls `futex_tick` in between scheduling threads to wake up any sleeping threads that have timed out.
+This tries to acquire the ticket lock but returns immediately if it fails.
+This can happen only if a thread is currently doing a futex call or if another thread is handling the tick.
+
+### LKL host interface synchronisation primitives
```
SeanTAllen:
```suggestion
#### LKL host interface synchronisation primitives
```

docs/Threading.md
```diff
@@ -0,0 +1,167 @@
...
+
+The futex code is used in [src/lkl/posix-host.c](../src/lkl/posix-host.c) to implement two synchronisation primitives for use by LKL.
```
SeanTAllen, line 107:
```suggestion
The aforementioned enclave futex code is used in [src/lkl/posix-host.c](https://github.com/lsds/sgx-lkl/blob/oe_port/src/lkl/posix-host.c) to implement two synchronisation primitives for use by LKL.
```

to match the absolutes used most other places in the document. Changing to relative for those and keeping this relative would work as well.

docs/Threading.md
```diff
@@ -0,0 +1,167 @@
...
+These are not currently used anywhere else, though that may change in the future.
+
+#### `lkl_mutex`
```
SeanTAllen:
```suggestion
##### `lkl_mutex`
```

docs/Threading.md
```diff
@@ -0,0 +1,167 @@
...
+If the exchange moved from one of the locked states, the lock routine waits on the futex word with the expected state of locked-with-waiters.
+When the futex call returns, it retries the atomic exchange.
+
+On the unlock path, recursive mutexes simply decrement their recursion count until they are in the same state as a non-recursive mutex.
+The final unlock then does an atomic fetch-and-decrement on the futex word.
+This gives the old state.
+If this decrement transitioned from the locked-with-no-waiters state, there is nothing more to do.
+If the decrement instead transitioned from the locked-with-waiters state, the unlock routine transitions explicitly to the unlocked state and then calls `enclave_futex_wake` to wake up all waiters.
+
+#### `lkl_sem`
```
SeanTAllen:
```suggestion
##### `lkl_sem`
```

docs/Threading.md
```diff
@@ -0,0 +1,167 @@
...
+At this point, the kernel scheduler and the lthread scheduler may have different views of which thread is running.
+LKL's [`switch_to_host_task`](https://github.com/lsds/lkl/blob/385f721b339fe48b188b4924c2663e1ea2cdeb13/arch/lkl/kernel/threads.c#L199) routine is responsible for synchronising these states.
+This first marks the desired task as runnable and then runs the scheduler until it gets to that task.
+This time, `__switch_to` routine does a [long jump](https://github.com/lsds/lkl/blob/385f721b339fe48b188b4924c2663e1ea2cdeb13/arch/lkl/kernel/threads.c#L170) back out of the scheduler, leaving the expected thread as the currently running one.
+
+When a thread leaves a system call handler, its task is left in the `TASK_UNINTERRUPTIBLE` state.
+This prevents the Linux scheduler from attempting to reschedule it.
+It can then run, outside of the kernel, for as long as possible.
+
+Pthreads
```
SeanTAllen, line 173:
```suggestion
## Pthreads
```

SeanTAllen (commented):
I think that this is too long and covers too much for 1 document. In particular, the locking section under lthreads while good and important is a massive detour from the overview of thread types so that when I got to pthreads at the end I went (o right this was about threading not locking primitives).

I think that the locking related content should be moved to its own doc.

There's a lot of assumed knowledge to understand the document. I think it might be helpful to lay out at the beginning what it is expected that folks will understand.

I think knowledge of what a spinlock is, what a futex is, cas, atomic operation, mutex. recursive vs non-recursive are all required by the locking section.

I don't think that you really need much of that for understanding the threading beyond "there are things that control access to important bits" which is is my mind a further argument for breaking them apart into different docs.

docs/Threading.md
davidchisnall: I don't think we need two kinds of markup to say that this is h1.

docs/Threading.md
```diff
@@ -0,0 +1,167 @@
...
+
+The Linux kernel (LKL) manages its own task abstraction.
+Each Linux task is associated with an lthread.
+The LKL documentation refers to 'host threads', these are lthreads in our use.
+
+The userspace libc (musl) provides a POSIX threads (pthreads) implementation that sits atop Linux's tasks.
+Each pthread is run by the lthread scheduler but may be caused to block by LKL.
+
+Ethreads
+========
```
davidchisnall:
```suggestion
--------
```

docs/Threading.md
davidchisnall: I prefer the underline syntax for h1 and h2 in Markdown.

docs/Threading.md
davidchisnall:
```suggestion
The [lthread context switch](https://github.com/lsds/sgx-lkl/blob/796b346d4762a93be7bfc2ba5ce97b9ab840a4bd/src/sched/lthread.c#L466) [switches the FS segment register](https://github.com/lsds/sgx-lkl/blob/796b346d4762a93be7bfc2ba5ce97b9ab840a4bd/src/sched/lthread.c#L411) and makes this available for higher layers in the threading stack.
```

docs/Threading.md
davidchisnall: Looking at the doc, all of the headers are off by one in their level - both the main title and the section headings are h1.

docs/Threading.md
davidchisnall: There are only absolute links because GitHub doesn't let you link to markers in a file.  They'd all be relative links if they could be.  The rest of the docs also use relative links where possible.

docs/Threading.md
davidchisnall:
```suggestion
A thread acquiring a ticket lock will spin in a compare and swap (CAS) loop until it successfully acquires the lock.
```

docs/Threading.md
davidchisnall:
```suggestion
This may, in turn, prevent the thread holding the lock from being able to run.
In the simplest case, consider a system running with a single ethread.
If a thread yields while holding a ticket lock, the next thread that attempts to acquire the same ticket lock will spin and not yield to the scheduler.
The first thread is never rescheduled and so cannot release the lock.
Similar situations can occur with multiple ethreads if more than one lthread attempts to acquire ticket locks held by yielded lthreads.
```

docs/Threading.md
```diff
@@ -0,0 +1,167 @@
...
+### Ticked locks
+
+Ticket locks are defined and documented in [src/include/enclave/ticketlock.h](src/include/enclave/ticketlock.h).
+Acquiring a ticket lock will spin in a compare and swap (CAS) loop.
+If an lthread yields while holding a ticket lock, there is a chance of deadlock because the next thread to try to acquire the lock will spin in a CAS loop and not yield.
+This may, in turn, prevent the thread holding the lock from being able to run.
+
+Ticket locks are used to implement some parts of the lthread scheduler.
+This means that there is no way for them to be made to yield in their spin loop.
+Yielding transfers control from the current lthread to the scheduler and so the concept does not exist when the lthread is running.
```
davidchisnall:
```suggestion
Looking at the ticket lock code in isolation may lead you to believe that this limitation could be removed by adding a yield call in the the CAS loop.
This is not possible because yielding transfers control from the current lthread to the scheduler.
If the caller is the scheduler then there is no associated lthread that can yield.
```

docs/Threading.md
davidchisnall:
```suggestion
The `enclave_futex_wait` call suspends the calling thread until another thread makes a corresponding `enclave_futex_wake` call for the same futex word address, or until a timeout specified by the caller is reached.
```

docs/Threading.md
davidchisnall: I can't tell which three sentences these are referring to.  GitHub lets you select multiple lines for a comment - that makes it easier to see what the comment refers to.

docs/Threading.md
davidchisnall: See above: I wish GitHub had support for the in-document anchors in comments that the Listings package for LaTeX provides.

docs/Threading.md
```diff
@@ -0,0 +1,167 @@
...
+
+Ethreads are created by the host environment calling one of two ecalls, either [`sgxlkl_enclave_init`](https://github.com/lsds/sgx-lkl/blob/24467b08346cd7384eb93f845dec896a1d429711/src/enclave/enclave_oe.c#L395) (for the first ethread) or [`sgxlkl_ethread_init`](https://github.com/lsds/sgx-lkl/blob/24467b08346cd7384eb93f845dec896a1d429711/src/enclave/enclave_oe.c) (for subsequent ethreads).
+Each of these threads calls `lthread_run` to start an instance of the lthread scheduler.
+
+The ethreads do not use the FS segment on x86-64.
+The GS segment stores a pointer to the Open Enclave state and the lthread scheduler.
+The lthread context switch routine switches the FS segment register and makes this available for higher layers in the threading stack
+
+LThreads
+========
```
davidchisnall:
```suggestion
--------
```

docs/Threading.md
```diff
@@ -0,0 +1,167 @@
...
+
+The lthread library provides a *cooperative* threading model.
+The code lives in the [src/sched](../src/sched) directory in the SGX-LKL repository.
+
+The main scheduler loop
+-----------------------
```
davidchisnall:
```suggestion
### The main scheduler loop
```

docs/Threading.md
```diff
@@ -0,0 +1,167 @@
...
+If this thread is a new thread, its instruction pointer was [set during thread initialisation](https://github.com/lsds/sgx-lkl/blob/47a5f0e718badfa85694a9de6222af41d9bfbb84/src/sched/lthread.c#L540).
+If this thread has previously yielded, the instruction pointer is the [return address for the `_switch` call that switched away from the lthread](https://github.com/lsds/sgx-lkl/blob/47a5f0e718badfa85694a9de6222af41d9bfbb84/src/sched/lthread.c#L134).
+
+Note that `_switch` does *not* switch between two arbitrary lthreads.
+One of the threads involved in a `_switch` call is always the scheduler.
+The `_switch` call in `_lthread_resume` switches to another thread, the call [in `_lthread_yield_cb`](https://github.com/lsds/sgx-lkl/blob/47a5f0e718badfa85694a9de6222af41d9bfbb84/src/sched/lthread.c#L340) and [in `_lthread_yield`](https://github.com/lsds/sgx-lkl/blob/47a5f0e718badfa85694a9de6222af41d9bfbb84/src/sched/lthread.c#L346) switch back to the scheduler.
+
+
+Locking
+-------
```
davidchisnall:
```suggestion
### Locking
```

docs/Threading.md
```diff
@@ -0,0 +1,167 @@
...
+When `enclave_futex_wait` returns, this process repeats.
+
+The up operation is simpler.
+It does an atomic fetch-and-increment operation.
+If the transition is from zero to one then there may be waiters so it then calls `enclave_futex_wake` to wake them up.
+
+This implementation is not very efficient and could be improved.
+
+Linux tasks
+===========
```
davidchisnall:
```suggestion
Linux tasks
-----------
```

docs/Threading.md
```diff
@@ -0,0 +1,167 @@
...
+LKL's [`switch_to_host_task`](https://github.com/lsds/lkl/blob/385f721b339fe48b188b4924c2663e1ea2cdeb13/arch/lkl/kernel/threads.c#L199) routine is responsible for synchronising these states.
+This first marks the desired task as runnable and then runs the scheduler until it gets to that task.
+This time, `__switch_to` routine does a [long jump](https://github.com/lsds/lkl/blob/385f721b339fe48b188b4924c2663e1ea2cdeb13/arch/lkl/kernel/threads.c#L170) back out of the scheduler, leaving the expected thread as the currently running one.
+
+When a thread leaves a system call handler, its task is left in the `TASK_UNINTERRUPTIBLE` state.
+This prevents the Linux scheduler from attempting to reschedule it.
+It can then run, outside of the kernel, for as long as possible.
+
+Pthreads
+========
```
davidchisnall:
```suggestion
Pthreads
--------
```

docs/Threading.md
davidchisnall: Why?  The intro paragraph gives the overview of the contents of the sections.

docs/Threading.md
davidchisnall: Better?

docs/Threading.md
```diff
@@ -0,0 +1,167 @@
...
+====================================
+
+SGX-LKL has four kinds of threads: ethreads, lthreads, Linux tasks, and pthreads.
+This document explains how they are implemented and how they relate.
+
+Ethreads are the lowest-level threads are ethreads.
+These are roughly analogous to VCPUs in conventional VMs.
+In SGX, each ethread corresponds to a thread-control structure (TCS), allocated by OpenEnclave.
+
+The lthreads layer provides a cooperative threading library on top of ethreads.
```
KenGordon, line 11: lthreads are some library https://lthread.readthedocs.io/en/latest/intro.html ? The name confuses me, l is not for linux i guess.

docs/Threading.md
```diff
@@ -0,0 +1,167 @@
...
+Each ethread runs an instance of the lthread scheduler, which pulls lthreads from a multi-producer, multi-consumer queue (MPMCQ) to run.
+These threads run until they explicitly yield.
+
+The Linux kernel (LKL) manages its own task abstraction.
```
KenGordon, line 15: The Linux kernel (LKL) manages its own task abstraction, **in the usual linux way** perhaps? Or is it special to LKL and different to linux?

KenGordon (commented):
Other than the couple of things inline, and kind of skipping locking, I thing the main help to me would be to indicate which bits are normal Linux things, which bits specific to LKL, which bits SGX-LKL and which bits imported from other places, even if LKL also imports them. This lets me think "ah, so we have a linux scheduler on top of the lthreads scheduler or whatever...

KenGordon (approved):

docs/Threading.md
davidchisnall: It's based on that library, but heavily modified.

KenGordon: Mentioning that will help people understand that l is not for linux and that they are on their own anyway. I guess ethreads are enclave threads too, implied but not stated. It may be me being thick or that it is such a long time since I did guts level linux stuff.

docs/Threading.md
```diff
@@ -0,0 +1,171 @@
...
+
+If the thread is a new thread, its instruction pointer was [set during thread initialisation](https://github.com/lsds/sgx-lkl/blob/47a5f0e718badfa85694a9de6222af41d9bfbb84/src/sched/lthread.c#L540).
+If the thread has previously yielded, the instruction pointer is the [return address for the `_switch` call that switched away from the lthread](https://github.com/lsds/sgx-lkl/blob/47a5f0e718badfa85694a9de6222af41d9bfbb84/src/sched/lthread.c#L134).
+
+Note that `_switch` does *not* switch between two arbitrary lthreads.
+One of the threads involved in a `_switch` call is always the scheduler.
+The `_switch` call in `_lthread_resume` switches to another thread, the call [in `_lthread_yield_cb`](https://github.com/lsds/sgx-lkl/blob/47a5f0e718badfa85694a9de6222af41d9bfbb84/src/sched/lthread.c#L340) and [in `_lthread_yield`](https://github.com/lsds/sgx-lkl/blob/47a5f0e718badfa85694a9de6222af41d9bfbb84/src/sched/lthread.c#L346) switch back to the scheduler.
+
+
```
bodzhang, line 57:
```suggestion
After the execution is switched back to the scheduler, `lthread_run` checks for virtio wake up events and might add a sleeping lthread to the run queue. 

`lthread_run` maintains a contiguous idle loop count. In each loop, if no lthread is dequeued from the run queue and no sleeping thread is enqueued to the run queue,
the loop is considered as an idle loop. Any non-idle loop resets the contiguous idle loop count. Once the contiguous idle loop count reaches a set threshold,
`lthread_run` invokes an OE OCALL to put the ethread on sleep for a set period. The host is expected to resume the ethread after the sleep period passes.

`lthread_run` breaks out of the loop only when the enclave is terminating.
```

docs/Threading.md
```diff
@@ -0,0 +1,171 @@
...
+The `enclave_futex_wait` call suspends the calling thread until another thread makes a corresponding `enclave_futex_wake` call for the same futex word address, or until a timeout specified by the caller is reached.
+The futex system is responsible for ensuring that wake-up events are not lost and so acquires a global lock and then checks that the expected value is still present at the memory location.
+This lock ensures that a `enclave_futex_wake` call either happens after the `enclave_futex_wait` has sent the calling thread to sleep or happens before and prevents the thread from sleeping.
+
+The threads suspended by `enclave_futex_wait` are stored in a linked list, threaded through the thread pointers.
+A `enclave_futex_wait` call can return (waking up the sleeping thread) as a result of one of two things.
+Either a `enclave_futex_wake` acquires the lock, finds a waiting thread, and schedules it, or the `futex_tick` call triggers the timeout.
+
+The lthread scheduler calls `futex_tick` in between scheduling threads to wake up any sleeping threads that have timed out.
```
bodzhang, line 101: How does the `spin` in the scheduler code work?

docs/Threading.md
```diff
@@ -0,0 +1,171 @@
+The SGX-LKL threading implementation
+====================================
+
+SGX-LKL has four kinds of threads: ethreads, lthreads, Linux tasks, and pthreads.
+This document explains how they are implemented and how they relate.
+
+Ethreads are the lowest-level threads.
```
davidchisnall:
```suggestion
Ethreads (enclave threads) are the lowest-level threads.
```

docs/Threading.md
davidchisnall:
```suggestion
The lthreads (lightweight threads) layer provides a cooperative threading library on top of ethreads.
```

docs/Threading.md
davidchisnall: Saying Linux does things in the usual Linux way feels tautological.

docs/Threading.md
```diff
@@ -0,0 +1,171 @@
...
+These are roughly analogous to VCPUs in conventional VMs.
+In SGX, each ethread corresponds to a thread-control structure (TCS), allocated by [Open Enclave](https://openenclave.io/sdk/).
+
+The lthreads layer provides a cooperative threading library on top of ethreads.
+Each ethread runs an instance of the lthread scheduler, which pulls lthreads from a multi-producer, multi-consumer queue (MPMCQ) to run.
+These threads run until they explicitly yield.
+
+The Linux kernel (LKL) manages its own task abstraction.
+Each Linux task is associated with an lthread.
```
davidchisnall:
```suggestion
LKL, unlike other Linux architectures, expects the host environment to provide a threading abstraction.
LKL associates each Linux task with a thread provided by its host environment ('host thread').
In the case of SGX-LKL, lthreads provide LKL's host thread implementation and so each Linux task is backed by an lthread.
```

docs/Threading.md
davidchisnall: Reformatting (one sentence per line, so future diffs are not so painful) and correcting slightly:

```suggestion
After the running lthread yeidds, `lthread_run` checks whether any sleeping threads (those blocked waiting for event channels or futexes) are runnable and, if so, adds them to the queue.

`lthread_run` maintains a count of consecutive loop iterations in which there were not runnable lthreads.
Once this reaches a threshold, the scheduler issues an ocall that suspends execution of the ethread until either an event channel is signaled or a timeout expires.

`lthread_run` exits the loop and returns only when the enclave is terminating.
```

docs/Threading.md
davidchisnall: I think this was addressed by the change one paragraph up.

docs/Threading.md
davidchisnall: See change below.

docs/Threading.md
davidchisnall: Do you mean `a_spin`?  If so, it's just an x86 `pause` instruction (a hint to superscalar CPUs that they want monitor for cache coherency traffic rather than filling up their load/store pipeline with speculative loads at the same address).

bodzhang: No, I meant the `spins` variable. It counts down in the while loop and also gets reset to 0 before the OCALL. The logic beyond the code is not clear to me.

KenGordon:
Can we just have this merged please? This documentation is definitely better than the previous documentation.

vtikoo (approved):

Issue #709, opened 2020-08-03
Everything that is bad about signal passing


KenGordon:
Catch all for fixes to hopefully solve a whole bunch of actual bugs.

davidchisnall:
We run signals here: https://github.com/lsds/lkl/blob/86819ac5446e1fa31ca67db68aa165bbf07b90a9/arch/lkl/kernel/syscalls.c#L234

This does a loop here, popping signals off the pending list and delivering them:

https://github.com/lsds/lkl/blob/86819ac5446e1fa31ca67db68aa165bbf07b90a9/arch/lkl/kernel/signal.c#L68

There are a few things wrong here:

 - Signals are not delivered to the right thread, they are delivered to whichever thread returns from a syscall.
 - When a syscall returns early as a result of a signal being delivered, the signal delivery code does not set the correct return value.
 - When a signal is delivered, it does not wake up pending tasks.

I think the correct fix is to:

1. Iterate through pending signals.
2. Associate them with the task structs for the correct Linux threads.
3. Wake up any of the tasks that are currently in the sleeping state.
4. Run the scheduler to ensure that they actually wake.
5. Handle only the signals intended for the current thread on syscall return.

We probably need to do steps 1-4 in the idle task loop so that we wake up threads with pending signals even if no threads issue syscalls:

https://github.com/lsds/lkl/blob/86819ac5446e1fa31ca67db68aa165bbf07b90a9/arch/lkl/kernel/syscalls.c#L277

prp:
Our code for injecting signals into LKL looks quite different from that of other Linux architectures:

- We handle signals by directly invoking a handler here: https://github.com/lsds/lkl/blob/eecb2a320f3f2dd85942526df0a3d3abd402912c/arch/lkl/kernel/signal.c#L38
- In contrast, other Linux architectures create a signal frame and then call into the kernel: https://github.com/lsds/lkl/blob/86819ac5446e1fa31ca67db68aa165bbf07b90a9/arch/mips/kernel/signal.c#L846. This may mean that the signal is associated with the correct thread because a context switch can happen before the signal handler is invoked (?).
- Note that for synchronous signals, we switch to a host task explicitly here: https://github.com/lsds/lkl/blob/86819ac5446e1fa31ca67db68aa165bbf07b90a9/arch/lkl/kernel/traps.c#L40

davidchisnall:
> Note that for synchronous signals, we switch to a host task explicitly

[...]

PR #809, opened 2020-08-19
[Test] Adding busybox ping network test

hukoyu:
Adding busybox ping network test

### DIFF
--- a/.azure-pipelines/other/disabled_tests.txt
+++ b/.azure-pipelines/other/disabled_tests.txt
@@ -2,3 +2,4 @@ tests/virtio/python_read/Makefile
 tests/languages/java/network/Makefile
 tests/basic/pthread_detach/Makefile
 tests/languages/nodejs/Makefile
+tests/network/busybox/ping/Makefile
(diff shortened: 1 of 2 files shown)


davidchisnall (approved):

SeanTAllen:
Is the idea to have this disabled until it can pass?

hukoyu:
No @SeanTAllen this was passing locally  but failing in pipeline. I will investigate and make it pass in pipeline as well.

SeanTAllen:
Based on how it uses signals, I'm surprised it passed. Unless they've rewritten it since I looked in March.

hukoyu:
@SeanTAllen  do you mean this test cannot pass until p0 issue  https://github.com/lsds/sgx-lkl/issues/709 fixed?
Then it is strange that this is passing locally then.

SeanTAllen:
Yes. It didn't pass in March nor when I tried again in June.

hukoyu:
As @SeanTAllen  said this test is blocked by signaling p0 bug. Disabling this test for now. Will be enabled when blocking bug  #709 fixed.

vtikoo (approved):

PR #814, opened 2020-08-20
Remove unused fs creation in LTP test setup

SeanTAllen:


hukoyu:
@prp @davidchisnall @letmaik , @SeanTAllen  and I have question about the proper way of creating an image for test and configuration.

in https://github.com/lsds/sgx-lkl/tree/oe_port/tests/ltp/tests/batch.mk we are creating an additional image in host , and then using it in configuration https://github.com/lsds/sgx-lkl/tree/oe_port/tests/ltp/ltp_host_config.json and https://github.com/lsds/sgx-lkl/tree/oe_port/tests/ltp/ltp_tstapp_enclave_config.json

So here are my questions:
1) In current solution we create "ltp_tst_mntfs.img" (256 MB disk image to use in some LTP tests ) on host VM
2) We mount "ltp_tst_mntfs.img" to /data in https://github.com/lsds/sgx-lkl/tree/oe_port/tests/ltp/ltp_host_config.json
3) We mount /dev/vdb to /data  and set LTP_DEV = /dev/vdb in https://github.com/lsds/sgx-lkl/tree/oe_port/tests/ltp//ltp_tstapp_enclave_config.json

- Is this the right way of doing this? Here there are two images sgx-lkl image and ltp_tst_mntfs.img (mounted through config)
- Is it allowed to pass multiple images like that to sgx-lkl enclave?
- Are we supposed to create this test image in sgx-lkl image? 

Actually I am a bit confused and lost here and need help.

cc @jxyang @paulcallen @mikbras @bodzhang

SeanTAllen:
The code that this PR removes is currently unused. I believe it was intended to be removed during a prior PR.

PR #815, opened 2020-08-21
User-space separation

mikbras:
This PR, separates the C runtime (**libc** and **ldso**) into its own ELF memory region, effectively isolating it from the kernel. The resulting library is called **libsgxlkl-user.so** (built under the **./user** directory).

This PR depends on two other PRs:
    -  https://github.com/lsds/sgx-lkl-musl/pull/37
    - https://github.com/openenclave/openenclave/pull/3425

For more information see the README below.

https://github.com/lsds/sgx-lkl/pull/815/files?short_path=9e7b9bd#diff-9e7b9bdf8e7469060b7afdf451e7e430


src/enclave/enclave_init.c
```diff
@@ -89,8 +91,60 @@ static void init_wireguard()
...
 
-static int startmain(void* args)
+static long _lkl_syscall_wrapper(long no, long* params)
+{
+    //sgxlkl_warn("syscall begin: no=%u\n", no);
+    long ret = lkl_syscall(no, params);
+    //sgxlkl_warn("syscall end: ret=%u\n", ret);
+    return ret;
+}
+
```
davidchisnall, line 101: This looks like dead code that was left in during debugging.
```suggestion
```

src/enclave/enclave_init.c
```diff
@@ -89,8 +91,60 @@ static void init_wireguard()
...
+    typedef int (*sgxlkl_user_enter_proc_t)(void* args, size_t size);
+    sgxlkl_userargs_t args;
+    sgxlkl_user_enter_proc_t proc;
+
+    memset(&args, 0, sizeof(args));
+
+    if (!(proc = __oe_get_isolated_image_entry_point()))
+        sgxlkl_fail("failed to obtain user space entry point");
+
+    args.ua_lkl_syscall = _lkl_syscall_wrapper;
```
davidchisnall, line 120: See above comment about dead code.
```suggestion
    args.ua_lkl_syscall = lkl_syscall;
```

src/enclave/enclave_init.c
```diff
@@ -89,8 +91,60 @@ static void init_wireguard()
...
-static int startmain(void* args)
+static long _lkl_syscall_wrapper(long no, long* params)
+{
+    //sgxlkl_warn("syscall begin: no=%u\n", no);
+    long ret = lkl_syscall(no, params);
+    //sgxlkl_warn("syscall end: ret=%u\n", ret);
+    return ret;
+}
+
+static void _enter_user_space(
```
davidchisnall, line 102: This is fine for now, but we should be setting up the stack in the kernel and using a clone syscall to launch the thread on a pre-prepared stack using the libc entry point.

user/stubs.c
```diff
@@ -0,0 +1,119 @@
...
+**
+**==============================================================================
+*/
+
+long lkl_syscall(long no, long* params)
+{
+    long ret = __sgxlkl_userargs->ua_lkl_syscall(no, params);
+
+    return ret;
+}
```
davidchisnall, line 23: Fine for now, but please open an issue to track the fact that we should be always referencing `lkl_syscall` via a function pointer (or, ideally, an ifunc).

user/stubs.c
```diff
@@ -0,0 +1,119 @@
...
+
+/*
+**==============================================================================
+**
+** syscall:
+**
+**==============================================================================
+*/
+
+long lkl_syscall(long no, long* params)
```
davidchisnall, line 18: Does this need a hidden-visibility attribute, or is the build system still hiding this from userspace?

user/stubs.c
```diff
@@ -0,0 +1,119 @@
...
+
+long __sgxlkl_log_syscall(
+    // sgxlkl_syscall_kind type,
+    uint32_t type,
+    long n,
+    long res,
+    int params_len,
+    ...)
+{
+    sgxlkl_warn("__sgxlkl_log_syscall() unimplemented in user space");
```
davidchisnall, line 33: This is still called for `dup2` and `epoll_wait` (redirected system calls).  I think we should just remove the caller here, since we're removing the behaviour.

user/stubs.c
```diff
@@ -0,0 +1,119 @@
...
+{
+    /* ATTN: ignore variadic arguments */
+    return __sgxlkl_userargs->ua_sgxlkl_error(msg);
+}
+
+void sgxlkl_fail(const char* msg, ...)
+{
+    /* ATTN: ignore variadic arguments */
+    return __sgxlkl_userargs->ua_sgxlkl_fail(msg);
+}
```
davidchisnall, line 61: We shouldn't be calling any kernel code other than the syscall interface in userspace.  Please can you conditionally compile these so that they appear only in debug builds?

user/stubs.c
```diff
@@ -0,0 +1,119 @@
...
+
+void* enclave_mmap(
+    void* addr,
+    size_t length,
+    int mmap_fixed,
+    int prot,
+    int zero_pages)
+{
+    return __sgxlkl_userargs->ua_enclave_mmap(addr, length, mmap_fixed,
+        prot, zero_pages);
```
davidchisnall, line 76: Userspace's libc shouldn't issue this call directly.  We probably need to still call `enclave_mmap` because the kernel libc tries to mmap things before LKL is running, but this one should just do an mmap syscall.
```suggestion
    return mmap(addr, length, prot, MAP_PRIVATE | MAP_ANONYMOUS |  (mmap_fixed ? MAP_FIXED : 0, -1, 0); 
```

davidchisnall:
This PR contains some calls to functions that shouldn't exist.  I've opened #816 to track this - I don't believe the fix should block this PR landing.

user/stubs.c
davidchisnall: This is called in two places in malloc, neither sets `map_fixed`, so we don't actually need to handle this case.  It might be worth adding an assertion, something like:
```suggestion
    assert(prot == PROT_READ | PROT_WRITE);
    assert(!mmap_fixed);
    return mmap(addr, length, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0); 
```

This can then be removed once we are no longer linking a copy of libc into the kernel.

davidchisnall (approved):

mikbras:
Thank you David for your review. I'm sorry I created the PR without much background. Here are a few follow-up issues:
1. There are still some syscall bypasses that are beyond the scope of this PR to remove (see userargs.h).
2. This PR breaks syscall logging. We need to implement **syslog** as soon as possible.
3. A few builtin math functions are stubbed out (see stubs.c).

Also, the tests are broken since I forgot to add an install rule for **libsgxlkl-user.so**. I will update the PR in a couple of hours.

davidchisnall:
We moved syscall logging into the kernel already, the only thing that's broken is locking for translated syscalls (ones where we implement a legacy x86 syscall by calling its legacy version).  I don't think we actually need that.

user/stubs.c
```diff
@@ -0,0 +1,119 @@
...
+**
+** undefined builtins:
+**
+**==============================================================================
+*/
+
+#pragma GCC diagnostic ignored "-Wbuiltin-declaration-mismatch"
+
+void __muldc3()
+{
```
letmaik, line 115: I would call `abort()` here, just in case they actually get called.

src/main-oe/sgxlkl_run_oe.c
```diff
@@ -767,7 +767,19 @@ void get_signed_libsgxlkl_path(char* path_buf, size_t len)
...
+/* Determines path of libsgxlkl.so.signed */
+void get_signed_libsgxlkl_path(char* path_buf, size_t len)
+{
+    find_lib("libsgxlkl.so.signed", path_buf, len);
+}
+
+/* Determines path of libsgxlkl-user */
+void get_libsgxlkl_user_path(char* path_buf, size_t len)
+{
+    find_lib("libsgxlkl-user.so", path_buf, len);
```
vtikoo, line 782: For my understanding, is this shared object not signed like the kernel counterpart?

src/main-oe/sgxlkl_run_oe.c
```diff
@@ -1391,8 +1405,15 @@ void _create_enclave(
 
     setting.u.eeid = eeid;
 
+    // Format the follwing path <libsgxlkl>:<libsgxlkl_user>
```
vtikoo, line 1408:
```suggestion
    // Format the following path <libsgxlkl>:<libsgxlkl_user>
```

src/main-oe/sgxlkl_run_oe.c
```diff
@@ -1863,6 +1891,13 @@ int main(int argc, char* argv[], char* envp[])
     }
     sgxlkl_host_verbose_raw("result=%s\n", libsgxlkl);
 
+    sgxlkl_host_verbose("get_libsgxlkl_user_path... ");
+    if (!isolated_image_provided)
```
vtikoo, line 1895: This will always evaluate to true, I don't see where `isolated_image_provided` is being set again after initialization.

vtikoo: Maybe should happen on L1774?

src/enclave/enclave_init.c
```diff
@@ -166,9 +234,15 @@ int __libc_init_enclave(int argc, char** argv)
     SGXLKL_VERBOSE("calling _lthread_sched_init()\n");
     _lthread_sched_init(cfg->stacksize);
 
-    if (lthread_create(&lt, NULL, startmain, NULL) != 0)
+    /* Run startmain() in a new lthread */
     {
-        sgxlkl_fail("Failed to create lthread for startmain()\n");
+        static startmain_args_t args;
```
vtikoo, line 239: For my understanding, is this(creating code blocks with parantheses) a standard C feature or a gcc extension?

vtikoo: Also wondering if this was intentional, and if yes what was the rationale behind it.

src/main-oe/sgxlkl_run_oe.c
```diff
@@ -1746,6 +1770,10 @@ int main(int argc, char* argv[], char* envp[])
                 enclave_image_provided = true;
                 strcpy(libsgxlkl, optarg);
                 break;
+            case 'i':
+                enclave_image_provided = true;
```
vtikoo, line 1774:
```suggestion
                isolated_image_provided = true;
```

vtikoo (approved):
LGTM. Really appreciate the README, it helped a lot to understand the design!
Had two minor concerns, which should not be PR blocking -
- Passing a custom isolated image to `sgx-lkl-run-oe` seems to be broken. Right now, it will always use `libsgxlkl-user.so` even if the user specifies an override.
- Its not clear to me whether the userspace shared object is signed and whether we are using the signed version.

prp:
Merged as PR #818.

PR #817, opened 2020-08-25
Async signal work.

### LINKED ISSUE #709: Everything that is bad about signal passing

KenGordon:
Mostly this gets the signals to the correct thread. It also fixes a crash/lkl_bug when the cpu lock count got over 1. This was due to returning syscalls sending the signal while owning the cpu. My test harness signal handlers had printfs that would sometimes block, switch tasks and trigger the bug, so this code adjusts the syscall to only invoke signal handlers once it has given up the cpu.

I have concerns with whether my changes are appropriate to lkl when running without sgx constraints. 

This is still draft. I am on holiday today.

BTW Is there a way to tell if we are in "kernel mode" or not? Then code which expects only to be run is "user mode" can check it has not been reached by some unexpected path.

### DIFF
--- a/tests/basic/async_signal/Dockerfile
+++ b/tests/basic/async_signal/Dockerfile
@@ -1,0 +1,10 @@
+FROM alpine:3.6 AS builder
+
+RUN apk add --no-cache gcc musl-dev
+
+ADD *.c /
+RUN gcc -o signal -O0 -ggdb3 signal.c
+
+FROM alpine:3.6
+
+COPY --from=builder signal .
--- a/tests/basic/async_signal/Makefile
+++ b/tests/basic/async_signal/Makefile
@@ -1,0 +1,41 @@
+include ../../common.mk
+
+PROG=signal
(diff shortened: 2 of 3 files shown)


KenGordon:
PS please ignore the Makefile hackery. Also the locking needs changing back to a per thread lock. The one big lock was part of my debugging the lkl_bug going off.

PPS the lkl changes are on a branch ken/signal1 as well as being found via the update of the lkl submodule here.

Other obvious but not done stuff is wiring up the test, deciding how long it should run for and checking the per thread delivered counts are good.

davidchisnall:
Since most of these changes are in LKL, please raise an LKL PR for review.  IT looks as if you've also committed the results of the SGX-LKL build system patching the LKL build and some other bits.

SeanTAllen:
ill review this once there's a lkl PR as per @davidchisnall's comment.

KenGordon:
Replaced by fresh merges.

PR #820, opened 2020-08-27
Bump lkl to the latest version

SeanTAllen:
Brings in a fix from @AntonioND to logging of some messages related to signals.


SeanTAllen:
@davidchisnall @vtikoo you both approved https://github.com/lsds/lkl/pull/6, can you approve this as well so it can be bumped after it passes CI?

vtikoo (approved):

davidchisnall (approved):

PR #822, opened 2020-08-28
Update SGX-LKL to use LKL based on Linux v5.4.62

### LINKED ISSUE #821: Linux kernel version 5.3 is no longer being updated with functional and security patches
Linux kernel version 5.3 was end-of-life'd December, 2019 and is no longer being supported with patches. In addition to not receiving general bug fixes, version 5.3 is no longer being patched for security vulnerabilities.

The recommendation is to update to linux kernel version 5.4, LTS 2025.

prp:
This PR updates SGX-LKL to use LKL based on the Linux longterm kernel release v5.4. It uses the latest released revision, which is 5.4.62. The v5.4 longterm release will be maintained until December 2025.

Fixes #821

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,9 @@ src/include/shared/sgxlkl_args.h
 src/main-oe/sgxlkl_u.c
 src/enclave/sgxlkl_t.c
 
+user/enter.o
+user/stubs.o
+
 *.tar.gz
 *.class
 *.log
--- a/src/lkl/setup.c
+++ b/src/lkl/setup.c
@@ -1510,8 +1510,12 @@ extern inline int lkl_access_ok(unsigned long addr, unsigned long size)
     /* Set default state as access_ok */
     int ret = 1;
 
-    /* size 0 access should not be treated as invalid access */
-    if (!size)
+    /**
+     * NULL access or size 0 should not be treated as an invalid kernel access.
+     * (We have Linux file system code that calls this with a NULL pointer and
+     * expects it to succeed.)
+     **/
+    if (!addr || !size)
         return ret;
 
     ret = oe_is_within_enclave((void*)addr, size);
--- a/src/lkl/setup.c
+++ b/src/lkl/setup.c
@@ -190,7 +190,7 @@ static void lkl_mount_devtmpfs(const char* mntpoint)
 
 static void lkl_mount_shmtmpfs()
 {
-    int err = lkl_sys_mount("tmpfs", "/dev/shm", "tmpfs", 0, "rw,nodev");
+    int err = lkl_sys_mount("tmpfs", "/dev/shm", "tmpfs", 0, "mode=1777,rw");
     if (err != 0)
     {
         sgxlkl_fail("lkl_sys_mount(tmpfs) (/dev/shm): %s\n", lkl_strerror(err));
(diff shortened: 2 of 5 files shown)


KenGordon (commented):
Fine. Now to look at lkl/arch/lkl etc...

KenGordon:
These are all the changes in lkl/arch/lkl. They seem reasonable to me. Next I will try running some things.
```diff
kegordo@kegordo-acc2:/datadrive/kegordo/linux54/sgx-lkl/lkl/arch/lkl$ git diff -b origin/upstream-refactor . 
diff --git a/arch/lkl/include/asm/pgtable.h b/arch/lkl/include/asm/pgtable.h
index 9bb4593c9b33..3b2c832c67a6 100644
--- a/arch/lkl/include/asm/pgtable.h
+++ b/arch/lkl/include/asm/pgtable.h
@@ -40,11 +40,6 @@ void paging_init(void);
 extern void *empty_zero_page;
 #define ZERO_PAGE(vaddr)       (virt_to_page(empty_zero_page))

-/*
- * No page table caches to initialise.
- */
-#define pgtable_cache_init()   do { } while (0)
-
 /*
  * All 32bit addresses are effectively valid for vmalloc...
  * Sort of meaningless for non-VM targets.
diff --git a/arch/lkl/kernel/swiotlb_lkl.c b/arch/lkl/kernel/swiotlb_lkl.c
index 2c726d9d7668..46192d11eaa7 100644
--- a/arch/lkl/kernel/swiotlb_lkl.c
+++ b/arch/lkl/kernel/swiotlb_lkl.c
@@ -96,8 +96,8 @@ static dma_addr_t lkl_map_page(struct device *dev, struct page *page,
        }

        _start_dma_addr = __phys_to_dma(dev, (phys_addr_t)_lkl_io_tlb_start);
-       map = swiotlb_tbl_map_single(dev, _start_dma_addr, phys, size, dir,
-                                    attrs);
+       map = swiotlb_tbl_map_single(dev, _start_dma_addr, phys, size, size,
+                                    dir, attrs);
        if (map == (phys_addr_t)DMA_MAPPING_ERROR) {
                printk("LKL SWIOTLB: swiotlb_tbl_map_single map error\n");
                return map;
@@ -115,7 +115,7 @@ static void lkl_unmap_single(struct device *dev, dma_addr_t dev_addr,
        phys_addr_t phys = __dma_to_phys(dev, dev_addr);

        if (is_swiotlb_buffer(phys))
-               swiotlb_tbl_unmap_single(dev, phys, size, dir, attrs);
+               swiotlb_tbl_unmap_single(dev, phys, size, size, dir, attrs);

        return;
 }   
```

davidchisnall:
Looks good to me, with the proviso that I am not sure about the `lkl_access_ok` check.  I don't like carrying a patch in generic Linux code to work around the fact that our check passes when Linux code expects it to fail.  What broke when you made it return false for null pointers?

prp:
> Looks good to me, with the proviso that I am not sure about the `lkl_access_ok` check. I don't like carrying a patch in generic Linux code to work around the fact that our check passes when Linux code expects it to fail. What broke when you made it return false for null pointers?

@davidchisnall Yes, I also don't understand why this code otherwise worked before. It seems that the userspace access check gets called with NULL but the x86 version deems this to be fine...? It seems that all of this has been rewritten in 5.5 (https://github.com/torvalds/linux/commit/3c2659bd1db81ed6a264a9fc6262d51667d655ad), so this problem may go away anyway. Since LKL has a small number of patches outside of `arch/lkl` and `tools/lkl` anyway, it may be fine to keep this change as a workaround until we move to a more recent Linux kernel version.

bodzhang (approved):
Looks good.

davidchisnall (approved):

PR #827, opened 2020-09-15
Update to the current state of the lkl branh ken/sgxlkl709 which

KenGordon:
Contains signals fixes regarding some of issue 709. This is just to be able to present those fixes
to the CI.


KenGordon:
This attempt to run the CI against the unmerged branch failed, probably due to some changes in lkl adopting Linux 5.4. This PR has be replaced by a final one no 830.

PR #828, opened 2020-09-15
Take the changes from ken/sgxlkl709 dev branch

KenGordon:
Take the changes from ken/sgxlkl709 dev branch merged with upstream-refactor so as to check the CI is ok.


KenGordon:
This was to test it would pass the CI - replaced with actual PR that moves sgx-lkl to use the merged lkl at upstream-refactor.

Issue #231, opened 2020-05-11
Test samples in CI


letmaik:
Samples are not tested in CI and will inevitable break over time if they are not run manually regularly. They should be tested in CI, which may simply be a crash test.

letmaik:
Example of where this would have helped: https://github.com/lsds/sgx-lkl/issues/514. We really need to test samples in CI.

wintersteiger:
+1. I think this should be p0, it's the first thing prospective users will try!

bodzhang:
Enhancing test coverage with the samples will help to discover bugs and stability issues. I think this issue should stay as P1.

prp:
The samples almost completely overlap with the existing CI tests.

PR #829, opened 2020-09-15
CI: Add scripts to run samples automatically

AntonioND:
The scripts are a mix of bash and a bit of GNU expect for flexibility, as each sample needs to be tested in slightly different ways.

Each sample must have a test.sh script in its folder so that the CI detects it and runs it. They are executed by run_sample.sh script, created using run_test.sh as an example.

This PR adds several samples to the CI, but not all of them:

- The ml folder samples, for example, take far too long to be part of a regular CI run, and need caching of docker images, which is a task on its own.

- The openmp and nodejs samples seem to be broken.

This PR also fixes the following samples:

- ``languages/dotnet``: The following error happened when trying to run it:

```
FailFast:
Couldn't find a valid ICU package installed on the system. Set the configuration flag
System.Globalization.Invariant to true if you want to run with no globalization support.
```

By adding a new file with this configuration flag set to true, the sample works again. https://github.com/dotnet/core/issues/2186#issuecomment-472629489

- ``basic/helloworld``: This PR also simplifies the sample so that it matches the helloworld test.

- ``languages/java``:  When the number of ethreads is 1 the java app hangs after printing the message. With a higher number of ethreads it exits as expected.

This is a partial fix for https://github.com/lsds/sgx-lkl/issues/231, and is intended to make https://github.com/lsds/sgx-lkl/issues/332 easier by adding this new bash/expect framework.

### DIFF
--- a/.azure-pipelines/scripts/run_sample.sh
+++ b/.azure-pipelines/scripts/run_sample.sh
@@ -1,0 +1,86 @@
+#!/bin/bash
+
+if [ -z "$SGXLKL_ROOT" ]; then
+    echo "ERROR: 'SGXLKL_ROOT' is undefined. Please export SGXLKL_ROOT=<SGX-LKL-OE> source code repository"
+    exit 1
+fi
+if [ -z "$SGXLKL_BUILD_MODE" ]; then
+    echo "ERROR: 'SGXLKL_BUILD_MODE' is undefined. Please export SGXLKL_BUILD_MODE=<mode>"
+    exit 1
+fi
+
+#shellcheck source=.azure-pipelines/scripts/junit_utils.sh
+. "$SGXLKL_ROOT/.azure-pipelines/scripts/junit_utils.sh"
+#shellcheck source=.azure-pipelines/scripts/test_utils.sh
+. "$SGXLKL_ROOT/.azure-pipelines/scripts/test_utils.sh"
+
+# Initialize the variables and test case [mandatory].
+sample_mode=$1 # clean, init or run
+run_mode=$2 # run-hw or run-sw
+
+if [[ "$sample_mode" == "clean" ]]; then
+    ./test.sh clean
+    exit $?
+fi
+
+samples_dir=$SGXLKL_ROOT/samples
+sample_name="$(realpath --relative-to="$samples_dir" "$(pwd)")"
+sample_name="${sample_name//\//-}"
+sample_name+="-($SGXLKL_BUILD_MODE)-($run_mode)-($SGXLKL_ETHREADS-ethreads)"
+sample_class=$(realpath --relative-to="$samples_dir" "$(pwd)/..")
+test_suite="sgx-lkl-oe"
+
+if [[ -z $sample_name || -z $sample_class || -z $sample_mode ]]; then
+    echo -e "\n ERROR: sample_name sample_class or sample_mode not passed \n"
+    exit 1
+fi
+
+if [[ "$sample_mode" == "init" ]]; then
+    InitializeTestCase "$sample_name" "$sample_class" "$test_suite" "$run_mode"
+fi
+
+# Get the timeout from the test module
+DEFAULT_TIMEOUT=300
+if ! timeout=$(./test.sh gettimeout 2> /dev/null); then
+    timeout=$DEFAULT_TIMEOUT
+fi
+echo "Execution timeout: $timeout"
+
+case "$run_mode" in
+    "run-hw")
+       echo "Will run samples for run-hw"
+       ;;
+    "run-sw")
+       echo "Will run samples for run-sw"
+       ;;
+    *)
+       echo "Invalid run_mode parameter: $run_mode. Valid options: run-hw/run-sw"
+       exit 1;
+       ;;
+esac
+
+if [[ $sample_mode == "init" ]]; then
+    timeout --kill-after=$((timeout + 60)) $timeout ./test.sh init
+    script_exit=$?
+elif [[ $sample_mode == "run" ]]; then
+    timeout --kill-after=$((timeout + 60)) $timeout ./test.sh run "$run_mode"
+    script_exit=$?
+else
+    echo "Invalid sample_mode parameter: $sample_mode. Valid options: clean/init/run/gettimeout"
+    exit 1
+fi
+
+if [[ "$script_exit" == "124" ]]; then
+    echo "$run_mode: TIMED OUT after $timeout secs"
+elif [[ "$script_exit" != "0" ]]; then
+    echo "$run_mode: FAILED WITH EXIT CODE: $script_exit"
+fi
+
+if [[ $sample_mode == "init" ]]; then
+    echo "Sample initialization completed with EXIT CODE $script_exit"
+    return $script_exit
+fi
+
+echo "Sample run completed with EXIT CODE $script_exit"
+
+exit $script_exit
--- a/samples/basic/attack/read_memory.sh
+++ b/samples/basic/attack/read_memory.sh
@@ -21,7 +21,7 @@ sudo chown "$(id -u -n):$(id -g -n)" "$mem_file"
 
 echo "Searching memory for string \"${search_string}\" in \"${mem_file}\"..."
 
-if (strings "${mem_file}" | grep -i "${search_string}"); then
+if grep -Fxq "${search_string}" "${mem_file}"; then
         echo Match found.
 else
         echo No match found.
--- a/samples/basic/helloworld/test.sh
+++ b/samples/basic/helloworld/test.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+
+test_mode=$1
+run_mode=$2
+
+set -e
+
+if [[ "$test_mode" == "clean" ]]; then
+    make clean
+elif [[ "$test_mode" == "init" ]]; then
+    make
+elif [[ "$test_mode" == "run" ]]; then
+    make "$run_mode"
+elif [[ "$test_mode" == "gettimeout" ]]; then
+    # Default
+    exit 1
+fi
+
+exit 0
--- a/samples/common.sh
+++ b/samples/common.sh
@@ -1,0 +1,20 @@
+samples_dir=$(dirname $(realpath "$BASH_SOURCE"))
+SGXLKL_ROOT=$(realpath "${samples_dir}/..")
+
+if [[ -z "${SGXLKL_PREFIX}" ]]; then
+	export SGXLKL_STARTER=${SGXLKL_ROOT}/build/sgx-lkl-run-oe
+	export SGXLKL_DISK_TOOL=${SGXLKL_ROOT}/tools/sgx-lkl-disk
+	export SGXLKL_DOCKER_TOOL=${SGXLKL_ROOT}/tools/sgx-lkl-docker
+	export SGXLKL_CFG_TOOL=${SGXLKL_ROOT}/tools/sgx-lkl-cfg
+	export SGXLKL_SETUP_TOOL=${SGXLKL_ROOT}/tools/sgx-lkl-setup
+	export SGXLKL_GDB=${SGXLKL_ROOT}/tools/gdb/sgx-lkl-gdb
+	export SGXLKL_JAVA_RUN=${SGXLKL_ROOT}/tools/sgx-lkl-java
+else
+	export SGXLKL_STARTER=${SGXLKL_PREFIX}/bin/sgx-lkl-run-oe
+	export SGXLKL_DISK_TOOL=${SGXLKL_PREFIX}/bin/sgx-lkl-disk
+	export SGXLKL_DOCKER_TOOL=${SGXLKL_PREFIX}/bin/sgx-lkl-docker
+	export SGXLKL_CFG_TOOL=${SGXLKL_PREFIX}/bin/sgx-lkl-cfg
+	export SGXLKL_SETUP_TOOL=${SGXLKL_PREFIX}/bin/sgx-lkl-setup
+	export SGXLKL_GDB=${SGXLKL_PREFIX}/bin/sgx-lkl-gdb
+	export SGXLKL_JAVA_RUN=${SGXLKL_PREFIX}/bin/sgx-lkl-java
+fi
--- a/samples/containers/encrypted/test.sh
+++ b/samples/containers/encrypted/test.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+test_mode=$1
+run_mode=$2
+
+set -e
+
+if [[ "$test_mode" == "clean" ]]; then
+    make clean
+elif [[ "$test_mode" == "init" ]]; then
+    make
+elif [[ "$test_mode" == "run" ]]; then
+    make "$run_mode"-verity
+    # TODO: This doesn't work
+    #make "$run_mode"-integrity
+elif [[ "$test_mode" == "gettimeout" ]]; then
+    # Default
+    exit 1
+fi
+
+exit 0
--- a/samples/containers/redis/README.md
+++ b/samples/containers/redis/README.md
@@ -1,15 +1,18 @@
 Running Redis with SGX-LKL-OE
 =============================
 
-1. Ensure that you have set up netoworking and TLS support by running `tools/sgx-lkl-setup`.
+1. Make sure that you have installed ``redis-cli``. In Ubuntu, the package that
+   contains it is called ``redis-tools``.
 
-2. Build the Redis file sytem image:
+2. Ensure that you have set up netoworking and TLS support by running `tools/sgx-lkl-setup`.
+
+3. Build the Redis file sytem image:
 
 ```
 make
 ```
 
-3. Run the Redis service with:
+4. Run the Redis service with:
 
 ```
 make run-hw
@@ -21,7 +24,7 @@ or
 make run-sw
 ```
 
-4. Execute client requests against this instance:
+5. Execute client requests against this instance:
 
 ```
 ./run-redis-client.sh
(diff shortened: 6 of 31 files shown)


AntonioND:
I need to make this work in the CI, it looks like the build output isn't saved where I expected it to be.

AntonioND:
So now the 8 samples I've added to the CI pass (the CI failures are not this PR's fault).

They also pass in my VM (with ``run-hw`` and ``run-sw``):

```
run_mode=run-sw \
SGXLKL_ETHREADS=1 \
SGXLKL_BUILD_MODE=debug \
SGXLKL_ROOT=~/sgx-lkl \
.azure-pipelines/scripts/test_runner.sh samples  
```

samples/basic/attack/read_memory.sh
```diff
@@ -21,7 +21,8 @@ sudo chown "$(id -u -n):$(id -g -n)" "$mem_file"
 
 echo "Searching memory for string \"${search_string}\" in \"${mem_file}\"..."
 
-if (strings "${mem_file}" | grep -i "${search_string}"); then
+if grep -Fxq "${search_string}" "${mem_file}"
+then
```
SeanTAllen: I believe our standard is always "; then" for bash if/then formatting. Based on what I've seen in the project.

samples/containers/encrypted/test.sh
```diff
@@ -0,0 +1,21 @@
...
+
+set -e
+
+if [[ "$test_mode" == "clean" ]]; then
+    make clean
+elif [[ "$test_mode" == "init" ]]; then
+    make
+elif [[ "$test_mode" == "run" ]]; then
+    make "$run_mode"-verity
+    # TODO: This doesn't work
```
SeanTAllen, line 14: is there an issue for this TODO or is it going to be added to an existing issue?

samples/containers/encrypted/test.sh
```diff
@@ -0,0 +1,21 @@
...
+    #make "$run_mode"-integrity
+elif [[ "$test_mode" == "gettimeout" ]]; then
+    # 20 minutes
```
SeanTAllen: wow. out of curiosity, how long does this test usually take?

samples/containers/redis/README.md
```diff
@@ -1,6 +1,9 @@
 Running Redis with SGX-LKL-OE
 =============================
 
+0. Make sure that you have installed ``redis-cli``. In Ubuntu, the package that
```
SeanTAllen: I think the list should start at 1.

This "0" is weird to me.

samples/languages/dotnet/README.md
```diff
@@ -1,6 +1,23 @@
...
+```
+make run-hw
+```
+
+or
+
+```
+make run-sw
+```
+
```
SeanTAllen, line 15: Do we normally give two options with no explanation of which you should do? If yes, then I think this is ok, because it isn't worse than what we usually do, however, unless someone knows what run-hw and run-sw do, this provides no context for picking one.

I think it should have a short explanation for each "run with SGX on hardware that supports it" or "run emulated". Something to that effect.

samples/languages/java/README.md
```diff
@@ -1,6 +1,12 @@
 SGX-LKL-OE Java Sample Application
 ==================================
 
+0. Install OpenJRE and OpenJDK. In Ubuntu you can do:
```
SeanTAllen: same for "0" rather than "1"

samples/languages/java/test.sh
```diff
@@ -0,0 +1,24 @@
...
+run_mode=$2
+
+set -e
+
+if [[ "$test_mode" == "clean" ]]; then
+    make clean
+elif [[ "$test_mode" == "init" ]]; then
+    make
+elif [[ "$test_mode" == "run" ]]; then
+
```
SeanTAllen: looks like you left a blank line here accidentally.

SeanTAllen (commented):
Would it be reasonable to say based on the number of things with 20 minute time outs that if 1 or 2 or these hang that not all of them will be run before the CI run is terminated?

SeanTAllen:
@AntonioND rather than using expect, would it be reasonable to merely check the output value like we do with existing tests?

samples/containers/encrypted/test.sh
AntonioND: I'm going to create an issue for this, and for the openmp and nodejs broken samples.

samples/containers/encrypted/test.sh
AntonioND: This one doesn't take as long, sorry. The problem was the ``basic/attack`` sample, that can easily take 10-15 minutes in my VM. I just copy pasted the ``test.sh`` script from that folder and forgot to change it.

SeanTAllen: I could be wrong but I think you have 20 minutes for most timeouts that I saw in this PR. I would suggest lowering them if it isn't needed for them.

AntonioND:
> @AntonioND rather than using expect, would it be reasonable to merely check the output value like we do with existing tests?

That's not possible in all cases. For example, the ``language/java`` sample doesn't exit after it's done printing the demo message, I have to wait until it's printed and then send a ``"^C"`` for the test to finish. For the ``basic/attack`` sample you have to wait until the initial process is running and it prints "I'm ready to be attacked", attack it from a second process, and send a ``"\n"`` to the initial one to end. I tried doing this by having a countdown in the initial process and launching the second one at the right time, but it wasn't reliable at all. It is pretty reliable with expect.

EDIT: Maybe that behaviour of the java sample is a bug that could be added to the issue I'm going to create to fix the other samples...

SeanTAllen:
>  It is pretty reliable with expect.

That sounds like it is flakey. Are you saying it should be expected to fail sometimes with expect?

samples/languages/dotnet/README.md
AntonioND: Other samples are as vague as this, and I didn't want to spend too much time on changing readmes in this PR.

I think this should be done as part of https://github.com/lsds/sgx-lkl/issues/332, which is a bigger task. It would be nice to get rid of the Makefiles, or to make then easier to understand. But, definitely, all samples need to be cleaned up a bit.

AntonioND:
> That sounds like it is flakey. Are you saying it should be expected to fail sometimes with expect?

I can't be 100% sure it is always going to work, specially after seeing that some samples (like the java one) don't exit when I would expect them to.

Regarding the ones I'm testing with expect, I've never seen them fail so far. Of course, if someone changes the text output of the demo code the expect script will break, that's what I was thinking about when I wrote "pretty reliable".

In general I agree that expect is not a nice thing to use, but in the case of ``basic/attack`` I don't know how to do it in a different way, and in the case of ``languages/java``, assuming that the hangup is not a bug (I don't know the expected behaviour after printing the test output), I don't know how to make the test end either.

samples/basic/attack/read_memory.sh
AntonioND: Done.

AntonioND:
> Would it be reasonable to say based on the number of things with 20 minute time outs that if 1 or 2 or these hang that not all of them will be run before the CI run is terminated?

I've also fixed this, now there is only one sample that has that timeout. It's not needed when I run the CI, but it is needed when I run the test in my VM.

samples/containers/encrypted/test.sh
AntonioND: Done.

samples/containers/redis/README.md
AntonioND: Done.

samples/languages/java/test.sh
AntonioND: Fixed.

samples/languages/java/README.md
AntonioND: Fixed.

samples/containers/encrypted/test.sh
AntonioND: https://github.com/lsds/sgx-lkl/issues/831

AntonioND:
I've left a note in the ``languages/java`` sample that mentions that the sample isn't supposed to need expect scripts, and to remove them when the issue is fixed. I also mention in the comment the GitHub issue that needs to be fixed.

samples/basic/helloworld/enclave_config.json
```diff
@@ -1,8 +0,0 @@
-{
```
wintersteiger: I think we should keep `enclave_config.json`; it's the "new" way of passing a configuration to the launcher and the samples should show how they're used.

wintersteiger (approved):
I agree, it would be nice if we could get of `expect`, but I don't have a good alternative solution either.

samples/basic/helloworld/enclave_config.json
AntonioND: Makes sense. Restored and modified the Makefile to use it again.

AntonioND:
So @prp mentioned to me that the java sample is failing because the number of ethreads was 1. I've increased the number of ethreads in that sample and now it's working fine and it doesn't need expect scripts. However, the ``basic/attack`` one still needs them (but it's the only sample that does).

SeanTAllen:
LGTM other than the CI failures which I believe are unrelated.

vtikoo (approved):
LGTM

.azure-pipelines/scripts/run_sample.sh
```diff
@@ -0,0 +1,86 @@
...
+# Get the timeout from the test module
+DEFAULT_TIMEOUT=300
+if ! timeout=$(./test.sh gettimeout 2> /dev/null); then
+    timeout=$DEFAULT_TIMEOUT
+fi
+echo "Execution timeout: $timeout"
+
+case "$run_mode" in
+    "run-hw")
+       echo "Will run tests for run-hw"
```
hukoyu: Will run samples for run-hw
will be better. It will be better if you change messages with "tests" to "samples" in run_sample.sh

.azure-pipelines/scripts/run_sample.sh
```diff
@@ -0,0 +1,86 @@
...
+elif [[ "$script_exit" != "0" ]]; then
+    echo "$run_mode: FAILED WITH EXIT CODE: $script_exit"
+fi
+
+if [[ $sample_mode == "init" ]]; then
+    echo "Test initialization completed with EXIT CODE $script_exit"
+    return $script_exit
+fi
+
+echo "Test run completed with EXIT CODE $script_exit"
```
hukoyu: Same here. Use sample instead of test

samples/basic/helloworld/test.sh
```diff
@@ -0,0 +1,19 @@
...
+
+if [[ "$test_mode" == "clean" ]]; then
+    make clean
+elif [[ "$test_mode" == "init" ]]; then
+    make
+elif [[ "$test_mode" == "run" ]]; then
+    make "$run_mode"
+elif [[ "$test_mode" == "gettimeout" ]]; then
+    # Default
+    exit 1
```
hukoyu, line 16: This is 1 second and it is too short to be default value. May be 60 seconds is better.

samples/common.mk
```diff
@@ -0,0 +1,18 @@
+mkfile_dir=$(dir $(abspath $(lastword $(MAKEFILE_LIST))))
```
hukoyu: Do you need to redefine common.mk? Can you just include ../tests/common.mk ?

samples/common.sh
```diff
@@ -0,0 +1,20 @@
+samples_dir=$(dirname $(realpath "$BASH_SOURCE"))
+SGXLKL_ROOT=$(realpath "${samples_dir}/..")
+
+if [[ -z "${SGXLKL_PREFIX}" ]]; then
```
hukoyu, line 4: These are already exported in Makefile. Do we need to export again? Also SGXLKL_SETUP_TOOL exists in common.sh and not in common.mk just FYI

samples/containers/encrypted/test.sh
```diff
@@ -0,0 +1,21 @@
...
+    make clean
+elif [[ "$test_mode" == "init" ]]; then
+    make
+elif [[ "$test_mode" == "run" ]]; then
+    make "$run_mode"-verity
+    # TODO: This doesn't work
+    #make "$run_mode"-integrity
+elif [[ "$test_mode" == "gettimeout" ]]; then
+    # Default
+    exit 1
```
hukoyu, line 18: 1 second is too small

samples/containers/redis/run-redis-client.sh
```diff
@@ -2,14 +2,17 @@
...
 do
   echo $ redis-cli -h 10.0.1.1 -p 6379 set samplekey value${COUNTER}
   redis-cli -h 10.0.1.1 -p 6379 set samplekey value${COUNTER}
   echo $ redis-cli -h 10.0.1.1 -p 6379 get samplekey
   redis-cli -h 10.0.1.1 -p 6379 get samplekey
-  sleep 1
   let COUNTER+=1
 done
+
+echo "Test succeeded"
```
hukoyu: Sample succeeded

samples/containers/redis/test.sh
```diff
@@ -0,0 +1,27 @@
...
+    make
+elif [[ "$test_mode" == "run" ]]; then
+    if [[ "$run_mode" == "run-hw" ]]; then
+        ./run-hw.exp
+    elif [[ "$run_mode" == "run-sw" ]]; then
+        ./run-sw.exp
+    fi
+elif [[ "$test_mode" == "gettimeout" ]]; then
+    # Default
+    exit 1
```
hukoyu, line 24: 1 second too short

samples/languages/dotnet/test.sh
```diff
@@ -0,0 +1,19 @@
...
+
+if [[ "$test_mode" == "clean" ]]; then
+    make clean
+elif [[ "$test_mode" == "init" ]]; then
+    echo "Nothing to do"
+elif [[ "$test_mode" == "run" ]]; then
+    make "$run_mode"
+elif [[ "$test_mode" == "gettimeout" ]]; then
+    # Default
+    exit 1
```
hukoyu, line 16: 1 second too short

samples/languages/java/test.sh
```diff
@@ -0,0 +1,19 @@
...
+if [[ "$test_mode" == "clean" ]]; then
+    make clean
+elif [[ "$test_mode" == "init" ]]; then
+    make
+elif [[ "$test_mode" == "run" ]]; then
+    make "$run_mode"
+elif [[ "$test_mode" == "gettimeout" ]]; then
+    # Default
+    exit 1
```
hukoyu, line 16: 1 second is too short

samples/languages/python/test.sh
```diff
@@ -0,0 +1,34 @@
...
+    if [[ "$run_mode" == "run-hw" ]]; then
+        "${SGXLKL_STARTER}" --host-config=host-config.json \
+                --enclave-config=enclave-config.json --hw-debug
+    elif [[ "$run_mode" == "run-sw" ]]; then
+        "${SGXLKL_STARTER}" --host-config=host-config.json \
+                --enclave-config=enclave-config.json --sw-debug
+    fi
+elif [[ "$test_mode" == "gettimeout" ]]; then
+    # Default
+    exit 1
```
hukoyu, line 31: 1 second is too short

hukoyu (approved):
@AntonioND overall looks good. I provided some feedback. After addressing them this PR is good to be merged.

samples/basic/helloworld/test.sh
AntonioND: No, this means "default". This is how it works in the case of the Makefiles of the tests. If there is no target called ``gettimeout``, ``make`` will return 1. If the target is there, it prints the timeout length.

Take a look at ``tests/ltp/batch.mk``, target ``gettimeout``, for example.

samples/containers/redis/test.sh
AntonioND: Same comment as before.

samples/languages/java/test.sh
AntonioND: Same comment as before.

samples/languages/python/test.sh
AntonioND: Same comment as before.

samples/languages/dotnet/test.sh
AntonioND: Same comment as before.

samples/containers/encrypted/test.sh
AntonioND: Same comment as before.

samples/common.sh
AntonioND: Some samples need this in the bash script because not all of the commands are run in the Makefile. For example, ``SGXLKL_SETUP_TOOL`` is needed in ``containers/redis/test.sh``. A couple of them are needed in ``languages/python/test.sh``.

I only added ``SGXLKL_SETUP_TOOL`` to the bash script because it's not needed in the Makefiles.

.azure-pipelines/scripts/run_sample.sh
AntonioND: Done.

.azure-pipelines/scripts/run_sample.sh
AntonioND: Done.

samples/common.mk
AntonioND: Done.

samples/containers/redis/run-redis-client.sh
AntonioND: Done.

AntonioND:
I'm going to merge this PR now after addressing the last few comments. I'm still going to be working on the samples for a while, so I'm happy to push a PR with more changes if needed.

PR #830, opened 2020-09-15
Fix: Update to lkl that contains signaling changes.

KenGordon:
Fixes 644 (signals going to the wrong thread)
Fixes parts of 709
More work is due, tests need to be updated and other added.

Note to reviewer. This is the same changes as reviewed in the lkl sub repo over the past few days.


vtikoo:
@KenGordon As `/ltp/testcases/kernel/syscalls/rt_sigqueueinfo/rt_sigqueueinfo01` is passing now, should we re-enable it as part of this PR?

vtikoo (approved):
LGTM, it would be nice we could reenable rtsigqueueinfo01.

SeanTAllen (approved):

PR #834, opened 2020-09-23
Return error code when dev_register() fails

### LINKED ISSUE #370: lkl_virtio_netdev_add ignores errors

AntonioND:
Instead of just printing an error message, return an error code so that
the caller of lkl_virtio_netdev_add() knows that it has failed.

Along with PR #753, this fixes #370.

### DIFF
--- a/src/lkl/virtio_netdev.c
+++ b/src/lkl/virtio_netdev.c
@@ -72,9 +72,11 @@ int lkl_virtio_netdev_add(struct virtio_dev* netdev)
         return -1;
 
     ret = dev_register(netdev);
-
     if (ret < 0)
-        sgxlkl_info("Failed to register netdev \n");
+    {
+        sgxlkl_fail("Failed to register netdev\n");
+        return -1;
+    }
 
     return registered_dev_idx++;
 }


prp (approved):
LGTM.

src/lkl/virtio_netdev.c
```diff
@@ -72,9 +72,11 @@ int lkl_virtio_netdev_add(struct virtio_dev* netdev)
         return -1;
 
     ret = dev_register(netdev);
-
     if (ret < 0)
-        sgxlkl_info("Failed to register netdev \n");
+    {
+        sgxlkl_info("Failed to register netdev\n");
```
davidchisnall: I think this should just be `sgxlkl_fail`.  What does the caller do with the `-1` return value?

AntonioND: It's used in ``src/lkl/setup.c:1401``:

```
// Register network tap if given one
int net_dev_id = -1;
if (shm->virtio_net_dev_mem)
    net_dev_id = lkl_virtio_netdev_add(shm->virtio_net_dev_mem);
```

It looks like returning -1 is just interpreted as "no network tap present".

AntonioND: Also, I've changed the info to fail as you suggested.

davidchisnall: I don't want to have to reason about the security implications of the network device being not present.  Simply failing is easy to reason about.

davidchisnall (approved):

src/lkl/virtio_netdev.c
AntonioND: Wait, I'm confused, do you mean that the changes are good as they are, or that I need a stronger way to fail?

SeanTAllen: He was explaining why he wanted to fail instead of having the error bubble up.

davidchisnall: Confusing conversation history caused by a race condition: I wrote that before your second command and the push of the change that turned it to a fail.

davidchisnall:
The CI failure looks unrelated.

prp:
Didn't @KenGordon recently reactivate the failing test? The test actually seems to have passed but it still returns an error code...?

vtikoo (approved):
LGTM

PR #832, opened 2020-09-17
Enable tests rt_sigaction02, rt_sigqueuinfo01, sigwaitinfo01

KenGordon:
Enable tests rt_sigaction02, rt_sigqueuinfo01, sigwaitinfo01 following signaling improvements (see issue 709).

Issue #644, opened 2020-07-18
Signals not delivered to correct thread


prp:
The following test uses `rt_sigqueueinfo` to send a signal to a thread: https://github.com/vtikoo/sgx-lkl/commit/7e539eeabb99973f2dda92ded49251db3ea86cab

When running the test directly, it succeeds:
```
$ ./rt_sigqueueinfo01
pid=1204
tid within main thread: 1204
errno: 0, Success
tid of child thread: 1205
tst_checkpoint_wake(0)
errno: 0, Success
ret: 0, errno: 0, Success
tid within signal handler: 1205
Received correct signal and data!
ret: 0, errno: 0, Success
rt_sigqueueinfo() was successful
```

When run with SGX-LKL, it fails as follows:
```
[[  SGX-LKL ]] libc_start_main_stage2(): Calling app main: rt_sigqueueinfo01
pid=36
tid within main thread: 1
errno: 0, No error information
tid of child thread: 47
tst_checkpoint_wake(0)
errno: 0, No error information
ret: 0, errno: 0, No error information
rt_sigqueueinfo() failed (ret=-1 errno=3 (No such process))
```

The problem seems to be that the assigned tid for the main thread is inconsistent with the pid.

(cc: @vtikoo @davidchisnall)

prp:
Ok, the above problem goes away when we remove the [syscall bypass for `gettid()`](https://github.com/lsds/sgx-lkl-musl/blob/13d6039488f62f39a5152a4818bc4529899ee9af/src/internal/syscall.h#L44). (The pid/tid assignment is then still not consistent, but this may be ok.)

The problem then is that the signal is always delivered to the thread calling `rt_sigqueueinfo` and not the thread whose tid is specified.

The reason is that the code in LKL's [`handle_signal()`](https://github.com/lsds/lkl/blob/86819ac5446e1fa31ca67db68aa165bbf07b90a9/arch/lkl/kernel/signal.c#L36)  does not create a proper stack frame for the signal handler, but rather just calls the handler directly.

[...]

PR #837, opened 2020-09-28
Disable rt_sigqueueinfo test which is failing with ethreads > 1

KenGordon:
Revert change to one test made in https://github.com/lsds/sgx-lkl/pull/832 as it fails in the CI with ethreads > 1.

### DIFF
--- a/tests/ltp/ltp-batch2/ltp_disabled_tests.txt
+++ b/tests/ltp/ltp-batch2/ltp_disabled_tests.txt
@@ -755,7 +755,7 @@
 #/ltp/testcases/kernel/syscalls/rt_sigaction/rt_sigaction03
 #/ltp/testcases/kernel/syscalls/rt_sigprocmask/rt_sigprocmask01
 #/ltp/testcases/kernel/syscalls/rt_sigprocmask/rt_sigprocmask02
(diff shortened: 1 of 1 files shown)


KenGordon:
Apologies. I was on holiday Thursday and Friday, travelling so I could turn up in person at the office...

vtikoo:
Does that mean #644  should be re-opened? That issue is specifically about rt_sigqueueinfo01.

KenGordon:
> Does that mean #644 should be re-opened? That issue is specifically about rt_sigqueueinfo01.

It passes with one ethread. I think there is a general problem with futex waiting. The return code of an interrupted futex wait ought to be EINTR (4) but is actually 516. I was investigating this on Wednesday before being away for the tgkill01 case. In that one if I replace the convoluted lkl tst_checkpoint call with a sleep the test passes. Perhaps I should create a new issue specifically for that but given I have discovered it as a contributory cause of the tgkill01 case I think it would be a duplicate.

This makes the CI pass again and I think that any discussion that delays that is not helping people know if they have broken anything.

vtikoo (approved):

SeanTAllen (approved):

davidchisnall (approved):

KenGordon:
See https://github.com/lsds/sgx-lkl/issues/838 for the futex problem.

PR #839, opened 2020-10-01
pthread support documentation

vtikoo:
### Summary
This document has two goals:
- capture LKL scheduling related concepts and gotchas relevant to pthreads/clone() threads. 
- document thread creation and termination flows.

### DIFF
--- a/docs/LKLSchedulingAndPthreadsSupport.md
+++ b/docs/LKLSchedulingAndPthreadsSupport.md
@@ -1,12 +1,12 @@
 # LKL Scheduling and Pthreads Support
 
 ## Pre-requisites
-/doc/Threading.md
+Please read the [threading docs](Threading.md) for a good overview of LKL threading concepts.
 
 ## Thread identities
 A thread has 3 identities:
 - Linux process context - `task_struct`
-- LKL arch specific process context - `thread_info`
+- LKL arch specific process context - `[thread_info](https://github.com/lsds/lkl/blob/e041aa71e03a142ecef542c005b07d13e2a3b722/arch/lkl/include/asm/thread_info.h#L29)`
 - LKL host thread context, implemented in SGX-LKL by `lthread`s.
 
 Not every lthread corresponds to a LKL thread. In conventional LKL usage, the Linux process contexts are allocated lazily on the first system call. In cloned threads, the Linux process context is allocated eagerly during thread creation.
@@ -44,10 +44,10 @@ From the LKL paper -
 
 O. Purdila, L. A. Grijincu and N. Tapus, "LKL: The Linux kernel library," 9th RoEduNet IEEE International Conference, Sibiu, 2010, pp. 328-333.
 
-The environment-provided semaphore is the `sched_sem` field in thread_info.
+The environment-provided semaphore is the [`sched_sem`](https://github.com/lsds/lkl/blob/e041aa71e03a142ecef542c005b07d13e2a3b722/arch/lkl/include/asm/thread_info.h#L34) field in `thread_info`.
 
 ## thread_sched_jb(): setjmp/longjmp with LKL scheduler
-`thread_sched_jb` provides a mechanism to do non-local jumps between the current thread context and the scheduler. LKL uses a [host provided implementation](https://github.com/lsds/sgx-lkl/blob/0c95637ca94ca9a92680169fb8941ffc12e33c3f/src/lkl/jmp_buf.c) for the setjmp and longjmp operations.
+[`thread_sched_jb`](https://github.com/lsds/lkl/blob/e041aa71e03a142ecef542c005b07d13e2a3b722/arch/lkl/include/asm/sched.h#L7) provides a mechanism to do non-local jumps between the current thread context and the scheduler. LKL uses a [host provided implementation](https://github.com/lsds/sgx-lkl/blob/0c95637ca94ca9a92680169fb8941ffc12e33c3f/src/lkl/jmp_buf.c) for the setjmp and longjmp operations.
 
 Before calling the scheduler, a flag(TIF_SCHED_JB) is added to the task's flags. This helps identify this task later, in the architecture dependent task switching routine `__switch_to`.
 Once the thread is in `__switch_to(prev, next)`, the cpu ownership is transferred to the next task(It is implied that `thread_sched_jb` was called with the CPU lock held). And then if the `TIF_SCHED_JB` flag is set, the thread long jumps back to `threads_sched_jb`. 
@@ -131,7 +131,7 @@ switch_to_host_task(task: task_struct)
 
 ## pthread_create
 The libc code calls `clone()` to create a new thread.
-There are two different types of integration points for LKL during process creation. (All of the following functions are defined in arch/lkl/kernel/threads.c):
+There are two different types of integration points for LKL during process creation. (All of the following functions are defined in [`arch/lkl/kernel/threads.c`](https://github.com/lsds/lkl/blob/e041aa71e03a142ecef542c005b07d13e2a3b722/arch/lkl/kernel/threads.c)):
 i) Setting up the arch specific thread_info struct.
     - `alloc_thread_stack_node()`
     - `init_ti()`
@@ -148,7 +148,7 @@ _do_fork(*args: kernel_clone_args)
     V
     - copy_process(...,args)
         - ...
-        | // where is current picked from?
+        |
         V
         - dup_task_struct(current, node)
             - ...
@@ -194,7 +194,7 @@ A pthread issues a `SYS_exit` [here](https://github.com/lsds/sgx-lkl-musl/blob/c
 
 As part of the exit system call, the thread eventually yields to the LKL scheduler and sleeps on its scheduler semaphore [here](https://github.com/lsds/lkl/blob/e041aa71e03a142ecef542c005b07d13e2a3b722/arch/lkl/kernel/threads.c#L194).
 
-At some later point, the LKL architecture specific integration point for thread cleanup - `free_thread_stack`, is called from the `ksoftirqd` kthread. 
+At some later point, the LKL architecture specific integration point for thread cleanup - [`free_thread_stack`](https://github.com/lsds/lkl/blob/e041aa71e03a142ecef542c005b07d13e2a3b722/arch/lkl/kernel/threads.c#L126), is called from the `ksoftirqd` kthread. 
 
 Stacktrace at the entry of `free_thread_stack`:
 ```
--- a/docs/LKLSchedulingAndPthreadsSupport.md
+++ b/docs/LKLSchedulingAndPthreadsSupport.md
@@ -1,17 +1,17 @@
 # LKL Scheduling and Pthreads Support
 
 ## Pre-requisites
-Please read the [threading docs](Threading.md) for a good overview of LKL threading concepts.
+Please read [Threading.md](Threading.md) for a good overview of SGX-LKL threading concepts.
 
 ## Thread identities
 A thread has 3 identities:
 - Linux process context - `task_struct`
-- LKL arch specific process context - `[thread_info](https://github.com/lsds/lkl/blob/e041aa71e03a142ecef542c005b07d13e2a3b722/arch/lkl/include/asm/thread_info.h#L29)`
+- LKL arch specific process context - [`thread_info`](https://github.com/lsds/lkl/blob/e041aa71e03a142ecef542c005b07d13e2a3b722/arch/lkl/include/asm/thread_info.h#L29)
 - LKL host thread context, implemented in SGX-LKL by `lthread`s.
 
 Not every lthread corresponds to a LKL thread. In conventional LKL usage, the Linux process contexts are allocated lazily on the first system call. In cloned threads, the Linux process context is allocated eagerly during thread creation.
 
-In general once LKL knows about a lthread, there is a 1:1 mapping between these identities. We'll detail some exceptions later in the `pthread_create` section.
+In general once LKL knows about a lthread, there is a 1:1 mapping between these identities. Later we'll detail an exception to this in the [pthread_create](#pthread_create) section.
 
 ## LKL's CPU Lock: 
 The LKL CPU lock allows a single thread to be running inside LKL.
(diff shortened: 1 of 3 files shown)


tests/basic/pthread_detach/README.md
```diff
@@ -0,0 +1,181 @@
+# Preliminaries
+## Thread identities
+A thread has 3 identities:
+- Linux process context - `task_struct`
+- LKL arch specific process context - `thread_info`
+- LKL host thread context - `lthread`
```
davidchisnall: lthread is SGX-LKL-specific terminology, it might be worth explicitly stating that this is called the host thread in LKL terminology.
```suggestion
- LKL host thread context, implemented in SGX-LKL by `lthread`s.
```

tests/basic/pthread_detach/README.md
```diff
@@ -0,0 +1,181 @@
...
+
+For most of a thread's lifetime these 3 identities should have a 1:1 mapping between them.
```
davidchisnall: It's not quite bijective.  Not every lthread corresponds to an LKL thread.  In conventional LKL usage, the Linux process contexts are allocated lazily on first system call.  In cloned threads, the Linux process context is allocated eagerly.

tests/basic/pthread_detach/README.md
```diff
@@ -0,0 +1,181 @@
...
+
+The environment-provided semaphore is the `sched_sem` field in thread_info.
+
+## thread_sched_jb()
+`thread_sched_jb` is LKL's way of doing co-operative multitasking, by yielding to the scheduler. Its invoked from - 
+
+1. `lkl_cpu_put`: If the Linux scheduler run queue has >= 1 tasks.
+2. `switch_to_host_task(task)`: If `current` process context != task.
+3. `lkl_syscall`: If syscall is NR_REBOOT, yields to scheduler after running the syscall.
+
```
davidchisnall: Document how `thread_sched_jb` exits?  That's the most confusing thing here.

tests/basic/pthread_detach/README.md
```diff
@@ -0,0 +1,181 @@
...
+
+## switch_to_host_task(task)
+`switch_to_host_task` function ensures that task being passed as input parameter is the `current` process context. If this is already the case the function returns early. Else, this causes transfer of control to the LKL scheduler so that it can schedule `task`. 
+A flag(TIF_SCHED_JB) is added to the task's flags before yielding to the LKL scheduler. This helps identify this task from the other tasks in the architecture dependent switching routine `__switch_to`. Once the thread is in `__switch_to`, control is switched back to `switch_to_host_task` via `threads_sched_jb`. After that the thread sleeps on its scheduler semaphore and waits for the LKL scheduler to wake it.
```
davidchisnall: Minor nit, but if you put a newline after each sentence break then future diffs are a lot easier to review and read (and smaller and edits to different sentences are easier to automatically merge).

tests/basic/pthread_detach/README.md
```diff
@@ -0,0 +1,181 @@
...
+
+Call graph for switch_to_host_task to __switch_to interactions:
+```
+switch_to_host_task(task: task_struct)
+    - task.ti.tid = lthread_self()
```
davidchisnall: Explain why this is needed, I remember that we were both confused by it the first time we saw it.

tests/basic/pthread_detach/README.md
```diff
@@ -0,0 +1,181 @@
...
+    - wake_up_process(task)
+    |
+    V
+    - thread_sched_jb()
+        - current.ti.flags = TIF_SCHED_JB
+        - current.state = TASK_UNINTERRUPTIBLE
+    (long jump to scheduler)
```
davidchisnall: Are you sure?  I thought this called the scheduler and long-jumped out of it.

tests/basic/pthread_detach/README.md
```diff
@@ -0,0 +1,181 @@
...
+            |
+            V
+            - context_switch(rq, prev, next)
+                |
+                V
+                - __switch_to(prev, next)
+                    - set abs_prev
+                    - change cpu ownership to next
+                    - Wake lthread backing _next
+                    - As prev has TIF_SCHED_JB flag set, long jump back.
```
davidchisnall: Explain here what the currently running thread will be (from the kernel's perspective and from the lthread scheduler's perspective).

tests/basic/pthread_detach/README.md
```diff
@@ -0,0 +1,181 @@
...
+    |
+    V
+    (back in threads_sched_jb())
```
davidchisnall: This might be easier to understand if one indent corresponded to one call depth.

tests/basic/pthread_detach/README.md
```diff
@@ -0,0 +1,181 @@
...
+
+# Pthreads support
+
+## pthread_create
+The libc code calls `clone()` to create a new thread.
+There are two different types of integration points for LKL during process creation. (All of the following functions are defined in arch/lkl/kernel/threads.c):
+i) Setting up the arch specific thread_info struct.
+    - alloc_thread_stack_node()
+    - init_ti()
+    - setup_thread_stack()
```
davidchisnall:
```suggestion
    - `alloc_thread_stack_node()`
    - `init_ti()`
    - `setup_thread_stack()`
```

tests/basic/pthread_detach/README.md
```diff
@@ -0,0 +1,181 @@
...
+ii) Creating the host thread(lthread) backing the Linux task.
+    - copy_thread_tls()
```
davidchisnall:
```suggestion
    - `copy_thread_tls()`
```

This function has a confusing name.  I initially assumed that it copied the TLS for the thread.  In fact, it is a variant of `copy_thread` that takes a TLS pointer as well as a stack pointer and instruction pointer.

tests/basic/pthread_detach/README.md
```diff
@@ -0,0 +1,181 @@
...
+Call graph within LKL caused by pthread_create()->clone():
+```
+_do_fork(*args: kernel_clone_args)
+    - ...
+    |
+    |
+    V
+    - copy_process(...,args)
+        - ...
+        | // where is current picked from?
```
davidchisnall: The value of `current` changes only if the scheduler is invoked, so it is the thread that was set as current in the `switch_to_host_task` in the `lkl_syscall` entry path.

tests/basic/pthread_detach/README.md
```diff
@@ -0,0 +1,181 @@
...
+```
+
+After the clone() syscall is run, the state of newly created child process must be set as non-runnable in LKL. This in turn requires the child process be the `current` process context. If this wasn't enough there is one more complication here. For LKL to schedule the thread, it needs to transfer the CPU lock ownership to it. As mentioned earlier, CPU lock is assigned by lthread id. So briefly the lthread identity of the parent is shared with the child thread. 
+
+After setting the state as `TASK_UNINTERRUPTIBLE`, the lthread id for the child thread is restored.
+
+
+## pthread_exit
+
+A pthread issues a `SYS_exit` [here](https://github.com/lsds/sgx-lkl-musl/blob/oe_port/src/thread/pthread_create.c#L123), or if its detach attribute is set it exits [here](https://github.com/lsds/sgx-lkl-musl/blob/oe_port/src/thread/x86_64/__unmapself.s#L18).
```
davidchisnall: If you're including line numbers, it's probably a good idea to link to a specific revision.  This link will drift away from the anchor if the code above it is modified.

tests/basic/pthread_detach/README.md
```diff
@@ -0,0 +1,181 @@
+# Preliminaries
+## Thread identities
+A thread has 3 identities:
+- Linux process context - `task_struct`
+- LKL arch specific process context - `thread_info`
+- LKL host thread context, implemented in SGX-LKL by `lthread`s.
+
+For most of a thread's lifetime these 3 identities should have a 1:1 mapping between them.
+One instance where this doesn't hold is on the return path of the clone() system call. (This is covered in more detail later in the `pthread_create` section.)
```
SeanTAllen: What is the purpose of introducing 1 instance here? To me, its a bit confusing and could be covered later.

My impression is that you are trying to establish basic intutions. I think you would be best served by that. Something like "in general, you can think of this as... BLAH" and then "we'll detail some exceptions".

And perhaps something to the effect of "these rules arent' really true but they can ease you into understanding the more complicated picture better"

SeanTAllen (commented):
I'm not sure how valuable this will be to someone just stepping in to work on threading. My suspicion is "not much". Once you understand the basics of things, I think this could help a lot.

If there's another document that explains the relationship between different types of thread like things in more detail and how to start navigating, it should be linked to from here.

If there isn't I think that broader overview would help a great deal.

Really it depends on the intent here and who the target audience is.

vtikoo:
@SeanTAllen totally agree that this is not a beginner-friendly document. My intention was to cover the tricky parts(according to me) of SGX-LKL threading stack. And establish just enough ground to be able to do that.
I think https://github.com/lsds/sgx-lkl/blob/oe_port/docs/Threading.md is sufficient pre-requisite for this document. I can add that in the document too.

SeanTAllen:
If that was your intent then I think is good in terms of content.

tests/basic/pthread_detach/README.md
vtikoo: Have documented how `switch_to_host_task` usage of this function returns in the next section. Let me look up how the other two usages exit.

tests/basic/pthread_detach/README.md
vtikoo: According to my understanding, the perspectives are different only during the epilogue in `lkl_syscall` for `SYS_clone`.

tests/basic/pthread_detach/README.md
vtikoo: That was not the behavior I was seeing, for some reason this value was always the crypto manager task.

tests/basic/pthread_detach/README.md
vtikoo: Added comment to doc.

tests/basic/pthread_detach/README.md
vtikoo: Done.

tests/basic/pthread_detach/README.md
vtikoo: Done.

tests/basic/pthread_detach/README.md
vtikoo: Thanks for catching that! Fixed.

SeanTAllen (approved):

docs/LKLSchedulingAndPthreadsSupport.md
```diff
@@ -0,0 +1,226 @@
...
+
+The environment-provided semaphore is the [`sched_sem`](https://github.com/lsds/lkl/blob/e041aa71e03a142ecef542c005b07d13e2a3b722/arch/lkl/include/asm/thread_info.h#L34) field in `thread_info`.
+
+## thread_sched_jb(): setjmp/longjmp with LKL scheduler
+[`thread_sched_jb`](https://github.com/lsds/lkl/blob/e041aa71e03a142ecef542c005b07d13e2a3b722/arch/lkl/include/asm/sched.h#L7) provides a mechanism to do non-local jumps between the current thread context and the scheduler. LKL uses a [host provided implementation](https://github.com/lsds/sgx-lkl/blob/0c95637ca94ca9a92680169fb8941ffc12e33c3f/src/lkl/jmp_buf.c) for the setjmp and longjmp operations.
+
+Before calling the scheduler, a flag(TIF_SCHED_JB) is added to the task's flags. This helps identify this task later, in the architecture dependent task switching routine `__switch_to`.
+Once the thread is in `__switch_to(prev, next)`, the cpu ownership is transferred to the next task(It is implied that `thread_sched_jb` was called with the CPU lock held). And then if the `TIF_SCHED_JB` flag is set, the thread long jumps back to `threads_sched_jb`.
```
prp:
```suggestion
Once the thread is in `__switch_to(prev, next)`, the CPU ownership is transferred to the next task (It is implied that `thread_sched_jb` was called with the CPU lock held). And then if the `TIF_SCHED_JB` flag is set, the thread long jumps back to `threads_sched_jb`. 
```

docs/LKLSchedulingAndPthreadsSupport.md
```diff
@@ -0,0 +1,226 @@
...
+1. `switch_to_host_task(task)`: If `current` process context != task.
+2. `lkl_cpu_put`: If the Linux scheduler run queue has >= 1 tasks.
+3. `lkl_syscall`: If syscall is NR_REBOOT, yields to scheduler after running the syscall.
+
+Usages 2. and 3. use `thread_sched_jb` are in the epilogue of `lkl_syscall`, and effectively allow the thread to invoke the schedule and relinquish the CPU lock.  
+
+The `switch_to_host_task` usage is more involved and discussed next.
+
+## switch_to_host_task(task)
+`switch_to_host_task` function ensures that task being passed as input parameter is the `current` process context.
```
prp:
```suggestion
`switch_to_host_task` function ensures that a task being passed as input parameter is the `current` process context.
```

docs/LKLSchedulingAndPthreadsSupport.md
```diff
@@ -0,0 +1,226 @@
...
+                - Set process name field "task_struct->comm"
+                - Set ti.cloned_child for parent
+                - return 0 on success or -ENOMEM
+    |
+    V
+    - ...
+
+```
+
+After the clone() syscall is run, the state of newly created child process must be set as non-runnable in LKL. This in turn requires the child process be the `current` process context. If this wasn't enough there is one more complication here. For LKL to schedule the thread, it needs to transfer the CPU lock ownership to it. As mentioned earlier, CPU lock is assigned by lthread id. So briefly the lthread identity of the parent is shared with the child thread.
```
prp:
```suggestion
After the clone() syscall is run, the state of newly created child process must be set as non-runnable in LKL. This in turn requires the child process be the `current` process context. If this wasn't enough there is one more complication here. For LKL to schedule the thread, it needs to transfer the CPU lock ownership to it. As mentioned earlier, the CPU lock is assigned by lthread id. So briefly the lthread identity of the parent is shared with the child thread. 
```

prp (approved):
Looks great -- caught some minor typos.

